PR #49, opened 2025-11-10
fix: template auto-discovery for go run and lvt serve

adnaan:
## Summary

Fixes template auto-discovery when running apps via `go run` from `lvt serve` command.

## Root Cause

Template auto-discovery failed because `runtime.Caller` returned Go build cache paths when running via `go run`, instead of the project directory where templates are located.

## Solution

Implemented a **3-tier discovery strategy** with environment variable support and smart fallback:

**Priority Order:**
1. Explicit `WithTemplateBaseDir()` option (highest priority)
2. `LVT_TEMPLATE_BASE_DIR` environment variable  
3. `runtime.Caller` detection (existing behavior - backward compatible)
4. Smart fallback (if auto-detected and no templates found):
   - Try `./templates` if it exists
   - Otherwise use `.` (current directory)

### 1. Added `WithTemplateBaseDir()` Option ([template.go](https://github.com/livetemplate/livetemplate/blob/fix/template-auto-discovery/template.go#L245-L252))
```go
func WithTemplateBaseDir(dir string) Option {
    return func(c *Config) {
        c.TemplateBaseDir = dir
    }
}
```

### 2. Environment Variable Support ([config.go](https://github.com/livetemplate/livetemplate/blob/fix/template-auto-discovery/config.go))
- Added `TemplateBaseDir` field to `EnvConfig` struct
- Parse `LVT_TEMPLATE_BASE_DIR` environment variable
- Convert to option in `ToOptions()` method

### 3. Smart Fallback Logic ([internal/discovery/discovery.go](https://github.com/livetemplate/livetemplate/blob/fix/template-auto-discovery/internal/discovery/discovery.go))
- Try `runtime.Caller` first (backward compatible)
- If no templates found with auto-detected directory, try smart fallback
- Ensures `baseDir` is never empty

### 4. lvt serve Integration
Companion PR in lvt repository sets `LVT_TEMPLATE_BASE_DIR` to project root when starting apps.

## Testing

**New Test:** `TestDiscoverTemplateFiles_SmartFallback` with 3 scenarios:
- ✅ Fallback to `./templates` when it exists
- ✅ Fallback to `.` (current directory) when `templates/` doesn't exist  
- ✅ Nested discovery from current directory (recursive search)

**All existing tests pass** ✅

## What This Fixes

✅ **lvt serve with go run** - Explicit `LVT_TEMPLATE_BASE_DIR` env var works  
✅ **go run directly** - `runtime.Caller` + smart fallback works  
✅ **Compiled binaries with templates/** - Smart fallback finds `./templates`  
✅ **Colocated templates** - Smart fallback uses `.` (discovers all templates recursively)  
✅ **Production deployments** - Users can set `LVT_TEMPLATE_BASE_DIR` env var  
✅ **Backward compatibility** - Existing apps continue to work with `runtime.Caller`

## Related

- Addresses known issue documented in CODE_REVIEW_BACKLOG.md #22
- Companion lvt PR needed for full integration

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -486,6 +486,28 @@ func TestLoadEnvConfig_AllVariables(t *testing.T) {
 	}
 }
 
+// TestLoadEnvConfig_TemplateBaseDir tests template base directory configuration
+func TestLoadEnvConfig_TemplateBaseDir(t *testing.T) {
+	clearEnv(t)
+	os.Setenv("LVT_TEMPLATE_BASE_DIR", "/custom/templates")
+	defer os.Unsetenv("LVT_TEMPLATE_BASE_DIR")
+
+	config, err := LoadEnvConfig()
+	if err != nil {
+		t.Fatalf("LoadEnvConfig failed: %v", err)
+	}
+
+	if config.TemplateBaseDir != "/custom/templates" {
+		t.Errorf("Expected TemplateBaseDir to be /custom/templates, got %s", config.TemplateBaseDir)
+	}
+
+	// Verify it converts to option
+	opts := config.ToOptions()
+	if len(opts) != 1 {
+		t.Errorf("Expected 1 option, got %d", len(opts))
+	}
+}
+
 // clearEnv removes all LVT_ environment variables
 func clearEnv(t *testing.T) {
 	t.Helper()
@@ -496,6 +518,7 @@ func clearEnv(t *testing.T) {
 		"LVT_DEV_MODE",
 		"LVT_WEBSOCKET_DISABLED",
 		"LVT_LOADING_DISABLED",
+		"LVT_TEMPLATE_BASE_DIR",
 		"LVT_SHUTDOWN_TIMEOUT",
 		"LVT_LOG_LEVEL",
 		"LVT_METRICS_ENABLED",


adnaan:
Thank you for the thorough code review! I've addressed the testing recommendation:

## Changes Made

**✅ Added integration test for `LVT_TEMPLATE_BASE_DIR` env var** (commit `a350779`)
- New test: `TestLoadEnvConfig_TemplateBaseDir`
- Verifies environment variable is loaded correctly
- Verifies it converts to option via `ToOptions()`
- Updated `clearEnv()` helper to include the new env var

## Response to Code Quality Observations

**1. Duplicate fallback logic:**

The apparent duplication is actually intentional and handles two distinct scenarios:

- **Lines 46-54**: Fallback when `runtime.Caller` fails **completely** (baseDir is still empty after trying all depths)
- **Lines 100-117**: Fallback when `runtime.Caller` **succeeds** but returns a path with no templates (e.g., Go build cache)

The second fallback also includes an optimization (line 111: `if fallbackDir != baseDir`) to avoid searching the same directory twice. This is working as designed.

**2. Environment variable validation:**

Agreed - the current approach (no strict validation) is appropriate because:
- The path might be valid at runtime but not during config loading
- The `DiscoverTemplateFiles` function will properly handle invalid paths with clear error messages
- Allows for flexible deployment scenarios

**3. Test isolation:**

The current approach with defer at the parent level is sufficient. The subtests run sequentially (not in parallel) which is appropriate since they all modify the global working directory state.

---

All tests pass ✅ and the code is ready for merge. The companion lvt PR changes have been pushed to the `add-claude-skills` branch.

Issue #303, opened 2026-03-31
WS lifecycleCtx uses context.Background() — Mount/OnConnect don't inherit request cancellation


adnaan:
## Context

Follow-up from PR #301 review (Copilot). This is a pre-existing issue amplified by the Mount-on-reconnect change.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/301

## Problem

In `handleWebSocket`, the lifecycle context is created with `context.Background()`:

```go
lifecycleCtx := NewContext(context.Background(), "", wsQueryData)
```

Now that `Mount()` runs on every WS connect/reconnect, DB work initiated by Mount won't be canceled when the client disconnects. This can leave orphaned queries running.

## Suggested Fix

Use `r.Context()` (the upgraded HTTP request context) when building the lifecycle context, so Mount/OnConnect inherit cancellation from the connection lifecycle.

---
*Follow-up from PR #301 bot review.*

PR #338, opened 2026-04-15
docs(proposals): Session 3 complete + server-push pattern lessons

adnaan:
## Summary

- Mark all 8 Session 3 checklist items complete (Loading & Progress patterns #14–#16 merged as livetemplate/examples#70, with library wiring fix shipped in #336 / v0.8.18).
- Capture the concrete lessons Session 3 surfaced in the Implementation Notes section so Session 4+ can reach the same patterns without re-iterating.

## What the new Implementation Notes cover

**Framework version requirements (both repos)** — Expand the narrow "Client version required" block to cover both the client (v0.8.23+) and the library (v0.8.18+). Document the Session.TriggerAction wiring gap and the companion ctx.GetInt/GetFloat native-type fix so a future session can diagnose "my goroutine isn't pushing" as a version problem in seconds.

**Server-push patterns (goroutine → session.TriggerAction)** — New subsection with the canonical shape:

1. Action handler ordering: re-entrancy guard → session nil-check → mutate → spawn (with the rationale for each step — the session-check-before-mutate ordering in particular is subtle and took iterations to get right).
2. Canonical goroutine cancellation via TriggerAction error return. No context.Context threading.
3. Why ephemeral state (no `lvt:"persist"` tags) is naturally self-healing on reconnect — cloneStateTyped() produces zero-value state, so patterns like ProgressBar/AsyncOps do NOT need an OnConnect recovery path. This was raised as a bug multiple times during review before I traced through mount.go and proved it.
4. Which patterns DO need OnConnect (spinner-replace shapes like LazyLoad) and the Mount guard `if ctx.Action() == ""` to prevent POSTs from resetting mid-flight state.
5. Reconnect-during-loading double-fire semantics: groupID is stable across reconnects (cookie-bound), and the framework does NOT invalidate old sessions — a false claim in an earlier version of the inline comment was caught by review.
6. Flash scoping in branched templates: flashes are ephemeral and must live in the branch that renders after the transition.

**Per-session workflow reminders** — New workflow rules distilled from Session 3:

- Library+client worktree setup when a session needs framework changes. Grep-check `newLocalSession`/`WithSession` in mount.go at the START of any server-push session to verify the framework plumbing exists before writing patterns.
- Generalized dependency-release rule: release any new library OR client version BEFORE opening the examples PR (Session 3 required shipping v0.8.18 first).
- Bot review loop convergence guidance: recognize two consecutive rounds of cosmetic-only feedback as the signal to stop pushing.
- Explicit-decline pattern: post a PR reply with rationale rather than silently ignoring a bot suggestion — prevents the same suggestion returning under different framing.
- Project guidance trumps bot suggestions when they conflict (e.g., "don't DRY a 5-line goroutine" for a patterns example).

## Test plan
- [x] File is docs-only; pre-commit hook runs go fmt + golangci-lint + full test suite — all pass locally.
- [ ] Link this PR from livetemplate/examples#70 once merged.


adnaan:
Round 1: all four items addressed in \`e74a77c\`.

1. **Stale call site count** — Replaced "5 mount call sites" with a grep-pointer (\`search for WithSession(newLocalSession to see the current set\`). Current count is 7, but any hardcoded number will drift — pointing readers at the query-of-truth is better than chasing the number.
2. **Redundant \`-r\` flag in grep** — Removed. Single-file grep does not need recursive.
3. **Non-idempotent double-fire race** — Resolved by stating the contract explicitly: every goroutine-push pattern in the examples repo MUST be idempotent by construction, verified true for Session 3 (\`DataLoaded\` just sets a string, \`UpdateProgress\` writes a monotonic value, \`FetchResult\` is a terminal transition). Future non-idempotent patterns are pointed at livetemplate/livetemplate#342 (the library-side reconnect-gap work) rather than leaving readers with a footgun and no guidance.
4. **Bot-review workflow bullets belong elsewhere** — Moved out of \`patterns.md\` (a technical design doc) and into \`livetemplate/CLAUDE.md\` as a new "AI Code Review Workflow" section alongside the existing Pre-commit Hook section. The three bullets (loop convergence, explicit decline, project-guidance-trumps-bot) are preserved verbatim there plus a fourth one about factual claims getting fact-checked. \`patterns.md\` now cross-references the new section.

The PR now touches two files: the proposal and CLAUDE.md. All four items were legitimate — good catch on the scoping one especially.

adnaan:
Round 2: four of five items addressed in \`2b92a76\`, one declined.

1. **CLAUDE.md section placement** — Moved "AI Code Review Workflow" from between Pre-commit Hook and Common Tasks to the end of the file after Future Improvements. Groups with meta/tooling content rather than sandwiched inside operational how-to-run-the-repo sections.
2. **Audience ambiguity** — Added a framing sentence distinguishing the three author-facing rules from the fourth bot-observation rule, plus inline \`*(for the PR author)*\` / \`*(reminder for both audiences)*\` tags on each bullet. Readers now immediately see which rules are for the human driving the PR and which are meta-observations about the review bot's own behavior.
3. **Mount guard comment imprecise** — Fixed. The old "GET only" comment was wrong because \`ctx.Action() == ""\` also fires on initial WS connect. The new comment reads: *"Only reset on non-action calls (initial GET or WS connect); skip when a POST action is being dispatched, so the action's own state changes aren't clobbered mid-flight."*
4. **mount.go line anchor** — Added a grep hint (\`grep -n "restorePersistedState\\|cloneStateTyped" livetemplate/mount.go\`) so a future session author can verify the self-healing path with one command instead of re-reading the whole PR description.

**Declined**: fully-qualifying \`*Context\` → \`*livetemplate.Context\` in the code examples. The entire \`patterns.md\` proposal uses the \`*Context\` shorthand (grep shows ~15 other call sites); qualifying only the new Session 3 blocks would be inconsistent with the rest of the document.

Thanks for the cross-reference verification in the "Fact-checking" section — the "7 call sites" confirmation of PR #336 validated the round-1 fix.

adnaan:
Round 3: all three items addressed in \`a5b0dfc\`. The multi-instance caveat is the biggest fix — thanks for catching that.

1. **Error string imprecise** — Corrected. The docs now say *"error with the prefix \`livetemplate: no connected sessions for group\`"* which is accurate (matches \`session_impl.go:111\`) and avoids hardcoding the groupID portion of the format string.

2. **Multi-instance cancellation caveat** — This was a real gap. The canonical \`err != nil\` pattern works only in single-instance mode because \`session_impl.go:108-112\` only returns the no-connections error when \`!hasRemote\`. Under PubSub the function returns \`nil\` and an unbounded loop would silently run forever (even during a PubSub outage, emitting warnings each iteration). Added a new bullet that:
   - Names single-instance as the mode the canonical pattern covers.
   - Documents the multi-instance silent-loop failure mode with a reference to the \`session_impl.go:75-90\` lifecycle comment.
   - Prescribes two alternatives for future patterns that need to work in both modes: (a) hard iteration bound (ProgressBar already uses this implicitly), (b) done channel or external context (for indefinite work).
   - Notes that Session 3's three patterns are all bounded by design, so none of them exhibit the bug in practice, but documents the constraint so future sessions don't copy the error-return idiom into an unbounded loop.
   - Cross-references livetemplate#303 which would eventually allow proper \`context.Context\` cancellation.

3. **Grep paths assume parent directory context** — Fixed. Both commands now say *"From the \`livetemplate/\` repo root, run \`grep -n ... mount.go\`"*. Explicit working directory removes the ambiguity.

Happy the round-2 audience framing and section move landed cleanly — those weren't strictly blocking but made the CLAUDE.md addition feel less bolted-on.

adnaan:
Round 4: all five items addressed in \`2425c08\`. Bot explicitly said "no blocking issues" — these were pre-merge durability/accuracy polish.

**patterns.md:**

1. **Hardcoded line numbers will drift** — Replaced \`session_impl.go:108-112\` and \`:75-90\` with a grep anchor (\`grep -n "hasRemote" session_impl.go\`) and a prose cross-reference to "the lifecycle comment block at the top of \`localSession.TriggerAction\`". Grep anchors are self-updating; line numbers silently drift.

2. **"Loops forever" qualifier placement** — Restructured the multi-instance caveat so the *"Session 3 patterns are all bounded by design"* reassurance appears BEFORE the mitigation options, not after. A reader skimming the warning now sees "this does not affect Session 3 patterns" immediately rather than after parsing the hard-iteration and done-channel alternatives.

3. **\`context.Context\` temporal framing** — Added *"as of \`livetemplate v0.8.18\`"* qualifier to the "does not thread a cancellable request context through \`Session\`" claim. A future reader can now tell whether the behavior was observed at the time of writing vs. persists in a later release.

**CLAUDE.md:**

4. **Dissonant "human PR author (Claude Code, the agent)" framing** — Replaced with *"Rules 1–3 apply to whoever is driving the PR (human or agent). Rule 4 is a reminder about the review bot's own behavior."* The previous phrasing was literally contradictory — Claude Code is not a human — and conflated the audience.

5. **"Fact-checks" accuracy nit** — Softened to *"The bot attempts to verify … and has caught genuine drift in past reviews. Its verifications are not infallible (it can hallucinate line numbers and occasionally miss a call site)"*. Also promoted the concrete takeaway from this round into a bold rule: **"Prefer grep anchors over hardcoded line numbers in code references"** — grep anchors are self-updating, line numbers silently drift.

The grep-anchor rule in particular is meta-valuable since this round's #1 was itself a drift risk that the rule would have prevented.

adnaan:
Round 5 addressed in \`65851c4\`. All three items were legitimate; #3 (OnConnect race comment) was the most substantive.

1. **OnConnect \`!state.Loading\` guard scope** — Replaced the terse \`// Skip re-spawn after completion\` with a multi-line comment explicitly stating what the guard does NOT prevent: a second goroutine can still spawn during a mid-sleep reconnect because \`state.Loading\` is still true at that point. The pattern is safe ONLY because DataLoaded is idempotent. This makes the idempotency requirement legible at the exact code location where the race can manifest, cross-referencing the "Reconnect-during-loading double-fire" section for the full analysis. Good catch.

2. **CLAUDE.md self-referential framing** — Added a framing sentence to the section intro: *"The rules below are maintainer reflections based on observed behavior across multiple sessions of patterns work — not generated self-assessment by the bot itself."* Addresses the bootstrap concern about the bot documenting its own behavior inside a file the bot reads.

3. **CLAUDE.md Rule 1 over-prescriptive threshold** — Partial accept. Rewrote to lead with the underlying criterion *("successive rounds aren't identifying any new functional issue")* while keeping the "two consecutive rounds" heuristic as a calibration anchor afterward. Also added: *"address remaining cosmetic nits by reply rather than by push"* so authors don't weaponize the rule to dismiss substantive feedback. Declined the exact suggested phrasing because the count remains a useful applicability anchor — the new wording combines both criteria.

---

Per the (just-updated) Rule 1: rounds 4 and 5 have both explicitly said "no blocking issues," and neither identified a new functional issue — they've been style/wording refinements on already-accepted content. This is the convergence signal. Unless round 6 surfaces a new functional issue, I'll address any further cosmetic nits by reply rather than pushing another commit.

adnaan:
Round 6: legitimate new functional issue. Fixed in \`b36a18d\`.

**Canonical shape teaches the dangerous form** — this was a real ordering bug in the doc. The previous "Action handler shape" code block used \`for { /* ... */ }\` as the canonical loop, with the multi-instance caveat sitting three bullets below. A reader copy-pasting the example would naturally write an unbounded loop before they ever reached the safety warning.

The rewrite:

1. **Uses a bounded loop by default** (\`for i := 0; i < maxTicks; i++\`) matching Session 3's \`ProgressBarController.Start\`. The canonical example is now correct in both single-instance and multi-instance modes by construction.
2. **Inline constraint comment at the \`go func()\` site**: *"Spawn with a BOUNDED loop (critical — see multi-instance caveat below). Unbounded \`for { ... }\` loops using only \`err != nil\` as the exit condition will silently run forever under PubSub, because TriggerAction returns nil when a broadcaster is configured. Always bound the iteration count, OR use a done channel — never rely on the error return as the sole stop signal in an unbounded loop."* This is legible at the exact location where a reader might be tempted to write \`for { ... }\`.
3. **Surrounding prose frames the example** as "intentionally bounded so it is safe in both single-instance and multi-instance deployments."
4. **The single-instance comment on the \`return\`** explicitly tags that cancellation semantic with its mode scope.

Good catch — this was the right kind of "teach by example" problem to flag, and explicitly matches the criterion I stated in CLAUDE.md Rule 1 for when a new round warrants a push vs. a reply (new functional issue, not cosmetic). Back to the convergence watch: if round 7 raises no new functional issue, I will merge.

adnaan:
Round 7 addressed in \`450e567\`. Bot's own summary said "one actionable fix"; I took it plus one minor consistency item.

1. **\`#303\` framing** — Corrected "open fix" → "open tracking issue". #303 is an issue describing the problem, not a PR with a pending code change. The new phrasing is accurate without implying work in flight.

2. **ProgressBar bound description drift** — Rewrote the prose to stop hardcoding \`i <= 100\` (which was in the prose but not in the pseudocode example using \`i < maxTicks\`). Now points at the actual file with a name-based anchor: *"ProgressBar's loop is bounded (iterates from \`progressStep\` to \`100\` in steps of \`progressStep\` — see \`patterns/handlers_loading.go\` in the examples repo; total run is 10 × 500ms = 5s max)"*. Kept the "10 × 500ms = 5s max" arithmetic since that's the load-bearing constraint; dropped the specific literal that could drift.

**Declined**: the "OnConnect ellipsis one-line comment" suggestion. The \`go func() { /* ... TriggerAction("dataLoaded", ...) ... */ }()\` ellipsis is already understood as pointing at the Mount-path goroutine body from the surrounding prose. Adding a cross-reference comment would be noise for a docs sketch.

Per CLAUDE.md Rule 1: rounds 4, 5, and 7 have been "no blocking issues" with only style/accuracy polish; round 6 was the one substantive catch. If round 8 is clean I will merge without further discussion.

adnaan:
Round 8: all three items addressed in \`73ec8ea\`. Item #1 was a genuine factual error — I invented the \`noopSession\` type name in a section that was itself supposed to be grep-verifiable. Exactly the drift the "prefer grep anchors" rule was designed to catch, and this time it caught me.

1. **\`noopSession\` doesn't exist** — Removed. The real story is that \`c.session\` was left as a nil Session interface value because \`WithSession(...)\` was never called at any lifecycle call site. Rewrote the paragraph to describe the actual absence-of-wiring shape, including both possible failure modes (silent no-op or nil-dispatch panic depending on call path). A future reader grepping for \`noopSession\` would have found nothing and wasted time; the new phrasing is verifiable against the current code.

2. **Grep anchor tightened** — Changed \`grep -n "newLocalSession\\|WithSession" mount.go\` to \`grep -n "WithSession(newLocalSession" mount.go\`. Identical output today, but the compound anchor excludes hypothetical future \`WithSession()\` calls that don't wrap \`newLocalSession\` (e.g., a mock session in a test helper). More drift-proof.

3. **CLAUDE.md framing honesty** — The previous "not generated self-assessment by the bot itself" was slightly strained because this section WAS authored with Claude Code assistance. Rewrote to: *"distilled from observed patterns across multiple PR review cycles (including the one that produced this section) and reviewed by the maintainer before landing. They are not the bot's own self-assessment, but they were written with Claude Code assistance — so treat them as operational heuristics to test against your own observations, not as a formal specification of the bot's behavior."* Honest about the authorship while still distinguishing maintainer-reviewed heuristics from raw bot output.

---

The \`noopSession\` finding is notable meta-evidence: even with the "prefer grep anchors" rule in the same PR, I still slipped a fabricated type name into the prose. The rule caught it on review — that's the system working correctly.

Scoreboard: rounds 1, 6, and 8 had genuine functional/factual catches that warranted pushes. Rounds 2, 3, 4, 5, 7 were polish. Per CLAUDE.md Rule 1, I'll merge if round 9 is clean.

adnaan:
Round 9 addressed in \`2aa2fb3\`. Bot explicitly flagged #1 and #2 as "worth fixing before merge"; both were real.

1. **\`session_impl.go\` anchor pointed at the wrong comment block** — Fixed. The old phrasing said "the lifecycle comment block at the top of \`localSession.TriggerAction\`" but the top of the function (lines 92-99) is the nil-safety/wiring guarantees block, not the multi-instance contract the surrounding prose was discussing. The full multi-instance lifecycle contract (Disconnect semantics + bounded-loop workaround) lives in the doc comment PRECEDING the function definition. Replaced the location pointer with a grep anchor: \`grep -n "Disconnect semantics" session_impl.go\`. This is the eighth or ninth time a grep anchor beat a location reference in this PR — the rule keeps earning its keep.

2. **CLAUDE.md Rule 3 examples are examples-repo-scoped but the rule lives in library CLAUDE.md** — Legitimate scoping bug. A library PR author touching \`internal/\` or \`mount.go\` could cite "duplicated goroutines are pedagogical, not DRY violations" to dismiss a genuinely correct bot suggestion. Added an explicit guard:

   > *"Important: the 'decline' examples below apply specifically to examples-repo patterns code — they are NOT license to dismiss DRY or const-declaration suggestions in library (\`internal/\`, \`mount.go\`, and other core library) PRs, where the bot's defaults are usually correct."*

   Also qualified each example bullet: **"in patterns examples"** and **"in a single examples file"**. Rule 3 now can't be weaponized against legitimate library feedback.

3. **External-repo path reference** — Low-priority but fixed. Dropped the specific \`patterns/handlers_loading.go\` path and replaced with *"grep \`progressStep\` in the examples repo's loading handlers"* — name-based, self-updating, doesn't pin a specific filename that could be renamed.

**Declined**: extracting the 92-line server-push section into a separate \`docs/references/server-push-patterns.md\` file. Bot itself flagged it as "no action required now"; since this doc is for session authors who already have \`patterns.md\` open as their index, inlining avoids extra indirection. A future refactor can split when the file becomes genuinely unwieldy.

**Rule 3's over-broad scoping was itself a good catch** — exactly the kind of "bot trained to make this mistake" risk Rule 3 is supposed to prevent, now with a guard to prevent the rule from being misused in the inverse direction.

adnaan:
Round 10: one legitimate clarification, one declined cosmetic churn item.

1. **\`r.Context()\` vs controller-stored context ambiguity** — Fixed in \`987f4aa\`. The previous phrasing could be misread as "never use context.Context in a goroutine-push pattern," which would lead a reader to write an unbounded loop with no done-channel. The real rule is narrower: don't thread the HTTP request context through Session. A **controller-stored** \`context.Context\` or \`done chan struct{}\` (created in the action handler, closed in OnDisconnect) IS the correct pattern and is shown verbatim in the doc comment preceding \`localSession.TriggerAction\`. Added an explicit cross-reference and an unambiguous "IS correct" statement.

2. **CLAUDE.md section placement under Development Conventions** — **Declining** with rationale. The bot has now suggested three different placements for this section across three rounds:
   - Round 1 original: between Pre-commit Hook and Common Tasks
   - Round 2 bot: "move to the end of the file after Future Improvements" → I moved it
   - Round 10 bot: "should live under Development Conventions"
   
   Bot itself says "Not blocking … worth filing as a follow-up." This is exactly the pattern CLAUDE.md Rule 1 was written to stop — multi-round reworking of the same cosmetic organizational decision without identifying a new functional issue. Current placement (end of file after Future Improvements) was sanctioned by round 2 review and has been stable since round 4. I'm going to hold the current placement and treat further re-placement suggestions as below the convergence threshold.

---

Scoreboard heading into round 11:

- Rounds **1, 6, 8, 9** — genuine substantive catches (fabricated types, unbounded loop shape, wrong anchor, rule over-scoping)
- Rounds **2, 3, 4, 5, 7, 10** — accuracy polish / clarifications (rounds 3 and 10 both found legitimate near-functional clarifications; the others were drift-prevention)

Total: 10 rounds, 4 real catches, 6 polish rounds. Cosmetic-churn items (like CLAUDE.md placement iteration) start mixing with real catches from round 10 onward — this is the signal that each new round is approaching zero expected yield. I'll merge if round 11 is clean.

adnaan:
Round 11 addressed in \`085a7a6\`. Bot verdict was "Approve with minor suggestions" — two items accepted, one minor nit declined, one placement nit declined per Rule 1.

1. **\`!state.Loading\` guard unreachable in normal operation** — Real catch. Mount unconditionally resets \`Loading=true\` when \`ctx.Action()==""\`, and Mount's output flows into OnConnect, so the \`!state.Loading\` check in OnConnect is unreachable with the current Mount shape. My previous comment ("only skips re-spawn AFTER completion") was misleading because completion-then-reconnect goes through Mount first and \`Loading\` gets re-set to true before OnConnect sees it.

   Rewrote the inline comment to be honest: the guard is **defensive**, unreachable with the current Mount shape, exists for forward-compat with future Mount changes that might preserve already-completed state (e.g., adding \`&& state.Data == ""\` to the reset condition). Also split the "defensive guard" rationale from the "reconnect race idempotency" rationale into two distinct comment blocks — they were conflated previously, which the bot's catch helped tease apart.

2. **Mount + OnConnect in one fenced code block** — Split into two separate blocks with explicit **Mount:** / **OnConnect:** headers and a one-line lifecycle order description. Reader no longer has to track curly-brace depth to find the function boundary.

3. **Drop \`(10 × 500ms = 5s max)\` arithmetic parenthetical** — Accepted. The grep instruction is now the sole anchor for the ProgressBar bound; dropping the calculation eliminates drift risk if \`progressStep\` or \`tickRate\` change in the examples repo.

**Declined**: CLAUDE.md Rule 3 caveat placement. This is the third placement-related suggestion for the CLAUDE.md section across rounds 2, 10, and 11. Per Rule 1, multi-round reworking of the same cosmetic organizational decision without a new functional issue is the exact churn pattern the rule is designed to stop. Current placement has been stable since round 9 and is clear enough.

---

**Notable**: Round 11's #1 was a genuinely surprising catch. I'd looked at that guard multiple times across earlier rounds and never noticed that the Mount reset made it unreachable. The "two separate comment blocks" restructure that came out of the fix is arguably clearer than what I had before — the reconnect race and the defensive-guard rationale were conflated, and pulling them apart makes each more pointed.

Scoreboard heading into round 12: 5 substantive catches (rounds 1, 6, 8, 9, 11) + 1 near-functional clarification (round 3 & 10) across 11 rounds. The yield curve is not monotonically decreasing — genuine catches still appearing. Will continue the loop.

adnaan:
Round 12 addressed in \`5be3283\`. Two real items accepted, one placement nit declined.

1. **"Bounded = multi-instance safe" rationale sharpened** — Good catch. The previous text left the reason implicit, and a reader could misread "bounded" to include timeout-bounded loops like \`for time.Since(start) < 30*time.Second\`, which are NOT safe (under a wedged scheduler or clock-skew, time.Since can drift and the loop runs well past the intended wall-clock bound). Added an explicit sentence: the bound is safe because \`maxTicks\` is a **fixed integer** (not a wall-clock duration), and in multi-instance mode \`TriggerAction\` never returns an error, so only the iteration count stops the loop. Named the specific anti-pattern (\`for time.Since(start)...\`) so readers learn the distinction by example.

2. **Context provenance in the \`session_impl.go\` example** — Also real. A reader following the grep anchor to \`Disconnect semantics\` sees code using \`ctx.Done()\` with no annotation of where \`ctx\` comes from. They could reasonably infer it's \`r.Context()\` and defeat the bounded-lifetime intent. Added a clarifying note in the proposal: *"the \`ctx\` variable in that example is a controller-stored context (created and owned by the caller's controller), not \`r.Context()\`. The example is illustrative of the \`select { case <-ctx.Done(): ... }\` loop structure; when copying it, create the context in your action handler and close it on OnDisconnect, not via \`r.Context()\`."* Kept the fix in the proposal rather than touching \`session_impl.go\` itself since this is a docs-only PR.

**Declined**: CLAUDE.md section placement. This is the **fourth** placement suggestion across rounds 2, 10, 11, and 12. Bot itself says "Not a blocking concern — leaving this to maintainer judgment." Per Rule 1, multi-round churning on the same organizational decision without a new functional issue is exactly what the rule is designed to stop. Holding current placement.

---

Scoreboard through 12 rounds:

- **Substantive catches**: 1 (unbounded loop), 6 (bounded shape), 8 (fabricated type), 9 (rule over-scoping), 11 (unreachable guard), 12 (bounded rationale + context provenance) — 6 of 12
- **Near-functional clarifications**: 3 (multi-instance caveat), 10 (r.Context vs controller context) — 2 of 12
- **Pure polish**: 2, 4, 5, 7 — 4 of 12

The yield curve is flatter than I thought — round 12 produced two real catches, bringing the substantive-catch rate to ~50%. Rule 1's "two consecutive cosmetic rounds" hasn't been triggered since round 10 had one cosmetic item alongside a real one. Continuing the loop.

adnaan:
Round 13 addressed in \`e8349ff\`. Four items accepted, one placement nit declined. The time.Since correction (#2) was especially valuable — round 12's bot guidance and my acceptance of it produced an overclaim, and round 13 caught it.

1. **OnConnect example body filled in** — Replaced the pseudocode placeholder \`go func() { /* ... TriggerAction("dataLoaded", ...) ... */ }()\` with the full goroutine body matching Session 3's actual lazy-load shape (\`time.Sleep + TriggerAction + err return\`). Start and OnConnect examples are now both compilable-as-is shapes, no more inconsistency.

2. **time.Since overclaim corrected** — This one deserves an explicit "I got it wrong" note: in round 12, I accepted a sharpening suggestion and wrote that *"a timeout-bounded loop \`for time.Since(start) < 30*time.Second\` is NOT safe... time.Since can drift, and under a wedged scheduler..."*. That was wrong. \`time.Since\` is monotonic and CAN stop a loop directly via the condition check. The actual narrower rule is: \`if err != nil { return }\` inside the loop body cannot be the exit condition under PubSub. Rewrote the passage to focus on *"the exit condition must be something the loop body controls directly, not something that depends on TriggerAction's return value"* — no more unsafe-timeout-loop claim. Good catch; round 12's framing had eroded the accuracy.

3. **Compressed the context-provenance paragraph** — The round-12 "Note on that example" block was dense. Folded the controller-stored-context constraint directly into the sentence with the grep pointer, cutting redundancy without losing the safety point.

4. **CLAUDE.md preamble hedge dropped** — The "treat them as operational heuristics… not a formal specification" caveat was added in round 8 to address the meta-concern of the bot documenting its own behavior. Round 13 correctly flagged that it invites readers to discount the rules before reading them. Since these rules have been maintainer-reviewed AND battle-tested across 12 prior rounds of this exact PR, they deserve direct application. New preamble: *"distilled from observed patterns across multiple PR review cycles and reviewed by the maintainer before landing — apply them directly when working on PRs."*

**Declined**: Rule 3 scope caveat placement. This is the fifth placement-related suggestion for the CLAUDE.md section (rounds 2, 10, 11, 12, 13). Per Rule 1, multi-round churning on the same organizational decision is exactly what the rule stops.

---

Round 13 was another productive round — 3 of 4 accepted items were substantive (OnConnect body, time.Since correction, hedge drop). The time.Since fix in particular is the kind of thing I wouldn't have noticed without the loop, because I already accepted it as "correct" in round 12. The bot's willingness to push back on its own earlier framing (given this PR also has the rule "project guidance trumps bot suggestions") is the opposite failure mode and equally important.

Updated scoreboard: substantive-catch rate is now ~55% (rounds 1, 3, 6, 8, 9, 10, 11, 12, 13 all had real items) across 13 rounds. Continuing.

adnaan:
Round 14 addressed in \`73d0262\`. Three items accepted, two declined per Rule 1.

1. **OnConnect guard framing flip** — Good catch. The previous "Defensive guard ... exists for forward-compat" framing invited a future Mount refactor to remove the check as optional cruft. Flipped to "Load-bearing safety net": same code, but now framed as an active safety net for a specific future Mount change (preserving already-completed state). Added an explicit **"Do NOT remove when refactoring Mount"** instruction at the end of the comment. The guard's purpose is now unambiguous.

2. **progressStep cross-repo reference made explicit** — Previously said "grep \`progressStep\` in the examples repo's loading handlers" which mentioned the cross-repo location but buried it in the verification instruction. Rewrote to lead with *"the constants live in the **examples repo** at \`patterns/handlers_loading.go\`"* and explicitly note *"which is not visible from this repo"* so readers don't waste time searching the current repo for the bound.

3. **\`ctx.GetInt\`/\`ctx.GetFloat\` grep anchor added** — Real drift-protection addition. The library version block asserted the numeric-type fix without letting readers verify it in their version. Added \`grep -n "int8\\|int16\\|int32\\|int64" context.go\` which matches the type-switch branches in \`GetIntOk\`/\`GetFloatOk\`. Consistent with the grep-anchor-over-assertions rule from CLAUDE.md.

**Declined**:

- **#1 Rule 3 scope caveat placement** — Sixth placement-related suggestion for the CLAUDE.md section (rounds 2, 10, 11, 12, 13, 14). The current "Important: NOT license to dismiss..." callout is already at the start of the rule; further re-arrangement is pure churn. Rule 1.

- **#2 Rule 4 audience framing ("human contributors can skip it")** — Current text already says "Rule 4 is a reminder about the review bot's own behavior" which distinguishes it from author-facing rules. Adding "humans can skip it" is a rephrasing of round-5's framing decision without a new functional issue. Same churn pattern.

---

Scoreboard after 14 rounds: substantive catches in rounds 1, 3, 6, 8, 9, 10, 11, 12, 13, 14 (10 of 14). The pattern holds — this review loop is genuinely finding real improvements and the two declined items per round are visibly the cosmetic-churn variety that Rule 1 targets. Will continue.

adnaan:
Round 15 addressed in \`86f7f6d\`. One real factual error caught, one minor header-framing nit accepted.

1. **Wrong grep anchor file (confirmed drift I introduced in round 14)** — The previous text said \`grep -n "int8\\|int16\\|int32\\|int64" context.go\` but the numeric type switch actually lives in \`action.go\` (on \`ActionData.GetIntOk\` / \`GetFloatOk\`), not \`context.go\` (which only has thin wrappers delegating via \`c.data.GetInt(key)\`). Running the grep against \`context.go\` returned nothing useful. Corrected to \`grep -n "int8\\|int16\\|int32\\|int64" action.go\` and clarified the two-layer delegation chain inline (\`ctx.GetInt\` in \`context.go\` → \`ActionData.GetIntOk\` in \`action.go\`) so future readers see the shape rather than conflating the layers.

   **Meta-irony**: this is the second time the "prefer grep anchors over hardcoded references" rule has caught ME in this PR — first the fabricated \`noopSession\` type in round 8, now a grep anchor pointing at the wrong file in round 14/15. Both times, I introduced drift while promoting a rule designed to prevent exactly that drift, and the next round's fact-check caught it. The rule is load-bearing precisely BECAUSE authors (even authors who wrote the rule) don't notice their own drift. Round 15's catch isn't just a good fix; it's empirical evidence the review loop has value.

2. **Multi-instance caveat bullet header framing** — Changed the bold header from the failure mode ("silently loops forever under PubSub") to the rule ("indefinite-work loops MUST use a hard bound or done channel, not the error-return pattern"). Readers skimming the bullet now see the primary rule first; the failure mode lives in the explanatory prose below.

---

Scoreboard through 15 rounds: substantive catches in rounds 1, 3, 6, 8, 9, 10, 11, 12, 13, 14, 15 (11 of 15, ~73%). Pure polish in rounds 2, 4, 5, 7 (4 of 15). Rule 1's convergence signal still not triggered. The yield curve is genuinely flat — this is not the usual "bot loop tail."

adnaan:
Round 16: **no new functional issue; the fix is already in place**. Not pushing another commit.

The bot's "factual error: wrong grep target file" is a re-raise of the exact issue fixed in round 15 (commit \`86f7f6d\`, currently HEAD of this PR). The current text reads:

> The delegation chain is \`ctx.GetInt()\` (in \`context.go\`, a thin wrapper) → \`ActionData.GetIntOk()\` (in \`action.go\`, the actual type switch); verify the numeric-type switch is present in your version with **\`grep -n "int8\\|int16\\|int32\\|int64" action.go\`** (it should match the type-switch branches in \`GetIntOk\`/\`GetFloatOk\`).

The grep anchor explicitly targets \`action.go\`, not \`context.go\`. I suspect the bot's fuzzy matcher latched onto the word \`context.go\` appearing earlier in the delegation-chain description and re-raised round 15's issue without re-verifying against the current file.

Per CLAUDE.md Rule 1: "successive rounds aren't identifying any new functional issue" is the convergence signal. Round 16 has no new issue — it's hallucinating a regression on a line that's already correct. The PR has converged.

---

**Final scoreboard after 16 rounds:**

| Round | Outcome |
|---|---|
| 1 | Substantive (unbounded loop canonical shape) |
| 2 | Polish |
| 3 | Substantive (multi-instance caveat) |
| 4 | Polish |
| 5 | Polish |
| 6 | Substantive (bounded shape teaching) |
| 7 | Polish |
| 8 | Substantive (fabricated \`noopSession\`) |
| 9 | Substantive (rule over-scoping) |
| 10 | Substantive (r.Context clarification) |
| 11 | Substantive (unreachable guard) |
| 12 | Substantive (bounded rationale + context provenance) |
| 13 | Substantive (OnConnect body + time.Since overclaim) |
| 14 | Substantive (guard framing + GetInt anchor added) |
| 15 | Substantive (wrong grep file caught) |
| 16 | **False positive on already-fixed issue — convergence** |

11 of 16 rounds produced real catches. Rule 1 now triggers. Moving to merge.

Issue #343, opened 2026-04-16
flash: update examples and docs for persist-until-cleared lifecycle


adnaan:
## Context

PR `feat/navigate-action` changes flash message lifecycle from **auto-clear-after-render** to **persist-until-cleared** (or until optional expiry). This aligns with Phoenix LiveView's `put_flash` / `clear_flash` model and fixes a class of bugs where background `TriggerAction` refreshes would clobber flash messages set by user actions within one scan interval.

### New API

```go
ctx.SetFlash("error", "Something went wrong")                         // persists until ClearFlash
ctx.SetFlash("success", "Saved!", livetemplate.FlashExpiry(5*time.Second))  // auto-expires
ctx.ClearFlash("error")                                                // explicit removal
```

### Breaking change

Existing code that relies on flash auto-clearing after one render will see flash messages **persist indefinitely**. Each call site needs one of:
1. Add `FlashExpiry(duration)` for auto-dismissal (most common fix)
2. Add explicit `ClearFlash(key)` in a follow-up action handler
3. Add client-side JS dismiss (for interactive close buttons)

## Impacted code

Based on a full workspace scan:

### livetemplate server (`livetemplate/livetemplate`)
- `mount.go`: 7 `clearFlash()` → `pruneExpiredFlash()` call sites (already done)
- `context.go`: `SetFlash` signature updated (already done)
- `docs/references/error-handling.md` lines 433-545: flash lifecycle docs need update
- `docs/references/api-reference.md` lines 231-237: flash API section needs `ClearFlash` + `FlashExpiry`
- `lifecycle_integration_test.go` lines 169, 178: flash test may need adjustment for new persist behavior

### Examples (`livetemplate/examples`)
- `flash-messages/main.go`: 6 SetFlash calls — add `FlashExpiry(5*time.Second)` for demo UX
- `flash-messages/flash_test.go`: test "ClearsAfterAction" needs rework — flash no longer auto-clears
- `flash-messages/README.md`: lifecycle documentation update
- `patterns/handlers_loading.go`: 3 SetFlash calls (progress bar, async ops)
- `patterns/handlers_forms.go`: 5 SetFlash calls (bulk update, file upload, preserve inputs)
- `ws-disabled/ws-disabled.tmpl`: flash template usage
- `avatar-upload/avatar-upload.tmpl`: flash template usage
- `progressive-enhancement/progressive-enhancement.tmpl`: flash template usage

### lvt CLI (`livetemplate/lvt`)
[...]

PR #344, opened 2026-04-16
feat: __navigate__ action + flash persist-until-cleared lifecycle

adnaan:
## Summary

- **\`__navigate__\` action**: reserved action name that re-runs Mount on the existing WebSocket with fresh query data. Enables same-handler SPA navigation without reconnecting — the in-band equivalent of Phoenix LiveView's \`live_patch\` / \`handle_params\`.
- **Flash lifecycle refactor**: flash messages persist until explicitly cleared via \`ClearFlash(key)\` or until optional \`FlashExpiry(duration)\` elapses. Replaces the auto-clear-after-render behavior that caused flash messages set by user actions to be clobbered by the next background Refresh tick.

## Breaking Changes

**Flash lifecycle (WebSocket connections only):** In v0.8.x, \`ctx.SetFlash(key, msg)\` was inherently one-shot — flash was auto-cleared after the next render. In v0.9, flash persists on WebSocket connections until \`ctx.ClearFlash(key)\` is explicitly called or \`FlashExpiry(duration)\` elapses. Existing handlers that set flash and rely on auto-clear will accumulate flash indefinitely unless updated. Migration steps are documented in the \`SetFlash\` godoc. HTTP connections (PRG forms) remain one-shot as before.

**\`ctx.Action() == ""\` guard fires on navigate:** The \`__navigate__\` action uses \`actionCtx.WithAction("")\` so that Mount's \`if ctx.Action() == "" { ... }\` guard fires on navigate — this is intentional (navigate is conceptually a re-mount). Existing controllers using this guard for analytics or side-effects (e.g., \`trackPageView()\`) will fire on every navigate, which may be unexpected. Move initial-connect-only side-effects to \`OnConnect\`.

**\`__navigate__\` with no/empty data replaces (not merges) query params:** Sending \`__navigate__\` with a nil or empty data field passes an empty map to Mount — the original WebSocket connection's query string is NOT preserved. All \`ctx.GetString\` / \`ctx.GetInt\` calls return zero values. There is no implicit merge; callers must explicitly include all params they want Mount to see.

## Test plan

- [x] \`navigate_test.go\`: 5 tests — Mount re-runs with new data, routing bypass, non-navigate routes through DispatchWithState, empty-data resets params, error-path state unchanged, no-Mount is no-op
- [x] \`flash_lifecycle_test.go\`: 8 tests — persist, explicit clear, expiry (unit), selective prune, expiry-overwrite, WS action integration, navigate path integration, FlashExpiry public-API integration
- [x] Full test suite (70s including Redis integration) — green
- [x] Flash lifecycle tested end-to-end via devbox-dash chromedp: \`TestFlashSurvivesScanRefresh\` passes
- [ ] Deferred to #343: dispatched-action flash path (\`handleDispatchedAction\`) and pubsub broadcast flash path (\`handleServerActionMessage\`) — both call \`pruneExpiredFlash()\` correctly but lack dedicated flash regression tests; covered indirectly by the existing broadcast/pubsub test suites

## Follow-up

- livetemplate/livetemplate#343: update examples, docs, and migration guide for the new flash lifecycle and navigate behavior

### DIFF
--- a/.github/workflows/cross-repo-test.yml
+++ b/.github/workflows/cross-repo-test.yml
@@ -61,7 +61,7 @@ jobs:
         npm ci
         echo "Building client library..."
         npm run build
-        echo "✓ Client library built successfully"
+        echo "OK: Client library built successfully"
         ls -la dist/
 
     - name: Run core library unit tests (with race detector)
@@ -88,7 +88,7 @@ jobs:
         echo "Adding replace directive to use local core library..."
         go mod edit -replace=github.com/livetemplate/livetemplate=../livetemplate
         go mod tidy
-        echo "✓ Replace directive added"
+        echo "OK: Replace directive added"
 
     - name: Run LVT unit tests
       working-directory: lvt
@@ -153,7 +153,7 @@ jobs:
         npm ci
         echo "Building client library..."
         npm run build
-        echo "✓ Client library built successfully"
+        echo "OK: Client library built successfully"
         ls -la dist/
 
     - name: Install sqlc
@@ -168,7 +168,7 @@ jobs:
         go mod edit -replace=github.com/livetemplate/livetemplate=../livetemplate
         go mod edit -replace=github.com/livetemplate/lvt/components=../lvt/components
         go mod tidy
-        echo "✓ Replace directives added"
+        echo "OK: Replace directives added"
 
     - name: Run examples test suite
       working-directory: examples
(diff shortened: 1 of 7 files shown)


adnaan:
Re-triggering CI after examples fix: examples repo updated with `ClearFlash` calls + go.mod bumped to v0.8.19 pseudo-version.

PR #350, opened 2026-04-18
feat: in-band navigate action + flash persistence

adnaan:
## Summary

- **`__navigate__` action**: When the client sends `{action: "__navigate__", data: {s: "abc"}}` over an existing WebSocket, the server re-runs `Mount()` with the new query params — no reconnect needed. This is the server-side counterpart to the client's link interception (already shipping in `@livetemplate/client` v0.8.27). Equivalent to Phoenix LiveView's `live_patch` / `handle_params`.
- **Flash persistence**: Flash messages now persist until explicitly cleared via `ctx.ClearFlash(key)` or auto-dismissed via `livetemplate.FlashExpiry(duration)`. Previously, background `TriggerAction` refreshes (scan loops, PubSub) would wipe flash within one cycle (~2s), making them unreadable.

## Motivation

Without server-side `__navigate__` support, the client intercepts `<a>` clicks on same-pathname links and sends the action over WebSocket, but the server silently drops it — the click is lost. The only workaround is `lvt-nav:no-intercept` on every link, which forces full page loads and defeats live navigation.

## Test plan

- [x] `TestNavigateActionReMountsWithNewQueryData` — verifies query param changes propagate through Mount on the same WebSocket
- [x] `TestNavigateActionNotRoutedThroughDispatchWithState` — verifies no method-name collision; navigate is intercepted before dispatch
- [x] Full test suite passes (66s, all green)

### DIFF
--- a/action.go
+++ b/action.go
@@ -40,6 +40,27 @@ type message = send.ActionMessage
 // Maps to the Submit() method on the controller via methodNameToActions().
 const defaultFormAction = "submit"
 
+// actionNavigate is a reserved action name. When the client sends
+// {action: "__navigate__", data: {k: v, ...}} over an existing WebSocket
+// connection, the event loop re-runs Mount on the controller with the
+// given data as query params — without reconnecting the WebSocket. This
+// is the in-band equivalent of Phoenix LiveView's live_patch /
+// handle_params: query-param changes on the SAME handler don't tear down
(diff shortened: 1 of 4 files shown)


adnaan:
Closing — this was already squash-merged via PR #344 (d09160e). Duplicate.

PR #359, opened 2026-04-25
fix: prune expired flash before render (not after sendUpdate)

adnaan:
## Summary

- `FlashExpiry(d)` documented the message is pruned "on the next render that triggers" — but `pruneExpiredFlash` actually ran *after* each successful `sendUpdate`, so an expired flash got one extra render before being evicted from `connState`. Users observed flash that "stayed one extra interaction past the 5-second deadline."
- Move the prune into `getMessages` — the single funnel all template renderers go through — so the action's own render reflects the expiry. Remove the four post-`sendUpdate` calls (WS, broadcast, pubsub group, upload completion).
- Side effect: error renders now also prune. The old code guarded the WS prune with `actionErr == nil`. New behavior: an explicit timeout fires whether or not the next action succeeds — matches user intent for a deadline.

Surfaced while implementing the `Flash Messages / Toasts` pattern in livetemplate/examples Session 5: a 5s auto-expire demo couldn't deliver on its pitch — the next click after expiry still rendered the success flash one final time.

## Test plan

- [x] `TestExpiredFlashAbsentFromGetMessagesSnapshot` (new regression test) — verifies `getMessages()` prunes before snapshotting
- [x] Existing flash tests (`TestFlashPersistsAcrossRenders`, `TestFlashExpiryPrunesAfterDeadline`, `TestFlashSurvivesSubsequentWSAction`, `TestFlashSurvivesNavigateAction`, `TestFlashExpiryThroughPublicAPI`, `TestClearFlashThroughPublicAPI`, etc.) — all pass
- [x] Library full suite: `go test ./...` — green
- [x] Library race: `go test -race -run TestFlash .` — green
- [ ] After release: bump `go.mod` in `livetemplate/examples` Session 5 PR and re-verify the demo's "Save → wait 5s → success disappears on first interaction" works as advertised

### DIFF
--- a/context.go
+++ b/context.go
@@ -308,8 +308,9 @@ type flashConfig struct {
 }
 
 // FlashExpiry sets an auto-expiry duration for the flash message. After
-// the duration elapses, the message is pruned on the next action or server
-// push that triggers a render — there is no background timer. Use this for
+// the duration elapses, the message is pruned on the next render that walks
+// flash state — there is no background timer. The render that the user
+// triggers post-expiry is rendered without the expired flash. Use this for
 // transient feedback ("Settings saved") that doesn't need explicit
 // acknowledgement. Messages without an expiry persist until explicitly
 // cleared via ClearFlash.
@@ -320,10 +321,6 @@ type flashConfig struct {
 // Note: FlashExpiry has no observable effect on HTTP connections — HTTP
 // flash is inherently one-shot (per-request connSt is GC'd after the
 // handler returns) regardless of the expiry duration set here.
-//
-// Note: The expiry timer is only checked before successful renders. If
-// a connection receives only error responses after the expiry elapses,
-// the expired flash remains visible until the next successful render.
 //
 // Example:
 //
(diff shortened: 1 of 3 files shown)


adnaan:
The comment trims you flagged were already landed in commit [a2e4639b](https://github.com/livetemplate/livetemplate/commit/a2e4639b) (pushed before this review fired). Current file state:

- `getMessages()` comment: `// Prune before snapshot so renderers never observe expired entries.` (your suggested form)
- `transientFlashExpiry` const: `// Long enough for step 1's render to observe the flash as live; short enough\n// for subsequent steps to sleep past it cheaply.` (your suggested form)
- `TestExpiredFlashAbsentFromGetMessagesSnapshot` doc: trimmed to "verifies that getMessages prunes expired flash before returning a snapshot, so renderers never see an entry whose deadline has passed."

The review appears to have regenerated against an earlier commit. The substance of the convergence is: rounds 2 and 3 produced no new functional issues, and the cosmetic feedback they did surface was addressed in a2e4639b. Treating this as the convergence point per `CLAUDE.md` AI review loop guidance.

PR #360, opened 2026-04-28
docs(proposals): streaming range rendering

adnaan:
## Summary

- Proposes replacing per-item retention in `lastTree.Range.Items` with a per-range cache of `{keys, content-hashes, fingerprint}` (~24–40 B/item vs ~150–200 B/item today) so `{{range .Items}}` scales to large lists without controller, template, or wire changes.
- Whole-item `["u", key, item]` updates replace the partial-dynamics form; the bounded wire-cost regression on single-field-update workloads is owned in a "Why one path, not two" table rather than buried.
- Internal-only change. No new public type, no new tags, no new options. Existing examples (`DeleteRowController`, `ClickToLoadController`, `InfiniteScrollController`, `SortableController`) work unchanged.
- Frames Phoenix Streams from first principles: the wire ops `["a"]/["p"]/["i"]/["r"]/["u"]/["o"]` already exist; the missing piece is the server memory model.
- Includes a concrete Test Plan (unit, memory regression, hash collision soak, reconnect resync, het-range fallback, Clone independence, nil-Items wire format, extractRangeData regression gate, wire compat, browser E2E in lvt, benchmarks) and a Patterns Example commitment (10k-row LargeTableController exercising filter/sort/append/update/delete/reset).

## Test plan

- [ ] Read the proposal end-to-end; verify thesis, "Why one path, not two" table, and "Memory & Performance" estimates read honestly
- [ ] Spot-check that every cited symbol exists: `GenerateRangeDifferentialOperations`, `rangeContext`, `generateRemovalOps/Update/Insertion`, `extractRangeData`, `handleEmptyToItemsTransition`, `ClientNeedsStatics`, `mergeRangeItem` (cross-repo), `callMount`, `keys.GenerateItemHashFromSlice`
- [ ] Confirm the proposal honors the design decision that controllers, templates, and the wire format remain unchanged (no public-API surface anywhere in §5 / §6 / §10)
- [ ] Decide direction on the two remaining Open Questions (het-range long-term, volatile-field workloads); the other three (hash function placement, migration safety hatch, StreamState lifecycle) are resolved during review


adnaan:
Round 8 review feedback — five real findings addressed in 57ede0dd. One claim declined:

**Declined:** the symbol-verification table notes `isComplexInsertionPatternCtx` "actual location is `range_ops.go`" but the function is defined at `internal/diff/helpers_range.go:203`. `range_ops.go:91` only has the call site (`if len(ctx.addedKeys) > 0 && isComplexInsertionPatternCtx(ctx) {`). The §10 attribution to `helpers_range.go` is correct, so I'm leaving it. Per `livetemplate/CLAUDE.md` AI-review-loop rule "Its verifications are not infallible (it can hallucinate line numbers and occasionally miss a call site)" — flagging this one as a misattribution.

```
$ grep -n 'func isComplexInsertionPatternCtx' internal/diff/*.go
internal/diff/helpers_range.go:203:func isComplexInsertionPatternCtx(ctx *rangeContext) bool {
```

adnaan:
Round 13 review (claude-sonnet-4-6) processed. **Stopping the prcommentsfix loop here** — declaring convergence per `livetemplate/CLAUDE.md`'s AI-review-loop rule (no new functional issue surfaced this round; symbol verification all passes ✓; remaining items are cosmetic refinements, test-plan additions arguably belonging to the implementation PR review, or design reversals).

### Accepted but not pushed (cosmetic / implementation-PR-scope)

- **Auto-key stability test** (§11 gap noted in Round 13): valuable; the implementer should add a "render N → render N+1 with new auto-keys → assert no key collisions with `StreamState.Keys`" case to `internal/keys/loader_test.go`. Recording here so it isn't lost.
- **Adjacent-key collision soak case**: the implementer should ensure the §11 item 3 soak fixture exercises adjacent-key collisions specifically (where stale-content + actionable-payload mismatch is most consequential), not just random.
- **§5a empty-range precision** (`len(newItems) == 0` vs `newItems == nil`): the implementation should use `len(newItems) == 0` to handle both empty-initialized and nil slices uniformly. The proposal's intent is the former; a one-word tightening in the implementation PR is sufficient.
- **§5c nil-vs-`""` phantom-update mention**: already documented in the §10 `internal/keys/hash.go` row; not duplicating in §5c.
- **Op ordering verification**: covered by the §11 mixed-op golden in item 1; the implementer verifies against current `GenerateRangeDifferentialOperations` when writing it.
- **§6 PR-prerequisite framing**: §6 already says "this is a PR prerequisite, not a soft expectation"; the implementation PR description will link the `mergeRangeItem` grep output as already required.

### Declined with reason

**Round 13 §2 — env var escape hatch (`LVT_RANGE_STREAM_MODE=0`):** This reverses the §8 hard-cutover commitment that was explicitly resolved in earlier rounds and moved to "Resolved during review." Per `livetemplate/CLAUDE.md` "Project guidance trumps bot suggestions when they conflict" and the AI-review-loop rule "two consecutive rounds … is usually enough to stop": the env var is a public-surface design change — even if undocumented, env vars become de facto API contracts users find via grep, set in their deployments, and depend on. The §8 reasoning ("no user-facing API to be cautious about") still holds; recovery-by-revert is the correct policy for a docs-only design proposal whose first-implementation defect would be caught by the §11 acceptance bar before release. Re-litigating this would re-open a closed decision without new evidence.

### Round-by-round summary

13 rounds of bot review. Substantive findings addressed: function-name accuracy, statics-stripping reality (Items is dynamics-only), wire-protocol payload contract change, removal encoding (`""` for absent positions), `RangeStreamState` placement, hash-function placement (`keys.ItemHashUint64`), three-state lifecycle (with empty-range edge case), `extractRangeData` silent regression + new top-level entry point, `isComplexInsertionPatternCtx` + `areAllItemsAtStartCtx` + `areAllItemsAtEndCtx` co-adaptation (with `oldByKey` populated as `map[string]struct{}` presence-only — Round 12 caught a real bug in Round 11's fix here), `LoadExistingKeyMappings` for stream mode (with non-numeric and explicit-data-key handling), fuzz verifier `buildIDKeyMap`, dual-path `MarshalJSON` fix, `Clone()` deep-copy for `StreamState` slices, mixed-op + scattered-insertion + ≥4-tail-append + first-render goldens, fixed-PRNG-seed soak test, hash-collision consequence framing for actionable content, grep anchors per CLAUDE.md.

Resolved during review: hash placement, migration safety hatch, `StreamState` nil-state ambiguity. Open Questions remaining: het-range long-term; volatile-field workloads (answered by `LargeTableController.UpdateRandomRow` benchmarks).

Awaiting human review.

adnaan:
Round 4 review processed in `aca2f48f`. Five real findings addressed (nested-range clarification, breaking-change framing, test-files Phase 0 sweep, IssuedAutoKeys simplified away, hash-collision visibility). One declined:

**Declined: `grep "func.*TreeNode.*Clone"` "returns no hits"**

The regex matches the function declaration because `TreeNode` appears in both the receiver and the return type:

\`\`\`
$ grep -E "func.*TreeNode.*Clone" internal/build/types.go
func (tn *TreeNode) Clone() *TreeNode {
\`\`\`

The anchor is correct as-written. Per `livetemplate/CLAUDE.md` AI-review-loop note "Its verifications are not infallible (it can hallucinate line numbers and occasionally miss a call site)" — flagging this as a false-positive on the bot's part.

**Round 3's `IssuedAutoKeys` field reverted in this round.** Round 4's memory-cost concern (~16-48 B/entry × 10k = up to ~480 KB) made me re-examine the Round 11 framing that introduced it. The simpler "walk `StreamState.Keys`, find max numeric, set counter to max+1" approach is correctness-preserving: it guarantees every new auto-key is strictly greater than any existing numeric key, regardless of whether existing keys were auto-issued or explicit `data-key` values. Auto-keys are framework-internal IDs (not user-facing); "non-colliding" is what matters, not "low-numbered." `RangeStreamState` is back to three fields with no per-item map overhead.

adnaan:
Round 6 review processed in `00c89669` — declaring **convergence**, stopping the prcommentsfix loop.

### Round 6 outcome

The bot's Round 6 review opened with corrections to its *own previous round's misclaims* (FNV-1a presence, `RangeStreamState` placement decision, hash-input specification, safety-hatch open question — all four were already correctly stated in the proposal). That's the convergence signal per `livetemplate/CLAUDE.md`'s AI-review-loop rule: "successive rounds aren't identifying any new functional issue."

### What landed in `00c89669`

Round 6's "new observations" were test-plan additions, all real but small:
- **§11 item 1** — tightened assertion language for the no-ops golden ("empty ops array, not just non-panicking") and the N≥4 tail-append golden ("`len(ops) == 1 && ops[0][0] == 'a' && len(ops[0][1]) == 4`")
- **§11 item 11** (new) — `nil` → `""` phantom-update golden, pins down the "phantom no-op-render but real wire emission" semantics so a future refactor can't silently flip the contract
- **§11 item 12** (new) — empty-range deferral golden, locks in §5a's deferral path which would otherwise be dead code from a coverage standpoint

### Six rounds since the phased-plan addition

Rounds 1–6 of this loop produced ~25 substantive findings against the proposal as it grew from 290 → 388 lines:
- Phased plan added (§10 implementation phases with hard audit checkpoints)
- 11 production-code call sites enumerated for stream-mode adaptation (`extractRangeData`, `LoadExistingKeyMappings`, `buildIDKeyMap`, both `MarshalJSON` paths, `PrepareTreeForClient`, `helpers_compare.go:rangeItemsEqual`/`treeNodesEqual`, `internal/fuzz/invariants/helpers.go:convertedItems`, `helpers_value.go:HasRangeItems`, `handleMatchedRanges` no-ops branch, `diff_bench_test.go` benchmark mutations, both dispatch sites in `tree_compare.go`)
- 13 test-plan items with concrete acceptance assertions
- Phase 0 audit gates that close on paper before any implementation code lands
- IssuedAutoKeys field added in Round 3, then simplified away in Round 4 when the memory-cost analysis revealed the simpler "max numeric + 1" was correctness-preserving (auto-keys are framework-internal IDs)
- Two declined-with-reason replies for bot misclaims (`isComplexInsertionPatternCtx` location, grep-anchor false-positive)

### State

PR is `MERGEABLE` / `CLEAN`. CI green across all rounds. Awaiting human review.

PR #364, opened 2026-05-01
feat: streaming range Phase 3 — caller integration (cutover)

adnaan:
## Summary

Phase 3 of the streaming range refactor — the irreversible cutover. Wires `GenerateRangeStreamOperations` (built in Phase 2 #363) into production. Stream mode is now the default for homogeneous ranges; the legacy `GenerateRangeDifferentialOperations` path is reached only via the het-range fallback per proposal §5d.

- New `internal/diff/transition.go::TransitionToStreamMode` — top-level walker that builds `RangeStreamState` and nils `Range.Items` for homogeneous ranges
- Dispatch in `internal/diff/tree_compare.go` at `handleMatchedRanges` + `handleRangeMatch` — runs BEFORE any read of `Range.Items` (the regression gate from test plan item 8)
- Adapter helpers in `internal/diff/range_ops.go` (`handleStreamModeRange`, `handleStreamModeRangeMatch`, `dispatchStreamMode`, `extractStreamModeNewSide`)
- Single transition site at top of `template.go::buildTree` (before `LoadExistingKeyMappings`) — runs on the previous render's `lastTree`, avoiding the wire-aliasing pitfall at the `lastTree =` assignment slots
- Read-site adaptations: `helpers_value.go::HasRangeItems`, `keys/loader.go` (+ new `LoadExistingKeysFromSlice`), `fuzz/invariants/{verifier,helpers}.go`

## Test plan

- [x] `GOWORK=off go test ./...` — green
- [x] `GOWORK=off go test -race -count=1 ./` — race-clean (root pkg, 457s; covers the `t.mu` chain through transition + diff + wire serialization)
- [x] Memory regression at N ∈ {10, 100, 1k, 10k}: ~273 bytes/item delta at N=10k, ~9.5x reduction vs legacy (well above proposal's >=4x target)
- [x] Sibling-repo verification: `examples/` full chromedp suite green; `lvt/e2e/` browser-tagged tests including `livetemplate_core_test.go` (TestTemplate_E2E_CompleteRenderingSequence covers add, remove, complete, sort, insert at start/middle, multi-op, no-change, performance)
- [x] Manual UX testing on iPhone over Tailscale (patterns server: lists, forms, navigation patterns)
- [x] Audit checkpoint appended to `docs/proposals/streaming-range-impl-audit.md`

## User-visible behavior change

Templates with `{{if}}` inside `{{range}}` (a common pattern) now produce **full-tree replacements** instead of granular ops once item structures diverge. This is proposal §5d "het-range fallback" — accepted by design. Functionally identical to the user; payloads are larger only on those renders.

## Phase 3 follow-ups (deferred)

- `prepare.go` StreamState propagation + `helpers_compare.go::rangeItemsEqual` extension — both planned per §10 but determined to be defensive code (call paths today never see populated StreamState through these functions). Reinstate if a real call site materialises (e.g., a broadcast path).
- Test plan item 10 (browser E2E in `lvt` repo) — out of scope; the Go-side wire-format invariants are tested here, and the client's existing `mergeRangeItem` already accepts whole-item maps so no client change required.
- Phase 4 cleanup: delete `CompareRangeItemsForChanges` and `compareRangeItemsWithKeyPos`; update `docs/specifications/tree-update-specification.md` per §5c.

### DIFF
--- a/docs/proposals/streaming-range-impl-audit.md
+++ b/docs/proposals/streaming-range-impl-audit.md
@@ -331,6 +331,101 @@ A future Phase 3+ helper added here MUST audit which `ctx` fields it reads — `
 
 ---
 
+## Phase 3 audit checkpoint
+
+**Date**: 2026-05-01
+**Branch**: `phase3/streaming-range-cutover` → PR pending
+**Base commit**: `c07977d9` (post-Phase-2 merge on `main`)
+
+### Files touched
+
+| File | Change |
+|---|---|
+| `internal/diff/tree_compare.go` | `handleMatchedRanges` + `handleRangeMatch` stream-mode dispatch at top of function (BEFORE any read of `oldTree.Range.Items`); no-op fallback predicate at line 125-126 extended to recognise stream-mode-empty (`Items==nil` with non-nil StreamState) so a stream-mode no-op render doesn't fall into `*changes = *newTree` |
+| `internal/diff/range_ops.go` | Two adapter helpers: `handleStreamModeRange` (top-level) + `handleStreamModeRangeMatch` (nested) — extract new-side via new helper `extractStreamModeNewSide`, call `GenerateRangeStreamOperations`, write ops into changes (or full-tree replacement on het-fingerprint nil return) |
+| `internal/diff/transition.go` (NEW) | `TransitionToStreamMode(*build.TreeNode)` — top-level walker that runs the homogeneity check via `CalculateStaticsFingerprint` (Phase 2-driven), populates `RangeStreamState`, nils `Items`. Idempotent, nil-safe, top-level only per §5a |
+| `internal/diff/transition_test.go` (NEW) | 7 unit tests: homogeneous fires, heterogeneous defers, empty defers, single-item fires, idempotent re-entry, nested ranges NOT transitioned, nil tree safe |
+| `internal/diff/helpers_value.go` | `HasRangeItems` extended via existing `IsStreamMode()` predicate — answers the *logical* "does the range render anything" question for stream-mode trees |
+| `internal/keys/generator.go` | New `LoadExistingKeysFromSlice([]string)` — same numeric-max-tracking as `LoadExistingKeys` but reads keys directly from the slice (used by stream-mode loader path) |
+| `internal/keys/loader.go` | `LoadExistingKeyMappings` branches on `Range.StreamState != nil` → `LoadExistingKeysFromSlice(StreamState.Keys)` else legacy `LoadExistingKeys(Items)` |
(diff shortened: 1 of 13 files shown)


adnaan:
Replies on the two declined items from the round 3 review:

**1. "Dead `!ok` branch in `TransitionToStreamMode` inner loop"**

Declining. The round 1 review explicitly requested this defensive form ("Prefer the checked form `itemNode, ok := item.(*build.TreeNode)` with a continue/panic to make the invariant explicit at the assertion site... A future change that restructures the homogeneity check could silently break this"). Removing it now would contradict round 1's guidance and reintroduce the future-refactor fragility round 1 flagged. The two readings — "dead given current homogeneity check" vs "defensive against future restructure" — are both defensible; this PR follows round 1's call.

**2. Reachability of new nil-return guard in `GenerateRangeDifferentialOperations`**

The map-items path IS reachable: `TestCompareTreesAndGetChangesWithPath_TopLevelRange` and `TestHandleMatchedRanges_WithOps` (in `internal/diff/tree_compare_test.go`) construct ranges with `map[string]interface{}` items as fixtures, and depend on the diff engine's fall-back behavior when those items can't yield extractable keys. The guard was added in round 2 specifically to preserve those tests' fall-back path after refining the nil-vs-empty distinction in `handleMatchedRanges`. Removing it would re-break both tests.

Production paths use `*build.TreeNode` items only, so in production the guard is conservative cover for the nil-vs-empty contract that the test fixtures exercise.

adnaan:
Round 4 reply — both items are bot misreads of the current diff, no code change needed:

**Audit doc (round 3 item 3) — fixed in 8b5cf9f5**

The doc rename from `extractStreamModeNewSide` → `extractNewSideRange` did land in round 3:

\`\`\`
$ grep -n "extractStreamModeNewSide\|extractNewSideRange" docs/proposals/streaming-range-impl-audit.md
345:... extract new-side via new helper \`extractNewSideRange\` ...
\`\`\`

The audit doc currently matches the code. The round 3 commit message explicitly listed this as one of the fixes.

**Empty-key risk in `TransitionToStreamMode` — not reachable**

`getItemKeyWithPos` (in `internal/diff/helpers_keys.go:55-75`) has a fallback to `GenerateItemHash` (FNV hash of `itemNode.Dynamics`) for the `keyPos < 0` case:

\`\`\`go
func getItemKeyWithPos(item interface{}, keyPos int) (string, bool) {
    itemNode, ok := item.(*TreeNode)
    if !ok { return "", false }
    if itemNode.AutoKey != "" { return itemNode.AutoKey, true }
    if keyPos >= 0 {
        if key, exists := itemNode.GetDynamic(keyPos); exists {
            if keyStr, ok := key.(string); ok { return keyStr, true }
        }
    }
    return GenerateItemHash(itemNode), true   // <-- fallback
}
\`\`\`

So for a homogeneous range with no `data-key` attribute (`keyPos == -1`), each item still yields a non-empty hash-based key, and `keysList` is populated with FNV hashes rather than empty strings. The transition produces a valid `StreamState` and the diff engine sees distinct keys per item.

The auto-generated-keys path is exercised by every Phase 2 unit test that doesn't supply explicit `data-key` attrs (transition tests construct items with the key column at position 0, but production paths exist that omit the attribute and lean on the fallback).

PR #365, opened 2026-05-01
feat: streaming range Phase 4 — cleanup + spec update

adnaan:
## Summary

- Delete the partial-delta `["u"]` producer per proposal §314: `CompareRangeItemsForChanges`, `compareRangeItemsWithKeyPos`, `generateUpdateOps`, plus the now-dead `handleNestedTreeNodeChange` and `isMeaningfulValue` helpers.
- Update `docs/specifications/tree-update-specification.md` per §5c (full-snapshot producer contract) + §232 (consumer must replace, not merge).
- Close a Phase 3 coverage gap: `TransitionToStreamMode` now transitions root-level ranges in addition to direct-child ranges. Wrapper-less templates (`{{range .Items}}<x/>{{end}}`) now stream-mode correctly.

## The deletion cascade

`compareRangeItemsWithKeyPos` is the sole producer of partial-delta `["u"]` payloads, called only by `generateUpdateOps`. After Phase 4's spec update, partial-delta `["u"]` is non-compliant on the wire — so `generateUpdateOps` must also go (advisor framing: "compelled, not chosen"). With `generateUpdateOps` gone, kept-item content changes would silently emit no ops; a hash-based content-change detector in `GenerateRangeDifferentialOperations` signals nil-return on any mismatch, which `handleMatchedRanges` already handles by falling through to `*changes = *newTree` (full-tree replacement per §5d).

**Approach A** chosen over Approach B (delete `GenerateRangeDifferentialOperations` entirely): keep `generateRemovalOps`, `generateInsertionOps`, etc. so het ranges with pure-structural changes (add/remove/reorder only) still emit compact ops. The proposal §5d wording compels deleting `["u"]`-emission, not all per-op emission.

## Phase 3 coverage gap surfaced + closed

`TransitionToStreamMode` walked `tree.Dynamics` looking for nested ranges but never checked the root tree itself. For templates whose root tree IS the range (no wrapper element), the transition silently missed it → legacy path → partial-delta `["u"]` ops were emitted. Phase 3 didn't catch this because the legacy path masked it; Phase 4 surfaces it as a panic in `TestAllOperationTypes/Update`. Refactored to extract `transitionRangeIfHomogeneous` and call it on both `tree.Range` and direct-child Dynamics ranges.

## Net change

`8 files changed, 311 insertions(+), 1022 deletions(-)` — net −713 lines.

- 16 tests deleted (asserted deleted partial-delta behavior)
- 3 tests migrated to assert nil-return (`_Mixed`, `_UpdateAndReorder`, `_MultipleUpdatesAndReorder`)
- 4 new tests: `_KeptItemContentChange_FallsBack`, `_StructuralAndContent_FallsBack`, `_PureStructural_StillEmitsOps`, `TestTransitionToStreamMode_RootRangeFires`
- 2 e2e subtests removed (covered by the new internal/diff tests)

## Wire regression accepted

Het ranges with kept-item content changes now emit full-tree replacement instead of partial-delta `["u"]`. Pure-structural changes to het ranges still emit compact ops. Homogeneous ranges are unaffected — they transition to stream mode on render 2 and emit compact `["u"]` with full snapshots per §5c. Acceptable per §5d.

## Test plan

- [x] `GOWORK=off go test ./...` — green (root pkg ~108s, all internal pkgs <1s)
- [x] `GOWORK=off go test -race -count=1 ./` — race-clean (root pkg, 526s)
- [x] `grep -rn "CompareRangeItemsForChanges\|compareRangeItemsWithKeyPos\|handleNestedTreeNodeChange\|isMeaningfulValue\|generateUpdateOps" internal/ --include="*.go"` — zero hits
- [x] `examples/test-all.sh` — all 12 packages green with `-race` (chromedp E2E via patterns at 230s, todos, chat, etc.)
- [x] `examples go test ./...` — all 13 packages green (separate run)

Sibling-repo failures unrelated to Phase 4: `lvt` (4 `sqlc not installed`), `tinkerdown` (2 missing local scripts), `livetemplate/e2e/docker` (Dockerfile pins `golang:1.24-alpine` but project requires `go 1.26.0`). All fail on `main` for the same reasons.

### DIFF
--- a/docs/proposals/streaming-range-impl-audit.md
+++ b/docs/proposals/streaming-range-impl-audit.md
@@ -470,12 +470,12 @@ Het ranges with kept-item content changes now emit full-tree replacement instead
 ### Test count delta
 
 - 1 new unit test in `internal/diff/transition_test.go` (`TestTransitionToStreamMode_RootRangeFires`)
-- 3 new unit tests in `internal/diff/range_ops_test.go` (`_KeptItemContentChange_FallsBack`, `_StructuralAndContent_FallsBack`, `_PureStructural_StillEmitsOps`)
+- 5 new unit tests in `internal/diff/range_ops_test.go` (`_KeptItemContentChange_FallsBack`, `_StructuralAndContent_FallsBack`, `_PureStructural_StillEmitsOps`, `_StaticsOnlyChange_FallsBack`, `TestHasKeptItemChanged_NonTreeNodeReturnsTrue`)
 - 3 migrated tests in `internal/diff/range_ops_test.go` (`_Mixed`, `_UpdateAndReorder`, `_MultipleUpdatesAndReorder` — assert nil-return)
 - 16 deleted tests in `internal/diff/range_ops_test.go` (13 `TestCompareRangeItemsForChanges_*`, `TestGenerateUpdateOperations`, `TestGenerateRangeDifferentialOperations_Update`, `_UpdateNoReorder`)
 - 2 deleted subtests in `e2e_update_spec_test.go` (`update_single`, `mixed_operations`)
 
-**Net: 4 new + 3 migrated − 18 deleted = −11 tests.**
+**Net: 6 new + 3 migrated − 18 deleted = −9 tests.**
 
 The deletion delta is intentional: per-item ["u"] ops no longer exist as a wire shape, so tests that asserted them have no behavior left to verify. New tests cover the surviving fallback contract.
 
--- a/docs/proposals/streaming-range-impl-audit.md
+++ b/docs/proposals/streaming-range-impl-audit.md
@@ -426,6 +426,90 @@ Phase 4's cleanup audit can verify "no production caller of `compareRangeItemsWi
 
 ---
 
+## Phase 4 audit checkpoint
+
+**Date**: 2026-05-01
+**Branch**: `phase4/streaming-range-cleanup` → PR pending
+**Base commit**: `6d46c35e` (Phase 3 merge on `main`)
+
+### Files touched
+
+| File | Change |
+|---|---|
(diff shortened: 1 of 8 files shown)


adnaan:
**Round 1 fixes pushed in 2f5a58cd:**

1. **`hasKeptItemContentChanged` extended to compare statics fingerprint** (Copilot + Claude both flagged this). The hash check alone missed the conditional-branch flip pattern (`{{if .Done}}<s>{{end}}` toggling) where Dynamics stay identical but item structure changes. Added `TestGenerateRangeDifferentialOperations_StaticsOnlyChange_FallsBack` as the regression gate for the deleted `_StaticsDiffer`/`_ConditionalBranchChange` tests' coverage.

2. **Spec clarified to exclude key position from snapshot requirement** (Copilot). The implementation's `dynamicsToUpdatePayload` skips the key position because `itemId` already encodes it; the spec text and examples now match.

3. **Comment hygiene** (Claude). Stripped "Phase 3"/"Phase 4"/"legacy" references that rot. Kept the invariants.

4. **Tightened `_PureStructural_StillEmitsOps`** to assert the new key (id2) appears in the emitted op rather than just any insertion-class op (Claude minor).

**CI status — coordinated lvt PR opened:**

The "Test LVT against Core Changes" failure is the documented Phase 4 wire-shape change surfacing in lvt's golden files. The lvt todos template uses `{{if .Completed}}` and `{{if .Priority}}` inside its range, making items het → all 6 update tests in `TestTemplate_E2E_CompleteRenderingSequence` now take the full-tree fallback per spec §5d.

Coordinated golden refresh: https://github.com/livetemplate/lvt/pull/316. Merge order is lvt#316 first, then this PR's `Test LVT against Core Changes` re-runs green.

The "Test Examples against Core Changes" failure is `TestHighlightOnChange/Increment_Flashes_Both_Highlight_Targets` — a pre-existing flake on main (also failed on `main` run 25201322732 yesterday, 2026-05-01T03:49:13Z, before this PR existed). Not a Phase 4 regression.

The Claude review's "wrapper-less template integration test" suggestion is good but the existing `TestTransitionToStreamMode_RootRangeFires` covers the mechanism; the `TestAllOperationTypes/Update` test (root pkg) covers the full render→update path for a wrapper-less template (`{{range .Items}}<span>...{{end}}`) and went from panic on Phase 3 to PASS on Phase 4 via the root-range fix — that's the integration test the suggestion asks for, just under a different name.

The dead `if !ok { continue }` in `transitionRangeIfHomogeneous` is intentional defensive guarding (per Phase 3 round 1 reviewer's prior feedback to keep the safety bias). Declining to remove it for the same reason.

PR #366, opened 2026-05-01
feat: streaming range Phase 5 — benchmark gate + measured §7 numbers

adnaan:
## Summary

Phase 5 of the [streaming range rendering refactor](docs/proposals/streaming-range-rendering-proposal.md) — the **benchmark gate** that replaces §7's order-of-magnitude estimates with measured numbers and ships a CI regression test so the per-connection memory win can't silently regress.

Zero production code touched. New test files + doc updates only.

## Changes

**New files:**
- `internal/diff/range_ops_bench_test.go` — 9 stream-mode CPU benchmarks (`BenchmarkRangeDiff_Stream_{Append,Update,Reorder}_{Small,Medium,Large}` at N ∈ {10, 100, 10k}) + 7 wire-size benchmarks publishing per-op JSON byte counts via `b.ReportMetric` + a synthetic `LegacyPartialDelta_Update_WireSize` baseline (the legacy producer was deleted in Phase 4).
- `internal/diff/range_memory_test.go` — `TestRangeRetainedMemory_LegacyVsStream` regression gate. `runtime.ReadMemStats` HeapAlloc delta over 8 retained trees, double-GC stabilisation, **cold-start warmup discard** so N=10 is stable across runs. Per-N thresholds (1.8× / 4× / 5× / 5×) reflect the asymptote — fixed `RangeData/StreamState` overhead dilutes the per-item win at small N. Skipped under `-short`.

**Doc updates:**
- `docs/proposals/streaming-range-rendering-proposal.md` — status `Proposed` → `Implemented`; §7 table replaced with measured values; "Estimates, to be replaced…" lead deleted (audit invariant per §333).
- `docs/proposals/streaming-range-impl-audit.md` — Phase 5 audit checkpoint appended (files-touched table, measured numbers, methodology deviation note, partial OQ2 answer).
- `docs/performance/known-bottlenecks.md` — "Range Diff Operations" subsection refreshed with current measurements across legacy-map / legacy-TreeNode / stream-mode tiers; new "Streaming Range Retention" section captures the strategic win (10k-row × 100 connections: ~250 MB → ~40 MB).

## Measured §7 numbers

**Per-item retained heap bytes:**

| N | Legacy | Stream | Drop |
|---|---|---|---|
| 10 | ~270 B | ~80 B | ~3.4× |
| 100 | ~245 B | ~47 B | ~5.2× |
| 1000 | ~256 B | ~41 B | ~6.2× |
| 10000 | ~256 B | ~41 B | ~6.3× |

**Wire bytes per op (json.Marshal of stream-mode op output):**

| N | Append-1 | Update-1-field | Legacy synthetic |
|---|---|---|---|
| 10 | 52 B | 45 B | 33 B |
| 100 | 54 B | 47 B | 33 B |
| 10000 | 58 B | 51 B | 33 B |

Stream-mode update grows ~1.4× vs synthetic legacy partial-delta — well below the proposal's projected 3× ceiling. Inserts/deletes/reorders unchanged.

## Test plan

- [x] `GOWORK=off go test -count=1 ./...` — full suite green
- [x] `GOWORK=off go test -v -run TestRangeRetainedMemory_LegacyVsStream -count=10 ./internal/diff/` — stable across 10 consecutive runs (warmup fix eliminated cold-arena outlier)
- [x] `GOWORK=off go test -bench=. -benchmem -benchtime=1s ./internal/diff/` — all 39 benchmarks ran cleanly, 61s total
- [x] `awk '/^## Memory & Performance/,/^## Reconnect/' docs/proposals/streaming-range-rendering-proposal.md | grep -ni estimate` — zero hits (audit invariant per §333)
- [x] `gofmt -l ./internal/diff/range_*_test.go` — clean
- [x] Examples sibling repo: pre-existing `TestHighlightOnChange` flake reproduced and confirmed unrelated (passes on retry, same as Phase 4)

## What this PR does NOT do

- **`LargeTableController` example + `UpdateRandomRow` benchmark**: Phase 6 deliverable. Closes Open Question 2 once the multi-field volatile-field workload has measured wire cost.
- **Production code changes**: Phase 5 is measurement infrastructure + doc updates only. The ratios it asserts were already shipped end-to-end by Phases 1-4 (#362-#365).

### DIFF
--- a/docs/performance/known-bottlenecks.md
+++ b/docs/performance/known-bottlenecks.md
@@ -2,9 +2,10 @@
 
 **Last CPU Profiled:** 2026-03-19
 **Last Memory Profiled:** 2026-03-23
+**Last Range-Diff Benchmarked:** 2026-05-01 (Phase 5 streaming-range)
 **Go Version (profiles):** 1.26.0
 **Go Version (benchmarks):** 1.26.0
-**Architecture:** arm64 (Apple M2)
+**Architecture:** arm64 (Apple M2 for profiles; linux/arm64 for Phase 5 benchmarks)
 
 ## Profiling Methodology
 
@@ -147,16 +148,43 @@ Showing top allocators (44.6 GB total across all benchmark iterations)
 - Bytes allocated: ~5.2 KB/op
 - Example: BenchmarkTemplateExecuteUpdates/large-update-8 (6789 ns/op, 5187 B/op, 123 allocs/op)
 
-**Range Operations:**
-- Add items: 222 allocs/op, 9.4 KB/op
-- Remove items: 124 allocs/op, 5.5 KB/op
-- Reorder items: 157 allocs/op, 6.8 KB/op
-- Update items: 157 allocs/op, 6.8 KB/op
+**Range Diff Operations (per-call, N=100 items, refreshed 2026-05-01):**
+
+Legacy map-based (`BenchmarkRangeDiffUpdate/Insert/Remove`, items as `map[string]interface{}`):
+- Update / Insert / Remove: ~19 allocs/op, ~17.6 KB/op, ~11 µs/op
+
+Legacy TreeNode-based (`BenchmarkRangeDiff_TreeNode_*`, items as `*TreeNode`, het-range fallback path):
+- Update: 2056 allocs/op, 57.9 KB/op, 269 µs/op
+- Reorder: 4023 allocs/op, 103 KB/op, 514 µs/op
+- LargeList (N=1000) Update: 20065 allocs/op, 660 KB/op, 2.9 ms/op
+
+Stream-mode (`BenchmarkRangeDiff_Stream_*`, post-Phase-3 default for homogeneous ranges):
+- Append-1 / Update / Reorder at N=100: ~2030 allocs/op, ~60 KB/op, ~280 µs/op
+- Append-1 / Update / Reorder at N=10000: ~200k allocs/op, ~6.3 MB/op, ~28 ms/op
+
+Per-call allocation profiles for stream-mode and legacy-TreeNode are similar (both walk N items to build the new-side `rangeContext`); the **Phase 5 win is in retained memory between renders, not per-call diff cost** — see "Streaming Range Retention" below.
 
 **Complex User Journey (E2E):**
 - Total allocations: ~5,083 allocs/op
 - Bytes allocated: ~252 KB/op
 - Example: BenchmarkE2EUserJourney-8 (256978 ns/op, 251941 B/op, 5083 allocs/op)
+
+### Streaming Range Retention (Phase 5, 2026-05-01)
+
+Per the [streaming range rendering proposal](../proposals/streaming-range-rendering-proposal.md), homogeneous ranges now retain a per-range snapshot (`RangeStreamState{Keys, Hashes, Fingerprint}`) instead of the full per-item `*TreeNode` slice. Measured by `internal/diff/range_memory_test.go` via `runtime.ReadMemStats` HeapAlloc delta over 8 retained trees with double-GC + warm-up:
+
+| N | Legacy retained (B/item) | Stream retained (B/item) | Drop |
+|---|---|---|---|
+| 10 | ~270 | ~80 | ~3.4× |
+| 100 | ~245 | ~47 | ~5.2× |
+| 1000 | ~256 | ~41 | ~6.2× |
+| 10000 | ~256 | ~41 | ~6.3× |
+
+**Strategic impact:** for a 10k-row table held by 100 concurrent connections, retained memory drops from ~250 MB to ~40 MB. This is the single largest memory win shipped since the 2026-03 PR #219/220/224 cluster, and it does not show up in the benchmark allocation tables above (which measure per-call diff cost, not retained `lastTree` cost).
+
+**CI gate:** `TestRangeRetainedMemory_LegacyVsStream` enforces per-N ratio floors (1.8× at N=10, 4× at N=100, 5× at N≥1k). Skipped under `-short`. Catches regressions in the retention path before they merge.
+
+**Wire cost trade-off:** worst-case single-field update on a 3-field item grows from synthetic legacy ~33 B to ~45-51 B on the wire (~1.4×, well below the projected 3× ceiling). Inserts/deletes/reorders are unchanged.
 
 ### Cache Memory Usage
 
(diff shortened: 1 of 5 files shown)


adnaan:
## Round 1 bot feedback addressed (commit eee4b4b9)

**Substantive fixes:**
- ✅ `t.Fatalf` on HeapAlloc-no-growth softened to `t.Skipf` (Claude Medium) — concurrent GC noise won't fail CI; ratio assertion is the meaningful gate
- ✅ `buildStreamModeOldTree` now `b.Fatalf`s if `TransitionToStreamMode` silently leaves `StreamState` nil (Copilot L21) — without this, benchmarks would measure the het-range fallback path
- ✅ Wire-size benchmarks `b.Fatalf` on `json.Marshal` error AND on `ops==nil` (Copilot L133/L149/L169) — no more silent bogus bytes/op

**Cosmetic polish:**
- ✅ Dropped always-nil metadata third return (Claude Low)
- ✅ `Dynamics[2]` → `Dynamics[fixtureStatusPos]` named constant (Claude Low)
- ✅ Documented `"item-50"` legacy-baseline key as deliberately matching mid-list updates (Claude Low)

**Declined:**
- Copilot L29 (race-detector affects ReadMemStats): the memory test runs in the standard suite, not the race suite. Adding `-race` skip would mask future signal if someone wires the test into the race suite. Net negative.

**CI failure note:** `Test Examples / TestAsyncOperations/Concurrent_Fetch_Reaches_Single_Result` is **pre-existing on examples main** — reproduces 3/3 locally on examples@`41aa783` (sortable-list PR #84) with no Phase 5 changes involved. Phase 5 ships zero production code; the failure is in the examples-repo chromedp E2E suite. Safe to re-run; if it persists, it's an issue to file against the examples repo separately.

adnaan:
## Round 2 bot feedback addressed (commit b5e3ae77)

**Substantive (Medium):**
- ✅ Memory test: replaced single-shot `t.Skipf` with 3-attempt retry loop. Persistent zero-growth across all attempts now indicates the helper is broken (not transient noise) — silent skips can no longer defeat the CI gate. Extracted `measureRetainedBytesOnce` as the primitive.
- ✅ Memory test: N=10 floor 1.8x → 1.5x with comment. Run-1 of `-count=10` consistently hit ~1.81x (cold-arena cost survives even warm-up); the meaningful regression signal is "win exists at all," not "exactly 1.8x."
- ✅ Added `BenchmarkRangeDiff_Stream_Reorder_WireSize_{Small,Medium,Large}` validating the §7 "reorder identical to pre-implementation" claim. Measured: N=10 → 99 B, N=100 → 999 B, N=10k → ~119 KB (linear, single `["o", keys]` op).

**Polish (Low):**
- ✅ Commented `stripStatics=true` vs `false` asymmetry between wire-size and CPU benchmarks (a future "fix" to match would silently inflate §7 by ~10×).
- ✅ §7 grew a "Wire: reorder-pair" column with the new measurements; 1k row's `—` cells now annotated `— (not measured)`.

**Note on Claude's review correcting itself:** Round 2's review explicitly flagged 3 inaccuracies in Round 1's review (`t.Fatalf` vs `t.Skipf`, fake nil-metadata, fake magic number). Useful self-correction; addressed substance only.

adnaan:
## Round 3 bot feedback addressed (commit fd248d89)

All cosmetic — no functional changes.

**Polish:**
- ✅ N=10 floor 1.5x → 1.4x (~26% headroom vs prior 17%, against CI runner variance)
- ✅ Documented the read-only invariant on `streamState` in `buildStreamModeOldTree` godoc — verified by grep, no writes or method calls on `streamState` anywhere in `range_ops.go`. Makes the "reuse same streamState across all b.N iterations" pattern explicitly justified.
- ✅ Added `b.ReportAllocs()` to all four wire-size benchmarks (Append/Update/Reorder + LegacyPartialDelta) to match the CPU benchmarks
- ✅ Refined "item-50" baseline comment: the key is fixed across N (matches N=100's mid byte-for-byte; would be "item-5" at N=10, "item-5000" at N=10k). The §7 column reads constant 33 B because the *synthetic* key is constant length.

**Verified-not-changed:**
- `GenerateRangeStreamOperations` is read-only on `streamState` (grep `streamState\.[A-Z][A-Za-z]* *=|streamState\.[A-Za-z]+\(` returns zero hits in `range_ops.go`). Benchmark reuse pattern is correct.

**CI:** `Test Examples` passed on this run (no `TestAsyncOperations` flake recurrence).

Round 1 had 5 substantive fixes, Round 2 had 3 substantive fixes, Round 3 has 0 substantive + 4 polish — convergence pattern reached. Watching for one more clean round before signaling merge-ready.

adnaan:
## Round 4 review — convergence reached

Round 4 contains no substantive issues — only two minor doc-style suggestions and a few style judgment calls. Per CLAUDE.md AI Code Review Workflow ("Recognize the pattern and stop pushing; address remaining cosmetic nits by reply rather than by push"), addressing inline rather than via another commit cycle.

**Round-by-round substantive count:**
- Round 1 (b141ab05 → eee4b4b9): **5 substantive** (t.Fatalf softening, defensive checks, Marshal error handling, naming polish)
- Round 2 (eee4b4b9 → b5e3ae77): **3 substantive** (retry loop, reorder wire-size benchmark, stripStatics asymmetry comment)
- Round 3 (b5e3ae77 → fd248d89): **0 substantive + 4 polish** (floor adjustment, ReportAllocs, comment refinements, read-only invariant doc)
- Round 4 (fd248d89): **0 substantive + 2 minor doc observations**

Two consecutive non-substantive rounds = convergence per the project guideline.

**Round 4 inline replies:**

1. **§7 legacy column qualifier (`33 B (fixed 7-char key)`)** — Acknowledged. The §7 prose already explains "The 'legacy' wire-size column is a synthetic baseline (`['u', 'item-50', {'2': 'updated'}]`)" so a careful reader can deduce the fixed-key behaviour. The Phase 5 audit doc also flags this as a methodology shortcut. A column-header annotation would be a clearer signal to a casual scanner — agree it's a real readability improvement, but folding into a Phase 6 doc pass keeps Round 4 from spawning another bot review cycle. Filing as a Phase 6 follow-up.

2. **PR description "1.8× threshold" mismatch** — Confirmed: the original PR description quoted observed ratios (1.8× / 4× / 5× / 5×) as if they were the gate values. The actual code has 1.4× / 4× / 5× / 5× as gates (lowered in Round 2 + Round 3). The test-file comment is correct; the PR description is stale. Acknowledged here for the record so the audit doc reader can reconcile.

3. **Style judgment calls (multi-paragraph comments, runtime assertion vs. comment claim)** — Declined as ROI-negative. The `buildStreamModeOldTree` doc comment carries a load-bearing invariant (read-only on streamState). Replacing it with a runtime assertion (`re-read streamState.Keys, assert len unchanged`) would add complexity for marginal safety, since the invariant lives in `range_ops.go` source one grep away. The current explicit doc comment + grep-verifiable claim is the right balance.

**CI status:** All bot review workflows passing. Cross-Repository Tests still in progress on this commit; will await final green before requesting merge.

PR is functionally complete and ready for merge once Cross-Repo Tests finishes green.

PR #369, opened 2026-05-02
perf: streaming range Phase 7 — type-direct hash + parallel build

adnaan:
## Summary

Two independent optimizations cutting server-side \`ExecuteUpdates\` latency at N=10,000 from **306 ms to 201 ms (−34%)**, with wire bytes unchanged at 127.8 B/op:

| | What | Files |
|---|---|---|
| **(A)** | **Type-direct hash** — replace reflective \`json.Marshal\` in \`formatHashPart\` with a type switch (string/int/bool fast paths, JSON fallback). Strings take an extra fast path when no chars need JSON-escaping. | \`internal/keys/hash.go\`, \`internal/keys/hash_test.go\` |
| **(B)** | **Parallel range build** — above N≥256, \`iterateSlice\` dispatches to a goroutine-pool variant across \`runtime.NumCPU()\` workers. \`reflect.Value.Index().Interface()\` pre-extracted on main goroutine. | \`internal/parse/range.go\` |

## Phase 8 attempted, reverted

This PR originally also included Phase 8 (skip the dead \`renderHTMLWithData\` call on the steady-state path), which would have cut another 110 ms at N=10k for a 75% total reduction. Code-review safety analysis held, but \`examples/todos\` \`TestTodosE2E/Modal_Positioning_And_Cancel\` reliably failed under the skip — the modal Cancel handler dispatched but the modal stayed in the DOM. Reverting only the skip restored the test.

There is a side effect of \`html/template.Execute\` that LVT's action-dispatch or state-update pipeline depends on, independent of the rendered HTML output. Not yet identified. \`docs/proposals/build-latency-optimization-proposal.md\` records the failure mode and three candidate hypotheses for the unidentified side effect; Phase 8 is parked for a separate PR with deeper investigation.

## Measured impact

\`BenchmarkLargeTable_UpdateRandomRow_WireBytes\` (linux/arm64, Go 1.26.1):

| N | Pre-Phase-7 | Post Phase 7 (A+B) | Change |
|---|---|---|---|
| 200 | 6.6 ms | 5.6 ms | −15% |
| 1,000 | 36 ms | 24.8 ms | −31% |
| 10,000 | **306 ms** | **201 ms** | **−34%** |

Wire bytes unchanged at 127.8 B/op (byte-stability invariant holds).

## Wire-stability invariant

\`RangeStreamState.Hashes\` are persisted across renders; stream-mode transition fingerprints depend on hash byte-stability across versions. Gated by \`TestFormatHashPart_ByteEquivalentToJSON\` in \`internal/keys/hash_test.go\` which asserts \`formatHashPart\` output matches \`json.Marshal\` byte-for-byte across all supported types (string / int variants / uint variants / bool / JSON-special chars / HTML-special chars / unicode / slice / map / float64).

## Regression gates

- \`TestFormatHashPart_ByteEquivalentToJSON\` (new) — locks hash byte-equivalence
- \`TestRangeBuildLatency_PostPhase7\` (new top-level test, skipped under \`-short\` and on CI) — local sanity check; CI relies on the bench in the examples repo

## Related PRs (queue order)

1. ✅ \`livetemplate#368\` — Phase 6 (recursive transition fix) — MERGED
2. **This PR** — Phase 7 (A+B per-render latency)
3. \`livetemplate#370\` (next) — Phase 8.5 keyGen plumbing cleanup; needs rebase after this lands

## Test plan

- [x] \`go test ./...\` green (all 16 packages)
- [x] \`go test -race -count=1 ./internal/{parse,keys,diff}/...\` green
- [x] All existing TestStreamMode_* tests pass
- [x] Hash byte-equivalence golden runs and passes (uint8/16/32/64 + int8/16/32/64 + uint coverage added in round 2 fixes)
- [x] \`examples/todos\` TestTodosE2E passes against this branch (verified locally after Phase 8 revert)
- [x] \`examples/patterns\` TestLargeTable passes against this branch


adnaan:
Thanks for the careful review. Push 54f9f73 addresses the two highest-leverage items; declining the rest with rationale below.

**Addressed in 54f9f73:**

- **TOCTOU on \`hasInitialTree\`**: now loaded once at the top of \`buildTree\` and reused. The relaxed-read invariant (monotonic latch, stale read costs ≤1 extra render) is preserved, but \`currentHTML\` and \`isFirstRender\` can no longer disagree.
- **\`keyGen\` shared across workers**: added the explicit invariant comment on \`iterateSliceParallel\` documenting that the safety holds today only because \`.Next()\` has zero call sites in \`internal/parse/\` — and pointing forward to PR #370 (Phase 8.5) which removes the plumbing entirely. A future contributor reaching for \`keyGen.Next()\` inside a range item now hits the invariant before introducing a silent race.

**Declining (with reasoning):**

- **Replace \`keyGen\` with \`noopKeyGen\` in this PR**: would be defensive engineering churn — Phase 8.5 (#370, already pushed) removes the parameter entirely. The merge order is #368 → #369 → #370; the window where the dead-but-shared \`keyGen\` exists is whatever time elapses between #369 and #370 merging. Adding a noopKeyGen and removing it one PR later is more code change than the actual fix in #370.
- **Map test case in \`TestFormatHashPart_ByteEquivalentToJSON\` is "fragile over Go version upgrades"**: Go's spec guarantees \`json.Marshal\` for \`map\` produces sorted keys (since Go 1.12). The same \`json.Marshal\` is called on both sides of the comparison; if Go ever changed behavior, both sides change identically and the test still passes. Not a real fragility.
- **\`float32\` not in the type switch**: it's a deliberate omission for the JSON-fallback path. Adding it as an explicit \`case float32:\` that calls \`json.Marshal\` would be an empty branch — same code as the default fallback. Not worth a comment explaining a non-omission.
- **\`itemCtx\` shared across goroutines, "verify with -race"**: \`go test -race -count=1 ./internal/parse/...\` is in CI and was run locally pre-push (passes). The \`Context\` struct's fields are populated at construction and never written after; per project convention "Context is read-only after construction" is a documented invariant — adding a comment that re-states it on \`contextWithStatics\` would be redundant.
- **Wall-clock latency ceiling test fragility**: the ceilings have ~2.5× headroom over observed medians (60ms median vs 150ms ceiling at N=10k). CI variance from GC pauses is real but typically stays under 2× on the GitHub Actions runners. Replacing with a benchmark + baseline-compare would be more robust but adds a different class of operational complexity (storing/comparing baselines across PRs). Will revisit if this test flakes in practice.
- **\`errors.Join\` for parallel worker errors**: the first error per worker-chunk is returned today. Multiple errors across chunks would be a sign of a deeper issue (template malformed, data corrupt) where a single error is sufficient diagnostic. Plain return keeps the code simpler.
- **Line numbers in proposal doc**: noted, will fix in a follow-up doc-only PR if the review surfaces other doc cleanups.
- **Style: \`// Each worker handles a contiguous chunk\` is a what-comment**: the threshold-rationale (`why N≥256`) and thread-safety audit (`why this is safe`) survived the comment cleanup; the chunk-iteration comment is right next to the inscrutable index arithmetic and stays for readability.

Holding for #368 to merge first per the queue order, then this rebases cleanly.

adnaan:
Round 2 fixes pushed in 5a9b6ea:

**Addressed:**
- ✅ \`itemCtx\` safety invariant — explicit read-only contract documented on \`contextWithStatics\` godoc, with the trigger condition (any future ctx-field write requires reverting iterateSliceParallel to per-worker copies). The walk-path grep confirms today's invariant: every ctx access is \`ctx.ShouldIncludeStatics()\`, and \`ClientStructures\` / \`CurrentPath\` are only read in \`build/types.go\` from outside the parse layer.
- ✅ Cross-PR reference removed from \`iterateSliceParallel\` comment per CLAUDE.md "don't reference current task or PR" — kept the substantive safety note (grep shows zero \`.Next()\` calls in \`internal/parse/\`).
- ✅ Latency ceilings widened (N=1000: 25→50ms, N=10k: 150→250ms) for shared-CI variance.
- ✅ Type-switch test coverage filled in: uint, uint8/16/32/64, int8, int16 added to \`TestFormatHashPart_ByteEquivalentToJSON\`.

**Declining (with reasoning):**
- **panicKeyGen wrapper for \`iterateSliceParallel\`**: still declining. Same as round 1 — Phase 8.5 (#370, already pushed and approved) drops the parameter entirely. Adding a wrapper just to remove it one PR later is churn. The grep-enforced invariant + the explicit safety comment cover the case; the structural fix is in #370.
- **Replace wall-clock test with benchmark + baseline-compare**: out of scope for this PR. The benchmark+baseline approach has its own complexity (storing baselines, comparing across PRs, choosing what counts as regression) and is a separate engineering decision. Wider ceilings absorb the immediate flakiness concern. If the test does flake in practice, that's the trigger for switching mechanisms.
- **Comment verbosity** on \`appendJSONValue\` etc.: kept the wire-stability invariant block (the why is non-obvious — bytestability persists across renders) and trimmed where redundant. The bot's specific examples (godoc walking through types, error-handling block in buildTree) — reviewed and trimmed where the code is already self-explanatory; left where the constraint is non-obvious.

Holding for #368 to land first per the queue order.

Issue #374, opened 2026-05-03
FOLLOW-UP: Tighten fingerprintWith benchmark helper allocations (from PR #372)


adnaan:
## Context

This follow-up was identified during the review of PR #372 (Phase 1 of #367 — `unparam` cleanup).

**Source PR:** https://github.com/livetemplate/livetemplate/pull/372
**Suggested by:** @claude[bot] (Round 2 review)

## Task Description

`fingerprintWith` in `internal/build/fingerprint_test.go` was converted from `string`-returning to void in PR #372 (since callers discarded the result), but the body still computes `_ = hex.EncodeToString(h.Sum(nil))` — allocating a string that's immediately discarded:

```go
func fingerprintWith(tree *TreeNode, h hash.Hash) {
    visitPath := make(map[*TreeNode]struct{})
    hashStructureWithCircularDetection(tree, h, visitPath)
    _ = hex.EncodeToString(h.Sum(nil))
}
```

The current form was kept intentionally to mirror the production fingerprinter's full work (including stringification) so the MD5-vs-FNV1a benchmark accurately compares end-to-end timings. But if benchmark allocation noise becomes an issue, the helper could be tightened to just `_ = h.Sum(nil)` (no encoding).

This is a low-priority perf-hygiene cleanup — the bot's note: "the marginal cost is low and the current form is harmless. Worth noting for a future cleanup pass."

## Decision Needed

- **Keep as-is:** if the benchmark's purpose is to mirror production end-to-end (encoding included), document that explicitly in the helper's comment so the next reader doesn't optimize it away by mistake.
- **Tighten:** if only the hash compute matters, drop the `hex.EncodeToString` step (saves one ~40-byte allocation per benchmark iteration).

## Original Comment

> **`fingerprintWith` still allocates a string it discards** […] The purpose of this benchmark helper is to time hash computation, not encoding. The hex step allocates a string that's immediately discarded. Could be tightened to just `h.Sum(nil)` (no allocation). That said, since `h.Sum(nil)` allocates its own `[]byte` anyway, the marginal cost is low and the current form is harmless. Worth noting for a future cleanup pass.

[View original comment](https://github.com/livetemplate/livetemplate/pull/372#issuecomment-4365656878)

---
*This issue was automatically created by prmonitor from PR review comments.*

Issue #373, opened 2026-05-03
FOLLOW-UP: Resolve generateTreeInternalWithErrors deprecation notice (from PR #372)


adnaan:
## Context

This follow-up was identified during the review of PR #372 (Phase 1 of #367 — `unparam` cleanup).

**Source PR:** https://github.com/livetemplate/livetemplate/pull/372
**Suggested by:** @claude[bot] (review comment, both review rounds)

## Task Description

The internal helper `(*Template).generateTreeInternalWithErrors` (in `template.go`) carries a comment marking it deprecated:

```go
// DEPRECATED: Tests should use buildTree() directly. This wrapper exists only for
// existing test code and will be removed in a future version.
func (t *Template) generateTreeInternalWithErrors(data interface{}) (*treeNode, error) {
    return t.buildTree(data, nil)
}
```

PR #372 simplified the wrapper's signature (dropped the always-nil `messages` param) but did not migrate the ~10 test callers across `template_test.go`, `e2e_update_spec_test.go`, etc. to use `buildTree()` directly. So the deprecation notice remains aspirational rather than a live migration plan.

Two reasonable resolutions:

1. **Honor the deprecation:** migrate all test callers from `generateTreeInternalWithErrors(data)` → `buildTree(data, nil)`, then delete the wrapper.
2. **Drop the deprecation notice:** if the wrapper is staying for the foreseeable future, the comment is misleading. Replace it with neutral documentation describing what the helper does and why it exists.

## Original Comment

> But the PR *also* updates all existing callers (in `template_test.go`, `e2e_update_spec_test.go`, etc.) to use the simplified `generateTreeInternalWithErrors` form — they're not migrated to `buildTree()`. If the intent is to keep the wrapper for now and delete it in Phase 3, the deprecation comment is fine as-is. If the intent is to actually remove it, the Phase 2/3 follow-ups should include the migration. Not a blocker — just flag whether the deprecation doc reflects a real plan.

[View original comment](https://github.com/livetemplate/livetemplate/pull/372#issuecomment-4365651095)

---
*This issue was automatically created by prmonitor from PR review comments.*

PR #375, opened 2026-05-03
chore: enable unparam in pre-commit gate (Phase 2+3 of #367)

adnaan:
## Summary

Phase 2 and Phase 3 of #367, building on the cleanup that landed in #372. Adds `unparam` to the pre-commit gate so the 32 findings cleaned up in #372 can't drift back in, and documents the linter's scope and limitations in `CLAUDE.md`.

### `scripts/pre-commit.sh`
- Extends `--enable-only` from `errcheck,govet,ineffassign,staticcheck,unused` → `errcheck,govet,ineffassign,staticcheck,unparam,unused` (line 35)
- Adds an `unparam` block to the LLM-assistant fix-guidance section, matching the existing pedagogical tone of the other linter blocks. Covers all three patterns unparam catches (unused params, always-constant args, dead returns), with concrete recipes for each, including the special-case advice for typed-callback closures (defensive nil-check rather than `_ = msg` suppression — the pattern used in #372 for `pubsub.MessageHandler`).

### `CLAUDE.md`
- Expands the "Pre-commit Hook" subsection with:
  - The full linter set in a table (one-line description per linter)
  - An "`unparam` Scope and Limitations" subsection documenting **both** limits:
    1. No transitive tracing (the keyGen-style drift caught manually in #370)
    2. golangci-lint v2's default unparam is **less strict** than standalone `unparam -tests ./...` — documented so readers don't over-trust the gate

## Verification

- [x] `golangci-lint run --enable-only=errcheck,govet,ineffassign,staticcheck,unparam,unused ./...` → 0 issues against current main (post-#372)
- [x] Lint phase ~2.0s on this machine; pre-commit total well within the 300s `go test` budget
- [x] **Negative gate test:** temporarily added an always-nil-error helper to `internal/util/strings.go`; confirmed `unparam: result 1 (error) is always nil` fires and blocks the commit; reverted before push
- [x] `bash scripts/pre-commit.sh` passes on this PR's own changes
- [ ] CI on this branch

## Notes for reviewers

The "less strict than standalone" caveat in `CLAUDE.md` came from running the negative gate test — golangci-lint v2 reliably catches always-nil returns and always-constant args (which were the bulk of #372's findings) but skips some pure unused-param cases on unexported functions that the standalone `unparam -tests ./...` flags. Documenting this proactively so future maintainers don't get burned thinking the gate is catching everything.

This PR closes the rollout for #367. Two follow-ups from #372 review (#373, #374) are tracked separately.


adnaan:
Thanks for the review — addressed all 3 nits in 6b3ffddf:

1. **Tighten case-(c) guidance** (Claude bot) — Reframed so benchmark-style work is the only legitimate exception, with an explicit warning against using `_ =` as a production-code escape hatch. The previous wording could have read as license to silence findings; this version makes the boundary explicit.

2. **"inline at" → "inline into"** (Copilot grammar nit) — Fixed.

3. **`govet` description in linter table** (Claude bot) — Replaced "inline hints" with a concrete enumeration: "printf formats, struct tags, copy locks, shadowed vars, plus the `inline` analyzer that flags constants like `reflect.Ptr` aliased to a canonical `reflect.Pointer`". Cross-references PR #372's actual experience with that check, so future readers can see what kind of finding to expect.

Also picked up the cross-section consistency note about the 30s manual-run vs 300s gate timeout — added a parenthetical to the Pre-commit Hook section explaining why they differ (cold-cache headroom).

Pre-commit hook still passes clean on the new commit. CI re-running.

adnaan:
Thanks — addressed in 49817d8a.

**On the attribution question:** I empirically verified before changing. With a fresh `reflect.Ptr` usage in `internal/util/`:

\`\`\`
$ golangci-lint run --enable-only=govet ./internal/util/...
internal/util/test_reflect_ptr.go:6:21: inline: Constant reflect.Ptr should be inlined (govet)
$ golangci-lint run --enable-only=staticcheck ./internal/util/...
0 issues.
\`\`\`

So the finding **is** under govet (golangci-lint v2.12.1 bundles the `inline` analyzer into its govet integration). But your broader clarity concern is valid: a contributor running `go vet` directly won't see that analyzer, since it's not in upstream Go's standard suite. So I took the secondary suggestion — drop the parenthetical entirely. Cleaner regardless of attribution.

The four core govet checks (printf, struct tags, copy locks, shadows) stand on their own without needing the example.

CI re-running.

adnaan:
Round 3 — addressed the two substantive nits in c520aab4, declining the two style-only ones with rationale.

**Addressed:**

1. **Defensive nil-check example tightened.** You're right, `if msg == nil { return err }` had `err` unbound — confusing example. Replaced with the real sentinel-error pattern that landed in PR #372's `pubsub/redis_test.go` (`var errNilMsg = errors.New(...)` then `if msg == nil { return errNilMsg }`). Now the example actually compiles and matches an existing precedent in the codebase.

2. **Parenthetical on the 300s timeout dropped.** Fair point — after the 30s→300s correction, the parenthetical explained a discrepancy that no longer exists in the doc. Pure noise; gone.

**Declined with reasoning:**

3. **Linter guidance block ordering** in `pre-commit.sh` — the existing block order (errcheck → unused → staticcheck → govet → ineffassign) was the pre-existing convention before this PR. I added unparam between `unused` and `staticcheck` to keep "code shape" linters grouped (unused/unparam) before "analysis" linters (staticcheck/govet). Reordering all five blocks to match `--enable-only` flag order would be a separate scope and a re-architecture of how the help text reads. Happy to file as a follow-up if you'd like the ordering rationalized, but not in this PR.

4. **PR number references in CLAUDE.md `unparam` section** — the CLAUDE.md philosophy you cite ("don't reference the current task, fix, or callers") is in the *code-comments* guidance section, where it argues against rotting `// added for X flow` annotations. CLAUDE.md itself is a reference doc with very different rotting characteristics: linking to PR #370 and PR #372 gives future maintainers concrete examples of what the limitations *look like in practice* (PR #370 is the canonical "transitive-tracing blind spot" instance; PR #372 is what the gate now catches). Without those anchors, the limitation paragraphs are abstract. Keeping the refs.

CI re-running.

adnaan:
Round 4 — thanks for the approve. Pushed one substantive fix in 72592657, declining the other two with rationale.

**Addressed:**

- **Linter list duplicated across 3 places** — added a one-line note above the `Linter Set Rationale` table marking it as descriptive only and pointing to the script's `--enable-only` as authoritative. Future linter changes need a single source-of-truth edit; the doc is now explicit about that.

**Declined (cosmetic):**

- **Block ordering vs flag order** — same nit as Round 3, same rationale: existing block order groups linters by intent (errcheck → unused → unparam → staticcheck → govet → ineffassign), not by flag-list order. Reordering all 5 would be a separate cleanup PR. Happy to file as a follow-up if it's worth doing on its own.

- **govet description differs slightly between CLAUDE.md and script** — both are intentionally partial summaries of govet's many checks; perfect alignment would require keeping two prose passages in sync forever, which is the maintenance burden we're trying to avoid (per the same anti-drift principle just addressed for the linter list). The two phrasings agree on the "this is suspicious-construct stuff" gist; that's enough.

**On the Unit Tests CI failure (Round 4 push):**

`TestPerConnectionState_ServerActionPersists` failed with `WebSocket read failed: i/o timeout` after 11.65s. PR2 has zero Go code changes (only `scripts/pre-commit.sh` and `CLAUDE.md`), and the failing test is a Redis-testcontainer + WebSocket integration test — classic CI-environment flakiness. Re-triggered the failed run; it's currently in progress.

Per CLAUDE.md convergence heuristic, this is the **second consecutive cosmetic-only round** (Rounds 1–2 were substantive: case-(c) framing, govet attribution; Rounds 3–4 are style nits). Stopping the push cycle here unless something functional surfaces.

adnaan:
Round 5 — fixed the govet shadow factual error in 998654d7, declining the other two (one of which is your own previous suggestion that contradicts itself).

**Addressed:**

- **govet "shadowed vars" inaccuracy** — verified empirically. A deliberate shadow case (\`x := 1; if … { x := 2; … }\`) does not fire \`golangci-lint run --enable-only=govet\` in our default-config gate. You're right that shadow is opt-in and requires \`shadow: true\` in config. Dropped "shadowed vars" from the description; the remaining three (printf formats, struct tags, copy locks) are all default-on and accurately describe what the gate catches.

**Declined (oscillation between rounds):**

- **30s vs 300s parenthetical** — Round 3 (your previous suggestion) explicitly told me to drop the parenthetical because "it explains a discrepancy that no longer exists once the 300s is the documented value." I dropped it in c520aab4. This round you're now suggesting to add it back. Per CLAUDE.md "review rounds can iterate indefinitely if each round only rewords previous suggestions" — this is an oscillation; sticking with the Round-3 decision.

- **Case-(c) framing** — Round 1 already addressed this with the explicit caveat "Don't use this exception to silence findings in production code — actually drop the dead return." Reordering "constraint before permission" is purely stylistic and doesn't change what the guidance says. Per CLAUDE.md "address remaining cosmetic nits by reply rather than by push."

**Convergence signal:**

This is the **third consecutive cosmetic-only round** (Rounds 3–5), with the bot now contradicting its own prior suggestion. Per CLAUDE.md "two consecutive rounds of cosmetic-only feedback is usually enough to stop." Stopping the push cycle here.

The 998654d7 commit is the final functional improvement on this PR — a real factual fix surfaced by your shadow observation. Thanks for that one.

adnaan:
Round 6 — both raised observations are oscillations from prior rounds. Per CLAUDE.md "AI Code Review Workflow" guidance, this is the convergence signal. Not pushing further.

**Block ordering:**
This is the third round (Rounds 4, 5, and now 6) raising guidance-block ordering. I declined in Round 3 with the rationale that the existing block order groups linters by intent (errcheck → unused → unparam → staticcheck → govet → ineffassign), and reordering would be a separate cleanup PR. That rationale hasn't changed. Same decline.

**Timeout 30s vs 300s parenthetical — explicit oscillation receipts:**

| Round | Bot said |
|---|---|
| Round 3 | "the parenthetical can be dropped; the number speaks for itself" |
| Round 5 | "a parenthetical … would make the intent explicit" |
| Round 6 | "Recommend adding a parenthetical or note to clarify these are different contexts" |

This is exactly the pattern CLAUDE.md describes: *"review rounds can iterate indefinitely if each round only rewords previous suggestions."* I'll stick with the Round-3 decision (dropped) and not flip-flop based on round-to-round wording changes.

**The 30s and 300s are intentionally different**: the manual-run example uses 30s as a quick check during iteration; the pre-commit gate uses 300s to absorb cold-cache full-suite runs without false-positive timeouts. A reader who infers "one of these is wrong" can resolve it in 30 seconds by running the script. No edit needed.

**Convergence summary:**

- Round 1: substantive (case-(c) framing, govet inline, grammar) → 3 fixes pushed
- Round 2: factual question (govet attribution) → empirical verification + cleaner fix
- Round 3: 4 cosmetic nits → 2 fixed, 2 declined
- Round 4: "Approve" + 3 cosmetic → 1 fixed (linter-list duplication note), 2 declined
- Round 5: 3 cosmetic, including a real factual error (govet shadow) → 1 fixed, 2 declined
- Round 6: 2 cosmetic, both oscillations → 0 fixed, both declined

Per CLAUDE.md heuristic, **two consecutive cosmetic-only rounds = convergence**. We're past three. Halting the push cycle.

CI on the latest commit (998654d7) is finishing the cross-repo tests; benchmark already posted "no significant regressions." The PR is functionally ready to merge.

Issue #327, opened 2026-04-06
FOLLOW-UP: Extract shared button-name detection helper (from PR #326)


adnaan:
## Context

This follow-up task was identified during the review of PR #326.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/326
**PR Title:** fix: parse individual form fields in multipart submissions
**Suggested by:** @claude[bot]

## Task Description

The button-name-as-action detection loop is duplicated in `parseMultipartForm` and `parseURLEncodedForm` (both in `internal/send/message.go`). Extract a shared private helper to eliminate the duplication:

```go
// detectButtonAction scans form fields for a single empty-value field,
// treating its key as the action name (standard HTML button-name pattern).
func detectButtonAction(fields map[string][]string, reserved map[string]bool) string
```

Similarly, the field-to-data-map conversion loop could be extracted:

```go
// fieldsToDataMap converts form fields to a data map, excluding reserved keys.
func fieldsToDataMap(fields map[string][]string, reserved map[string]bool) map[string]interface{}
```

## Original Comment

> The button-name detection + field-to-map loop is now duplicated in two places. Not a blocker for this PR (the fix is focused and correct), but extracting a `detectButtonAction` helper and a `fieldsToDataMap` would make future changes easier to keep in sync.

---
*This issue was automatically created by prmonitor from PR review comments.*

Issue #237, opened 2026-03-25
FOLLOW-UP: Replace empty-value button heuristic with explicit client-side action


adnaan:
## Context

This follow-up task was identified during the review of PR #233.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/233
**Suggested by:** @claude-bot, @copilot
**Priority:** High

## Task Description

The server-side button-name detection in `parseURLEncodedForm()` uses an empty-value heuristic: a form field with an empty string value is treated as a submit button whose name is the action. This has false positives:

- A user clearing a text input submits `field=` which looks identical to `button=`
- A select with an empty default option triggers the same pattern

**Fix:** Have the JS client extract the clicked button's name into the `action` field on interception (already done for the WebSocket path). For the no-JS HTTP POST path, consider requiring `name="action" value="X"` convention or adding a hidden `_lvt_action` field injected by the framework.

The ambiguous guard (skip if multiple empty-value fields) is a workaround, not a proper fix.

---
*This issue was automatically created from PR #233 review comments.*

adnaan:
**Phase 2 handoff note** — when the client-repo PR for Phase 2 (`livetemplate/client`) opens, it must include the e2e test referenced in Verification item 3 of [`docs/proposals/explicit-submitter.md`](https://github.com/livetemplate/livetemplate/blob/main/docs/proposals/explicit-submitter.md#verification):

> New e2e test (or extend an existing form submission e2e in the lvt repo at `e2e/livetemplate_core_test.go`) verifying the explicit submitter path round-trips through both the HTTP and WS paths.

This issue is the cross-repo tracking pin until that PR opens. Posting here (per the round-3 review of #401) so the requirement isn't dropped in the handoff between the two repos.

adnaan:
Closing — the full server↔client↔e2e chain is shipped.

**Original ask:** Replace the empty-value button-name heuristic in `parseURLEncodedForm()` with an explicit client-side action signal; the heuristic has false positives (cleared text inputs, empty selects) that the ambiguous-guard workaround only papers over.

**Delivered:**

| Step | Repo | SHA / PR | What |
|---|---|---|---|
| Proposal | livetemplate | [#401](https://github.com/livetemplate/livetemplate/pull/401) | `docs/proposals/explicit-submitter.md` rewritten for the alpha/no-users end state: two phases, no compat scaffolding, heuristic kept as permanent no-JS fallback |
[...]

PR #382, opened 2026-05-03
docs(proposals): explicit submitter on the wire (#237)

adnaan:
## Summary

Writes the design proposal at `docs/proposals/explicit-submitter.md` for replacing the empty-value button-name heuristic (issue #237, P1) with an explicit `lvt-submitter` wire field.

**No code changes.** This is design-only — the proposal surfaces the approach for review before any implementation PR lands.

## Why now

After PR #327 landed the `detectSubmitButtonName` helper extraction, the heuristic is now consolidated at one site, which makes the migration target much cleaner. The client *already* captures `SubmitEvent.submitter` natively (`dom/event-delegation.ts`) — the data exists on the wire side; the server just isn't reading it yet.

## What the proposal covers

- Why the heuristic is fragile (ambiguity, user-input collisions, no-explicit-submit case)
- The current client capture path (already shipping)
- Wire format extension: reserved `lvt-submitter` field on HTTP, top-level `submitter` key on WS action payloads
- Server resolution order: `lvt-action` → `lvt-submitter` → heuristic (fallback) → `""`
- Shared `resolveSubmitterFallback(msg)` helper called from all four parser call sites (HTTP JSON, WS, urlencoded, multipart)
- 4-phase migration plan ending with heuristic removal in v0.9.0 (alongside the `lvt-no-intercept` shim removal already planned)
- 6 open questions / risks flagged for reviewer input
- Optional `lvt-form:emit-submitter` directive for lite-JS forms (with CSP caveat)

## Verification

- [x] `go test ./...` passes (no code changed; pre-commit hook ran the full gate)
- [x] Code anchors are grep-style per the AI Code Review Workflow guidance in CLAUDE.md ("Prefer grep anchors over hardcoded line numbers in code references")
- [x] All factual claims about `internal/send/message.go` cross-checked against the live code

Refs #237. **No issue close** — implementation lands in a follow-up PR after this proposal is reviewed.


adnaan:
## Convergence note

This proposal has been through 9 rounds of bot review and ~150 review comments addressed in 8 fix-up commits. Each round has surfaced fewer and smaller findings — round 1 had structural design gaps, rounds 5-9 are precision tweaks and implementation-detail clarifications. Per CLAUDE.md "Bot review loop convergence" guidance, the proposal is now stable enough to ship as-is.

**Items I'm deferring to the Phase 1 implementation PR rather than addressing in this proposal:**

1. **Cross-repo deployment ordering note** (server-first vs client-first). This is operational guidance better placed in the Phase 1 release notes, not the proposal.
2. **Phase 4 ADR gating Phase 3 as well as Phase 4.** The Phase 3 deprecation warning is informational; it doesn't need to block on the ADR. The ADR is required before Phase 4 because that's when behavior actually changes.
3. **Explicit "no `ctx.Submitter()` API" callout.** Already implied by "`msg.Submitter` is resolved into `msg.Action` before routing". A separate callout would be redundant.
4. **`resolveSubmitterFallback` ordering in `parseMultipartForm`.** The proposal already says "place at the same control-flow position as `lvt-action`", which an implementer follows for both the read and the resolve call.
5. **Skiplist asymmetry doc-comment guidance** (round 8 #4). The implementer adding the doc comment will see the asymmetric `actionFields` and document it accurately.
6. **Phase 3 release notes Phase 4 callout** (round 8 #5). Release-notes authoring belongs to the Phase 3 PR.
7. **`sync.Map` rate-limit cache key encoding scheme** (round 7 #7). Implementation detail; the proposal says "tuple", encoding is the implementer's choice.
8. **`resolveSubmitterFallback` function-vs-method idiom** (round 7 #6). Pure Go style; the implementer can pick.

**Closing the bot loop:** if subsequent reviews surface only restatements of items in this list (or pure style/wording suggestions), I will not push further fixups. Substantive new functional issues will of course be addressed.

The proposal is design-only and ready for human review.

PR #388, opened 2026-05-08
fix(mount): auto-wire FormSchema into Context (#236)

### LINKED ISSUE #236: FOLLOW-UP: Auto-wire FormSchema into Context from template statics
## Context

This follow-up task was identified during the review of PR #233.

**PR Title:** feat: progressive complexity model for form handling
**Suggested by:** @claude-bot
**Priority:** High

## Task Description

`ctx.ValidateForm()` checks `c.formSchema`, but nothing in the framework populates it automatically. Users must call `ctx.WithFormSchema(ExtractFormSchema(statics))` manually, but there is no public API path from a handler to the template statics.

As written, `ctx.ValidateForm()` always returns nil for real users — the validation silently does nothing.

**Fix:** Auto-wire the schema in the HTTP/WebSocket dispatch path in `mount.go`, probably alongside the `applyDefaultAction` call. Extract statics from the template's cached tree and call `ExtractFormSchema()` once per template (cache the result).

---

adnaan:
## Summary

- `ctx.ValidateForm()` was silently a no-op for real users — it gated on `c.formSchema`, but nothing in the framework populated that field, and the public API offered no path from a handler to the template statics needed for `ExtractFormSchema`. Most code that called `ValidateForm` got `nil` back regardless of input.
- Cache the `FormSchema` once on `Template` at parse time, derived from `templateStr` with `{{...}}` directives stripped. Dynamic name attributes collapse and are discarded, matching the documented `ExtractFormSchema` limitation. `Clone()` shares the pointer like the other immutable parse-time caches (`cachedParseTemplate`, `cachedBodyContent`).
- Attach the cached schema to `actionCtx` in `mount.go` before action dispatch on the three dispatch paths that carry user form input: WS event loop, HTTP path, and the dispatched-action path. Handlers that call `WithFormSchema` explicitly still win because the chained `With*` returns a fresh context.

## Test plan

- [x] `TestMount_AutoWiresFormSchema_HTTP` posts a malformed form to a controller that calls `ctx.ValidateForm()` without manual `WithFormSchema` setup and asserts both per-field errors round-trip in the response (proves the auto-wire fires on the HTTP path).
- [x] `TestMount_AutoWiresFormSchema_WS` opens a WebSocket, dispatches a `Submit` action with malformed data, and asserts `meta.success=false` plus per-field errors come back over the wire (proves the auto-wire fires on the WS event-loop path — the more common real-world path).
- [x] `TestMount_AutoWiresFormSchema_NoFormFields` confirms `formSchema` stays `nil` for templates without inputs (auto-wire branch is a no-op for non-form templates).
- [x] `TestExtractFormSchemaFromTemplateStr_StripsDirectives` covers the `{{...}}` stripping helper.
- [x] Full suite: `GOWORK=off go test ./... -timeout=300s` — passes.
- [x] Pre-commit hook (`go fmt` + `golangci-lint` + full tests) — passes.

Closes #236

### DIFF
--- a/formvalidation.go
+++ b/formvalidation.go
@@ -33,10 +33,9 @@ type FormSchema struct {
 // inputAttrRegex matches HTML input/textarea/select elements and captures their attributes.
 var inputAttrRegex = regexp.MustCompile(`<(?:input|textarea|select)\b([^>]*)>`)
 
-// Strips Go template actions before HTML attribute extraction.
 var templateDirectiveRegex = regexp.MustCompile(`(?s)\{\{.*?\}\}`)
 
-// Pre-blanks partially-dynamic names like name="user_{{.ID}}" so they don't collapse to misleading literals after the directive strip.
+// name attrs containing any directive are blanked so partial-dynamic forms don't collapse to spurious literal rules.
 var dynamicNameAttrRegex = regexp.MustCompile(`(?s)(\bname\s*=\s*")[^"]*\{\{.*?\}\}[^"]*"`)
 
 func extractFormSchemaFromTemplateStr(templateStr string) *FormSchema {
--- a/formvalidation.go
+++ b/formvalidation.go
@@ -33,30 +33,13 @@ type FormSchema struct {
 // inputAttrRegex matches HTML input/textarea/select elements and captures their attributes.
 var inputAttrRegex = regexp.MustCompile(`<(?:input|textarea|select)\b([^>]*)>`)
 
-// templateDirectiveRegex matches Go template actions like {{...}} so we can strip
-// them from the raw template source before form-attribute extraction. Dynamic
-// attribute values (e.g. name="{{.Field}}") collapse to name="" and are then
-// discarded by ExtractFormSchema, matching the documented limitation that
-// dynamic field names are not auto-detected.
+// Strips Go template actions before HTML attribute extraction.
 var templateDirectiveRegex = regexp.MustCompile(`(?s)\{\{.*?\}\}`)
 
-// dynamicNameAttrRegex matches a name attribute whose double-quoted value
-// contains a {{...}} template directive (fully or partially dynamic). We blank
-// the value before stripping directives so partially-dynamic names like
-// name="user_{{.ID}}" do not collapse to a misleading literal (name="user_")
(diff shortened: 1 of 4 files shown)


adnaan:
Round 4 review explicitly contradicts round 3 on the same line of code, so declining further pushes per CLAUDE.md "AI Code Review Workflow" convergence rule (round 3 said remove, round 4 says add back).

**1. `handleServerActionMessage` formSchema wiring** — declining

- **Round 3 (this PR review)**: > "`handleServerActionMessage` is triggered by PubSub messages from other server instances — it's a programmatic server-to-server call, not a form submission. ... it's semantically wrong and should either be intentional (and documented in the PR) or removed."
- **Round 4 (this PR review)**: > "Functional gap: `handleServerActionMessage` is not wired ... A form-validating action reachable via server-action broadcast would still be a silent no-op."

Both framings are internally valid; round 3's removal rationale stands. Server-to-server `msg.Data` is not HTML form data — wiring `formSchema` would let a handler that misuses `ValidateForm()` from a server-action context silently produce wrong validation results. The hypothetical "form-validating action reached via server-action broadcast" round 4 raises is a real but separable concern; if it surfaces in practice, the right fix is a typed assertion at the boundary (or a separate `ValidateServerActionData`), not retroactively wiring an HTML-form schema into a non-HTML path.

**2. `WithFormSchema` ordering inconsistency in `handleDispatchedAction`** — declining

Bot acknowledges "no ordering dependency." This is a cosmetic nit and CLAUDE.md ("project guidance trumps bot suggestions when they conflict") favors leaving it; reshuffling would itself be churn.

**3. `(?s)\{\{.*?\}\}` regex edge case (e.g. `{{if eq .Status "}}"}}`)** — acknowledged as a documented limitation

Bot itself noted "this can't produce a false `name` match in form attributes." The non-greedy match is correct for the common case; a directive containing literal `}}` in a string argument is exotic and would degrade to a parse error elsewhere in the template pipeline before this regex sees it.

**4. WS test coverage** — already added in commit 04540a2e (`TestMount_AutoWiresFormSchema_WS`); included in this PR.

No further pushes intended. The bot's "two action items before merge" framing in round 4 is contradicted by round 3, so the maintainer's call wins.

PR #389, opened 2026-05-08
fix(pubsub): release mutex before Redis network calls (#215)

### LINKED ISSUE #215: FOLLOW-UP: Release mutex before Redis network calls in subscribeTo/reconnect (from PR #213)
## Context

This follow-up task was identified during the review of PR #213.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/213
**PR Title:** fix: track dynamic pubsub subscriptions for reconnect and wire into mount

## Task Description

`subscribeTo()` and `reconnect()` hold `b.mu.Lock()` for the entire duration of Redis network calls (`pubsub.Subscribe()`) and `time.Sleep()` (reconnect delay). This blocks all concurrent publish, subscribe, and close operations.

**Impact:**
- `subscribeTo()`: One slow Redis SUBSCRIBE blocks all other subscriptions and all publishing
- `reconnect()`: Default 1-second sleep + Redis call = guaranteed 1+ second full broadcaster stall

**Proposed fix for `subscribeTo()`:** Use check-lock-check pattern:
1. `RLock` for dedup check (fast path)
2. Release lock, perform Redis SUBSCRIBE outside any lock
3. `Lock` only to update `subscribedChannels`
4. Accept harmless duplicate Redis SUBSCRIBE (idempotent) in the race window

**Proposed fix for `reconnect()`:**
1. Close old pubsub under lock
2. Release lock
3. Sleep (reconnect delay)
4. Re-acquire lock, check `b.closed`
5. Re-subscribe under lock (or outside lock with same pattern)

---

adnaan:
## Summary

- `subscribeTo()` (via `trySubscribe`) now uses a check-lock-check pattern: RLock to fast-path-check + snapshot `b.pubsub`, release, run Redis SUBSCRIBE outside any lock, then Lock+verify-pubsub-not-swapped before recording the channel. Duplicate SUBSCRIBE in the racing window is harmless (idempotent on Redis).
- `reconnect()` now closes the stale pubsub and snapshots the channel list under lock, releases, sleeps via `select` on `b.ctx.Done()` so `Close()` can short-circuit a long delay, performs SUBSCRIBE/Receive outside any lock, then re-acquires to install the new pubsub (tearing it down if `Close()` raced ahead).
- Three new regression tests demonstrate concurrent `PublishGlobal` is no longer blocked while a peer subscribe or reconnect is mid-network-call, and that reconnect honors context cancellation.

## Test plan

- [x] `go test ./pubsub/... -race -timeout=120s` — all 22 tests pass (including 3 new ones).
- [x] `go test ./... -timeout=300s` — full suite green.
- [x] Verified each new test fails on a deliberately reverted version of the corresponding fix (subscribeTo regression test caught the held-lock; reconnect regression test reported `PublishGlobal blocked for 757ms during reconnect sleep`).
- [x] golangci-lint with the project gate (errcheck/govet/ineffassign/staticcheck/unparam/unused) — 0 issues.

Closes #215

### DIFF
--- a/pubsub/redis.go
+++ b/pubsub/redis.go
@@ -3,6 +3,7 @@ package pubsub
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
@@ -49,6 +50,7 @@ type RedisBroadcaster struct {
 	wg                  sync.WaitGroup
 	mu                  sync.RWMutex
 	closed              bool
+	reconnecting        bool                // True while reconnect() is in its sleep/SUBSCRIBE window (pubsub is nil)
 	reconnectDelay      time.Duration       // Delay before reconnecting after subscription failure (default: 1s)
 	subscribedChannels  map[string]struct{} // Tracks dynamic channel subscriptions for reconnect replay
 }
@@ -354,6 +356,19 @@ func (b *RedisBroadcaster) SubscribeToGroupAction(groupID string) error {
 // network call. Production code leaves this nil.
 var subscribeHook func()
 
+// reconnectHook is a test-only hook invoked from reconnect() immediately
+// after the lock has been released and before the reconnect-delay sleep.
+// It allows tests to deterministically wait for reconnect() to enter its
+// no-lock window without relying on time-based synchronization.
+// Production code leaves this nil.
+var reconnectHook func()
+
+// errPubsubNotReady is a sentinel signaling that pubsub is currently nil
+// because reconnect() is in progress. It's distinct from "broadcaster is
+// closed" (permanent) and lets subscribeTo() extend its retry window to
+// cover the reconnect-delay rather than failing fast in ~200ms.
+var errPubsubNotReady = fmt.Errorf("pubsub not ready (reconnect in progress)")
+
 // subscribeTo subscribes to a Redis channel with dedup and retry.
 //
 // The lock is released across the Redis SUBSCRIBE network call so concurrent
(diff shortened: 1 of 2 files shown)


adnaan:
Addressed bot review comments in 86aa8496 (`fix: address bot review comments on #389`).

**Applied:**

1. **Functional issue: subscribe-during-reconnect window** (Claude + Copilot, same issue). Added `b.reconnecting` flag set under the same lock that nils `b.pubsub` and cleared via deferred unlock on every exit path. `trySubscribe` now returns a sentinel `errPubsubNotReady` only when reconnecting is truly in progress; `subscribeTo` extends its retry deadline by `reconnectDelay + 2s` for that case so subscribes arriving mid-reconnect wait for the new connection rather than failing fast in ~200ms. The "permanent" case (`Subscribe` never called) still fails inside the normal 3-attempt budget, so `TestSubscribeTo_ExhaustsRetriesWhenPubSubNil` continues to pass quickly.

2. **Regression test** for the contract above: `TestSubscribeTo_SucceedsDuringReconnectWindow` (Copilot ask). Uses the new `reconnectHook` seam to deterministically wait until reconnect has nilled the pubsub, then asserts `SubscribeToGroup` succeeds.

3. **Timing fragility** in `TestReconnect_DoesNotBlockConcurrentPublish` (Claude). Replaced `time.Sleep(50ms)` with the same hookEntered/release channel pattern used in the sibling test (now via a new `reconnectHook` package var, mirroring `subscribeHook`).

4. **`broadcaster.cancel()` -> `broadcaster.Close()`** in `TestReconnect_InterruptibleByContextCancel` (Claude). Close() now runs in a goroutine and is drained at the end so its blocking-on-the-running-reconnect doesn't hang the test; this matches production semantics where ctx-cancel is always paired with `b.closed = true`.

**Declined:**

- **`subscribeHook` package var -> struct field** (Claude flagged "not a blocker"). Keeping the package var because: (a) `*redis.PubSub` is a concrete type, so the cleaner alternative — wrapping `UniversalClient.Subscribe` to inject from there — doesn't actually reach `b.pubsub.Subscribe` in `trySubscribe`; (b) tests don't call `t.Parallel()` and use `prevHook := subscribeHook; defer func() { subscribeHook = prevHook }()` to scope state; (c) a struct field would require an unexported setter and percolate the test seam through the public API surface for no behavioral benefit. The `reconnectHook` added in this round follows the same pattern.

Verified: `go test ./pubsub/... -race -count=10 -timeout=300s` passes clean for all four affected tests (`TestSubscribeTo_DoesNotBlockConcurrentOperations`, `TestReconnect_DoesNotBlockConcurrentPublish`, `TestReconnect_InterruptibleByContextCancel`, `TestSubscribeTo_SucceedsDuringReconnectWindow`).

adnaan:
Addressed Round 2's remaining minor finding in `f9246eda` (`fix(pubsub): separate notReady and normal-error retry counters`).

Split the unified `attempt` counter into two: `iteration` gates the inter-attempt sleep (so any retry, notReady or normal-error, waits `retryDelay`), while `normalAttempts` gates the `maxAttempts` break and is only incremented on non-notReady errors. With this split, notReady spins (up to `~reconnectDelay/retryDelay` of them) no longer silently consume the 3-attempt budget reserved for real subscribe failures, and the "after 3 attempts" error message now reflects what was actually retried.

Verified the discriminator: `TestSubscribeTo_ExhaustsRetriesWhenPubSubNil` still fails fast (`<1s`). The first iteration sees `b.reconnecting == false` so `trySubscribe` returns `"not subscribed"` (not `errPubsubNotReady`), the loop falls into the normal branch, and `normalAttempts` climbs `1→2→3` across the 3 iterations as before.

Pre-commit gate passed: `gofmt`, `golangci-lint`, full `go test ./...` clean.

PR #391, opened 2026-05-09
fix(pubsub): gate handler registrations on Subscribe() success (#357)

### LINKED ISSUE #357: FOLLOW-UP: Gate pub/sub handler registrations on Subscribe() success (from PR #355)
## Context

This follow-up task was identified during the review of PR #355.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/355
**PR Title:** fix: eliminate race in Redis pub/sub init and add subscription retry
**Comment type:** PR review comment (raised in review rounds 3-5)

## Task Description

If `Subscribe()` fails at startup (e.g. Redis unreachable), `template.go` logs the error but continues to call `SubscribeServerActions()` and `SubscribeGroupActions()`. Both succeed (they just register handlers), but `processMessages()` never starts, so no pub/sub messages are ever dispatched — silently.

Every WebSocket connection that calls `trySubscribe` will get "not subscribed" (since `b.pubsub` is nil), exhaust its 3 retries, and log errors — one per connection, which can be very noisy.

### Proposed fix

Gate handler registrations on `Subscribe()` succeeding:

```go
if err := mountCfg.PubSubBroadcaster.Subscribe(handler.handlePubSubMessage); err != nil {
    slog.Error("Pub/sub subscriber error — pub/sub disabled", slog.Any("error", err))
    // skip SubscribeServerActions / SubscribeGroupActions
} else {
    if err := mountCfg.PubSubBroadcaster.SubscribeServerActions(...); err != nil { ... }
    ...
}
```

Or add a "pub/sub disabled" log line so operators can distinguish "starting up" from "broken."

## Original Comment

> After a failed `Subscribe()`, `b.pubsub` is nil. `SubscribeServerActions` and `SubscribeGroupActions` just register handlers and succeed. But every subsequent per-connection `SubscribeTo*` call will fail with "not subscribed" and log an Error -- one per WebSocket connection, which could be very noisy. Now that `Subscribe()` errors are a concrete failure path rather than a silent background goroutine, it is worth a follow-up issue on whether the handler should refuse WebSocket connections when pub/sub setup has fully failed.

[View PR #355](https://github.com/livetemplate/livetemplate/pull/355)

---

adnaan:
## Summary

- Gate `SubscribeServerActions` and `SubscribeGroupActions` registrations on the initial `PubSubBroadcaster.Subscribe()` succeeding. Previously, when `Subscribe()` failed (e.g. Redis unreachable at startup), the downstream calls still ran and silently registered handlers in local maps even though `processMessages()` never started.
- Downgrade the failure log from `Error` to a single `Warn` ("cross-instance dispatch disabled") to make the degraded-mode signal clear without the per-connection `Error` spam from per-WS `trySubscribe` retries.
- Adds two unit tests in `handle_test.go` using a fake `Broadcaster + GroupActionBroadcaster` with call counters: one asserts the downstream `Subscribe*` calls are skipped on failure; one asserts they all run on success.

Closes #357

## Test plan

- [x] `GOWORK=off go test ./... -timeout=180s` passes (full suite, including new gate tests)
- [x] `golangci-lint run --enable-only=errcheck,govet,ineffassign,staticcheck,unparam,unused` clean
- [x] Pre-commit hook (gofmt + lint + tests) passes without `--no-verify`
- [x] New tests `TestHandle_PubSubSubscribeFailureSkipsRegistrations` and `TestHandle_PubSubSubscribeSuccessRegistersAll` pass

### DIFF
--- a/handle_test.go
+++ b/handle_test.go
@@ -2820,8 +2820,6 @@ func TestEphemeral_SyncStillWorks(t *testing.T) {
 	}
 }
 
-// fakeBroadcaster is a minimal pubsub.Broadcaster + GroupActionBroadcaster used
-// to verify that handler registrations are gated on Subscribe() success (#357).
 type fakeBroadcaster struct {
 	subscribeErr      error
 	subscribeCalls    int
@@ -2855,11 +2853,6 @@ func (f *fakeBroadcaster) SubscribeGroupActions(_ pubsub.GroupActionHandler) err
 	return nil
 }
 
-// TestHandle_PubSubSubscribeFailureSkipsRegistrations verifies that when
-// PubSubBroadcaster.Subscribe() returns an error (e.g. Redis unreachable),
-// SubscribeServerActions and SubscribeGroupActions are NOT called. Otherwise
-// per-WS trySubscribe calls would log "not subscribed" errors on every connection.
-// See https://github.com/livetemplate/livetemplate/issues/357.
 func TestHandle_PubSubSubscribeFailureSkipsRegistrations(t *testing.T) {
 	fb := &fakeBroadcaster{subscribeErr: errors.New("redis unreachable")}
 
@@ -2885,8 +2878,6 @@ func TestHandle_PubSubSubscribeFailureSkipsRegistrations(t *testing.T) {
 	}
 }
 
-// TestHandle_PubSubSubscribeSuccessRegistersAll verifies that when Subscribe
-// succeeds, both SubscribeServerActions and SubscribeGroupActions are called.
 func TestHandle_PubSubSubscribeSuccessRegistersAll(t *testing.T) {
 	fb := &fakeBroadcaster{}
 
--- a/handle_test.go
+++ b/handle_test.go
@@ -2819,3 +2819,95 @@ func TestEphemeral_SyncStillWorks(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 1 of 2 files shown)


adnaan:
Re: round-3 observations — declining all three as out of scope for this PR:

**Item 1 (inconsistent log levels in else branch):** `SubscribeServerActions`/`SubscribeGroupActions` failures stay at `slog.Error` because they're unexpected — `Subscribe()` succeeded, then a specific channel registration failed. The outer `Subscribe()` failure is `Warn` because it's an expected operational mode (Redis unreachable on boot). The asymmetry is intentional: "expected boot-time degradation" vs "unexpected per-channel failure after pump is running".

**Item 2 (`SubscribeServerActions` failure doesn't short-circuit `SubscribeGroupActions`):** Bot itself notes this is pre-existing and outside scope. The PR's contract is "gate downstream registrations on `Subscribe()` success" — making them mutually-gating is a separate refactor. Could be a follow-up if it manifests as a real operational issue; closing this without one.

**Item 3 (missing test for broadcaster without `GroupActionBroadcaster`):** Branch-coverage gap on the type-assertion guard, low priority per the bot itself. Out of scope for this PR which targets the `Subscribe()` gate, not exhaustive type-assertion coverage. Per CLAUDE.md "no features beyond what the task requires".

Issue #337, opened 2026-04-14
FOLLOW-UP: Depth/rate limit on chained Session.TriggerAction dispatches (from PR #336)


adnaan:
## Context

This follow-up was identified during review of PR #336 (`fix: wire Session.TriggerAction into lifecycle contexts`), which wired `ctx.Session()` into every action context — including dispatched actions via `handleDispatchedAction` and server-initiated actions via `handleServerActionMessage`.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/336
**Suggested by:** @claude[bot] (Claude Code Review, rounds 3 and 4)

## The problem

A controller action handler reached via the dispatch queue can legitimately call `ctx.Session().TriggerAction(...)`, which enqueues another dispatch on the same connection's event loop. There is currently **no depth counter, no rate limiter, and no observability log** on this recursive path.

Concrete failure mode: a handler that calls `ctx.Session().TriggerAction("self", nil)` — whether intentional or by mistake (e.g. a typo in a conditional) — will loop at full speed until the WebSocket disconnects. Each hop runs through the connection's event loop (so there's no stack-overflow risk), but the loop pins a goroutine per-connection and holds `connSt.messagesMu` in rapid cycles until the session tears down.

The existing `pendingBroadcasts` drop in `handleDispatchedAction` and `handleServerActionMessage` protects against **`ctx.BroadcastAction`** storms, but `ctx.Session().TriggerAction` bypasses that guard by design — it goes straight to `conn.EnqueueDispatch` rather than into the pending-broadcasts slice. The asymmetry is documented inline at `mount.go:1466` (the `WithSession` wiring in `handleDispatchedAction`) as an acknowledged footgun.

## Why it wasn't fixed in PR #336

Explicit caller intent is required (a typo in one place is rare), and the recursive pattern has legitimate uses (rescheduling, retry-with-backoff). Adding a framework-level depth limit has design trade-offs — hardcoded limits can be too aggressive for legitimate use cases, and the right default is unclear. The PR deferred this as \"reasonable future work but out of scope for this fix\" in the round-3 commit message.

### Option A: Observability-only (lowest risk, smallest change)

Add a `slog.Debug` line at each dispatch enqueue from a dispatched context, so infinite loops are detectable in production logs without changing behavior. This is the minimum bar Claude suggested in round 3.

```go
// In localSession.TriggerAction, when called from a context that was
// itself built from a dispatched action:
[...]

PR #393, opened 2026-05-09
feat(session): observability log on chained TriggerAction (#337 Option A)

adnaan:
## Summary

- Adds a `slog.Debug` line each time `Session.TriggerAction` is called from a context that was itself built from a dispatched action — making infinite-loop chained TriggerAction patterns (e.g., `ctx.Session().TriggerAction("self", nil)`) detectable **when debug logging is enabled** (default handler level is `Info`, so operators must lower their level to surface this).
- Implementation: new unexported `localSession.fromDispatched` flag + `newLocalSessionFromDispatched` constructor; wired into `handleDispatchedAction` (mount.go ~1531) and `handleServerActionMessage` (mount.go ~1883). Initial WS connect, HTTP POST action, and upload completion paths use `newLocalSession` and remain silent.
- Options B (depth counter) and C (rate limiter) are explicitly out of scope per the issue's design-tradeoff caveats and tracked separately.

## Test plan

- [x] `go test ./... -timeout=300s` passes
- [x] `go test -race -count=10` confirms no data races on the slog capture or controller state
- [x] Pre-commit hook passes (no `--no-verify`)
- [x] `session_impl_test.go` exercises both the dispatched-context (logs) and non-dispatched (silent) paths

Addresses #337 (Option A); Options B (depth counter) / C (rate limiter) tracked separately.

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -1525,9 +1525,8 @@ func (h *liveHandler) handleDispatchedAction(connSt *connState, connection *sess
 	// allowed — each hop runs through a connection event loop, so the
 	// only unbounded failure mode is a handler that recursively
 	// re-triggers itself, which is a caller bug rather than framework
-	// amplification. The "from dispatched" constructor variant flags the
-	// session so chained TriggerAction calls log a debug line, making
-	// such recursion visible in logs (#337).
+	// amplification. The dispatched-flagged session emits an
+	// observability log on chained TriggerAction (#337).
 	ctx = ctx.WithSession(newLocalSessionFromDispatched(h, connSt.groupID))
 	if schema := h.config.Template.formSchema; schema != nil {
 		ctx = ctx.WithFormSchema(schema)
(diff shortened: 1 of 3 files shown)


adnaan:
Round-2 items addressed in 91f84039:

- **Item 1 (log level vs production-visibility claim)**: Kept `slog.Debug` per the original issue spec; updated PR description to say "detectable when debug logging is enabled" so the expectation matches the level. Switching to Info/Warn would be noisier given chaining is sometimes intentional.
- **Item 2 (syncBuf placement broke godoc)**: Moved syncBuf above the test function's doc comment so godoc attaches correctly.
- **Item 3 (slog.SetDefault parallel test risk)**: Added explicit warning comment on the test function about not using t.Parallel().
- **Item 4 (doc comment duplication)**: Trimmed the 4-line constructor doc to a single-line summary; the field comment remains canonical.

**Item 5 (error-path test for fromDispatched=true + empty groupID)**: declining as scope creep. The error guard at `session_impl.go:115` is the same code path regardless of `fromDispatched` — the existing `TestLocalSession_TriggerActionWithEmptyGroupIDReturnsError` covers the guard. Adding a parallel test for the dispatched variant would only verify the log line ordering, which the bot itself flagged as low-priority. Per CLAUDE.md "no features beyond what the task requires".

adnaan:
Round-3 review:

**Item 1 (over-explained comments)**: addressed in 369a575d. Trimmed the 4-line struct field comment to one line, dropped the redundant constructor comment, condensed the inline observability comment, and removed the WHAT-prefix from both mount.go comments. The issue refs (#337) are preserved per the bot's own guidance.

**Item 2 (drop `TestLocalSession_FromDispatchedFlag`)**: declining. The structural fallback was an explicit choice — `session_impl_test.go:380-387` documents "if the integration test above is brittle, this still validates the wiring change." The integration test is sensitive to slog default-handler config + WS timing in CI; a structural test that survives those issues has positive value. Type-assertion coupling is the cost of the defense; if the implementation moves the flag, the linker catches it (the test stops compiling), not silent rot.

**Item 3 (count-based log assertion + WS-event sync instead of sleep)**: declining as optional hardening. Bot itself says "the 5-second outer timeout provides enough slack in practice". The sleep is short (50ms) and the outer timeout is 5s — adding event-based sync would be more robust but isn't needed at the current flake rate. Closing as scope creep per CLAUDE.md "don't add features beyond what the task requires".

Issue #251, opened 2026-03-26
Implement Change() auto-inference for Tier 1 live updates


adnaan:
## Summary

The `Change()` method convention is documented in the progressive complexity guide (Section 10) as a Tier 1 feature where having a `Change()` method on a controller auto-wires input change events without any `lvt-*` attributes. Currently this is **not implemented**.

## Current State

- **Server-side dispatch works**: If the client sends `{action: "change", data: {...}}`, `DispatchWithState` routes it to `Change()` via method name resolution
- **Client-side auto-binding does NOT exist**: The JS client does not detect the presence of a `Change()` method and auto-wire input events
- **Explicit Tier 2 approach works today**: Using `lvt-input="change" lvt-debounce="300"` on inputs manually routes to `Change()` — but this is Tier 2, not the intended Tier 1 auto-inference

## Desired Behavior (Tier 1)

If a controller has a `Change()` method, the framework should automatically:

1. Detect input fields in the template that are bound to state (e.g., `value="{{.Name}}"`)
2. Auto-wire `input` or `change` events on those fields to send the changed value to the server
3. Route to the `Change()` method on the controller
4. Apply sensible default debounce (e.g., 300ms)

```html
<!-- No lvt-* attributes needed -->
<form method="POST">
    <input name="Name" value="{{.Name}}">
    <div class="preview">Hello, {{.Name}}!</div>
    <button type="submit">Save</button>
</form>
```

```go
// Having this method auto-wires input change events
func (c *Controller) Change(state State, ctx *livetemplate.Context) (State, error) {
    if ctx.Has("Name") { state.Name = ctx.GetString("Name") }
    return state, nil
}
```

## Implementation Scope

This likely requires changes in both repositories:

### Client (`github.com/livetemplate/client`)
- Detect whether the controller supports `Change()` (server could communicate this in initial render metadata)
- Auto-bind `input` events with debounce on form fields that have dynamic `value` attributes
- Send change actions to server

### Server (`github.com/livetemplate/livetemplate`)
- Communicate to client whether controller has a `Change()` method (e.g., in initial render response metadata)
- Server-side dispatch already works — no changes needed there

## Related

- Progressive complexity guide Section 10: `docs/guides/progressive-complexity.md`
- Controller pattern reference: `docs/references/controller-pattern.md`
- PR #250 (docs update) depends on this being implemented or clearly documented as not-yet-implemented

PR #395, opened 2026-05-09
feat(capabilities): advertise validate, upload, progressive_enhancement (#252)

### LINKED ISSUE #252: Extend capabilities metadata for upload, validate, and progressive enhancement
## Summary

Issue #251 introduces a `capabilities` field in the initial render response metadata (`ResponseMetadata.Capabilities []string`). This tracks future capabilities beyond `"change"` that could be communicated to the client.

### Method-based (auto-detected via reflection at `Handle()` time)

- **`"validate"`** — Controller has `Validate()` method. Client could trigger server-side validation on blur/change without full submit.
- Any future conventional method that requires client-side auto-wiring.

### Config-based (from `mountConfig`/`TemplateConfig`)

- **`"upload"`** — Upload configs are present (`len(UploadConfigs) > 0`). Client could enhance `<input type="file">` with drag-and-drop, progress indicators, etc.
- **`"progressive_enhancement"`** — Non-JS form fallback is enabled. Client knows server handles plain POST.

## Design Principle

A capability should only be in the list if the **client changes its behavior** based on it. For example, `"submit"` wouldn't need to be here since it's always the default.

## Related

- #251 — Implements `"change"` capability and the `Capabilities` field on `ResponseMetadata`

adnaan:
## Summary

`ResponseMetadata.Capabilities` was set up by #251 to advertise server features the client can adapt to. Currently only `"change"` is advertised. Extend with three more (also from #252):

- **`"validate"`** — detected via reflection; controller has a `Validate()` method
- **`"upload"`** — detected from `mountCfg.UploadConfigs` (any `WithUpload` registered)
- **`"progressive_enhancement"`** — from `mountCfg.ProgressiveEnhancement` (default: `true`)

Detection is centralized in `dispatch.go`'s new `detectCapabilities()` helper so wire order is deterministic. Returns nil when empty so `omitempty` drops the field.

### Behavior change to call out

**`progressive_enhancement` is now advertised by default** because `mountCfg.ProgressiveEnhancement` defaults to `true`. Any existing application that previously got an empty capabilities list when no `Change()` method exists will now receive `["progressive_enhancement"]`. The pre-existing `TestHandle_NoCapabilitiesInWebSocketWithoutChangeMethod` test was updated to use `WithProgressiveEnhancement(false)` to keep its "no capabilities" assertion valid — the same fix any downstream test asserting an empty capabilities list will need to apply. (See companion PR [examples#98](https://github.com/livetemplate/examples/pull/98) which made the live-preview test tolerant.)

## Test plan

- [x] `TestHandle_CapabilitiesValidateDetected` / `_ValidateOmittedWithoutMethod`
- [x] `TestHandle_CapabilitiesUploadDetected` / `_UploadOmittedWithoutConfig`
- [x] `TestHandle_CapabilitiesProgressiveEnhancementDefault`
- [x] Pre-existing `TestHandle_NoCapabilitiesInWebSocketWithoutChangeMethod` updated to use `WithProgressiveEnhancement(false)` to keep the "no capabilities" assertion valid
- [x] `go test ./... -timeout=300s` passes
- [x] Pre-commit hook passes (no `--no-verify`)
- [x] Cross-repo CI passes (after companion examples-repo PR #98 merged)

Closes #252 (extends #251)

### DIFF
--- a/dispatch.go
+++ b/dispatch.go
@@ -13,8 +13,26 @@ import (
 // ErrMethodNotFound is returned when Dispatch cannot find a method matching the action.
 var ErrMethodNotFound = errors.New("method not found for action")
 
-// CapabilityChange is the capability name for controllers with a Change() method.
-const CapabilityChange = "change"
+// Capability names advertised in ResponseMetadata.Capabilities so the client
+// can adapt its behavior to features the server supports.
+//
+// A capability should only be in the list if the client changes its behavior
+// based on it. Detection happens once at Handle() time.
+const (
+	// CapabilityChange is set when the controller has a Change() method,
+	// detected via reflection (see HasActionMethod).
+	CapabilityChange = "change"
+	// CapabilityValidate is set when the controller has a Validate() method,
+	// detected via reflection (see HasActionMethod).
+	CapabilityValidate = "validate"
+	// CapabilityUpload is set when the template has at least one upload
+	// configuration registered via WithUpload.
+	CapabilityUpload = "upload"
+	// CapabilityProgressiveEnhancement is set when non-JS form fallback is
+	// enabled (default: enabled). Clients can use it to know the server
+	// handles plain POSTs end-to-end.
+	CapabilityProgressiveEnhancement = "progressive_enhancement"
+)
 
 // DispatchError provides context about a failed dispatch.
 type DispatchError struct {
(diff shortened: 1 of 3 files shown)


adnaan:
Round-2 review:

**Item 2 (progressive_enhancement default change should be called out)**: addressed in the updated PR description above.

**Item 1 (WS-path coverage for new capabilities)**: declining as scope creep. Bot itself notes this is "functionally fine — all capabilities flow through the same `h.config.Capabilities` slice set once in Handle()". The HTTP path test already validates the slice contents; adding a WS-mirror test would only assert the same data on a different transport. Per CLAUDE.md "don't add features beyond what the task requires".

**Item 3 (direct unit test for detectCapabilities)**: declining as scope creep. Bot notes "the integration coverage is solid". detectCapabilities is a small pure function; the integration tests already exercise every branch (Validate present/absent, UploadConfigs present/absent, PE on/off). Adding a redundant unit test would just duplicate coverage.

Both declined items are bot-flagged as "not blockers". Merging once cross-repo passes.

PR #396, opened 2026-05-09
feat(send): explicit submitter field on the wire (#237 Phase 1)

adnaan:
## Summary

Implements **Phase 1** of [`docs/proposals/explicit-submitter.md`](https://github.com/livetemplate/livetemplate/blob/main/docs/proposals/explicit-submitter.md) (PR #382): server-side acceptance of an explicit submitter field on the wire, replacing the empty-value button heuristic in collision-prone cases.

**Backwards-compatible.** Old clients keep working via the existing heuristic, which stays as a fallback. New clients (Phase 2, separate PR) will populate `lvt-submitter` (forms) and `submitter` (WS).

Closes the server-side portion of #237 (P1 correctness). Phase 2 (client emits) and Phase 3 (deprecation warning) are separate follow-ups.

## What changed

- `ActionMessage` gains `Submitter string \`json:"submitter,omitempty"\``
- New `resolveSubmitterFallback(msg)` helper: `msg.Action ← msg.Submitter` when `Action` is empty (idempotent, safe to call from any parser)
- All four parsers call it: `ParseActionFromHTTP` (JSON branch), `ParseActionFromWebSocket`, `parseURLEncodedForm`, `parseMultipartForm`
- `lvt-submitter` added to `actionFields` skiplist (urlencoded + multipart) so it isn't echoed back as user data
- In `parseMultipartForm`, `lvt-submitter` is read **before** the `jsonDataParsed` branch (same control-flow position as `lvt-action`) so it takes precedence regardless of whether a JSON `data` envelope is present
- Resolution-order doc comment added to `parseMultipartForm` mirroring `parseURLEncodedForm` — both now document: `lvt-action → lvt-submitter → heuristic → ""`

## Out of scope

- **Phase 2**: client emits `lvt-submitter` (HTTP) and `submitter` (WS). Cross-repo work in `livetemplate/client`.
- **Phase 3**: deprecation warning on the heuristic with rate-limited logging + `LVT_FORM_SUBMITTER_COMPAT=silent` env-var suppression.
- **Phase 4**: heuristic removal — gated on the no-JS support decision (see proposal Q3).
- **`lvt-form:emit-submitter` directive**: Phase 2/lite-JS opt-in.

## Test plan

- [x] New tests in `internal/send/message_test.go` covering all three content types (JSON, urlencoded, multipart)
- [x] Collision cases: legitimate empty input + explicit submitter, two empty fields + explicit submitter, `<button name="action">` resolution
- [x] Precedence: `lvt-action` always wins over `lvt-submitter`
- [x] WS path: `action="" + submitter="X"` → action becomes `"X"`; `action="Y" + submitter="X"` → action stays `"Y"`
- [x] `lvt-submitter` not echoed back into `msg.Data`
- [x] `go test ./internal/send/...` passes (with and without `-race`)
- [x] Full test suite passes via pre-commit (300s timeout, includes pubsub Docker tests)
- [x] Linters pass: errcheck, govet, ineffassign, staticcheck, unparam, unused

### DIFF
--- a/internal/send/message.go
+++ b/internal/send/message.go
@@ -12,6 +12,9 @@ import (
 const wsTextMessage = 1 // RFC 6455 text message type (mirrors WSTextMessage in root package)
 
 // ActionMessage represents an action message from the client (internal protocol).
+// Wire-format note: HTTP form paths use the form key "lvt-submitter"; JSON
+// (HTTP and WS) paths use the top-level key "submitter". Different conventions
+// for different envelopes — form fields take the lvt- prefix, JSON keys do not.
 type ActionMessage struct {
 	Action    string                 `json:"action"`              // Action name, may include store prefix (e.g., "counter.increment")
 	Submitter string                 `json:"submitter,omitempty"` // Optional explicit submitter name (e.g. SubmitEvent.submitter.name); used as Action when Action is empty.
@@ -118,7 +121,7 @@ func parseMultipartForm(r *http.Request) (ActionMessage, error) {
 	// Get action from lvt-action field (explicit routing for progressive enhancement)
 	msg.Action = r.FormValue("lvt-action")
 	// Capture explicit submitter (client-emitted SubmitEvent.submitter.name);
-	// resolveSubmitterFallback below promotes this to msg.Action only when
+	// resolveSubmitterFallback (called immediately after) promotes this to msg.Action only when
 	// msg.Action is empty, so lvt-action always wins. Read here (before the
 	// jsonDataParsed branch) so submitter routing applies whether or not a
 	// JSON "data" envelope is present.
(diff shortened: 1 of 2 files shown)


adnaan:
Round-2 reply (claude[bot] review):

**Substantive item — added.** Pushed test for the proposal Out of Scope #6 case (form `lvt-submitter` + `submitter` key inside JSON data blob). Form-level routing wins; JSON-data `submitter` stays in `msg.Data` as user data.

**Comment-duplication nit — declining.** The proposal explicitly requests symmetric "Action resolution order" doc comments in both `parseURLEncodedForm` and `parseMultipartForm` (Phase 1 bullet: *"Add an 'Action resolution order' doc comment to parseMultipartForm mirroring the one already in parseURLEncodedForm. Both functions will share the same 4-step order after Phase 1; documenting it in only one place is asymmetric."*). The trailing paragraph in `parseMultipartForm` is multipart-specific (explains the pre-`jsonDataParsed` control-flow position) and is not duplicated.

Project guidance for doc comments: **explicit clarity over DRY** when the duplication is the documentation itself. CLAUDE.md "AI Code Review Workflow" specifically calls out this kind of conflict.

PR #397, opened 2026-05-09
refactor(session): typed flashKey to prevent key-space confusion (#347)

### LINKED ISSUE #347: FOLLOW-UP: Harden flashExpiry key-space asymmetry against future misuse (from PR #344)
## Context

This follow-up task was identified during the review of PR #344.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/344
**PR Title:** feat: __navigate__ action + flash persist-until-cleared lifecycle
**Suggested by:** @claude-bot (early review rounds)

## Task Description

`connState` has a subtle key-space asymmetry:

- `messages` stores flash keys with the `_flash:` prefix (e.g., `"_flash:info"`)
- `flashExpiry` stores the same flash keys as **bare keys** (e.g., `"info"`)

This is correct and documented in the struct comment, but it's a maintenance trap. A future contributor adding a new flash-related code path who follows the `messages` key-naming convention by analogy would write:

```go
// WRONG — silently breaks expiry:
c.flashExpiry[lvtcontext.FlashPrefix+key] = time.Now().Add(d)
```

This would compile, pass all tests (the incorrect key is never looked up), and cause flash messages to never expire — with no error and no observable failure until a user notices their "5-second" notification persisting indefinitely.

### Possible Mitigations

1. **Type alias** for the bare key to make the distinction explicit at the type level:
   ```go
   type flashKey string // bare key, no _flash: prefix
   flashExpiry map[flashKey]time.Time
   ```

2. **Helper constructors** that centralize the key translation, reducing copy-paste surface:
   ```go
   func flashBareKey(key string) string { return key } // identity, but named
   ```

3. **Expanded comment** at every flashExpiry access site noting "bare key, not prefixed".

Option 1 is the most defensive — the type system prevents the mix-up. Option 3 is the lowest-effort interim fix.

## Original Comment

> `flashExpiry` stores bare keys (e.g., `"info"`) while `messages` stores prefixed keys (`"_flash:info"`). This is documented in the struct comment but the asymmetry is a maintenance trap. A future code path that writes `c.flashExpiry[lvtcontext.FlashPrefix+key]` (following the `messages` pattern by analogy) will silently break pruning and clearing for that key — no error, just flash that never expires.

[View original PR](https://github.com/livetemplate/livetemplate/pull/344)

---

adnaan:
## Summary

Closes #347. Defensive **defined type** to prevent silent breakage of the deliberate key-space asymmetry between `connState.messages` and `connState.flashExpiry`.

**No behavior change. No public API change.** Internal-only hardening.

## The trap this prevents

`connState` has two maps that store flash data:

- `messages` — keys prefixed with `_flash:` (e.g. `"_flash:info"`)
- `flashExpiry` — bare keys, no prefix (e.g. `"info"`)

This is correct, and documented in the struct comment. But a future contributor following `messages`'s naming convention by analogy could write:

```go
// WRONG — silently breaks expiry:
c.flashExpiry[lvtcontext.FlashPrefix+key] = time.Now().Add(d)
```

This compiles, passes all tests (the wrong key is never looked up), and causes flash messages to never expire — no error, just user-visible "my 5-second notification persists indefinitely."

## The fix (Option 1 from the issue)

```go
type flashKey string

flashExpiry map[flashKey]time.Time
```

Note: `type flashKey string` is a **defined type**, not a type alias. A true alias (`type flashKey = string`) would be structurally identical to `string` and would *not* prevent this misuse — the compiler would still accept `lvtcontext.FlashPrefix+key`. The compile-time enforcement works precisely because `flashKey` is a defined type.

Every access site converts explicitly: `flashKey(rawBareKey)`. The cast is the point — it's the moment a reader sees "we're deliberately keeping the bare key here, not the prefixed one." A future write that uses `lvtcontext.FlashPrefix+key` fails to compile against `map[flashKey]time.Time`.

## Why Option 1 over Options 2/3

The issue lists three mitigations. Option 1 (defined type) is the most defensive: the type system *enforces* the rule rather than relying on a comment that could rot. Options 2 and 3 (helper constructors / expanded comments) leave the same silent-failure path open.

## Test plan

- [x] All flash tests pass (`go test -run Flash ./...`)
- [x] Full test suite passes via pre-commit (300s timeout)
- [x] Linters pass: errcheck, govet, ineffassign, staticcheck, unparam, unused
- [x] No behavior diff in flash lifecycle tests — only the test setup line that constructs the map directly was updated

### DIFF
--- a/flash_lifecycle_test.go
+++ b/flash_lifecycle_test.go
@@ -16,7 +16,7 @@ import (
 func newFlashState() *connState {
 	return &connState{
 		messages:    make(map[string]string),
-		flashExpiry: make(map[string]time.Time),
+		flashExpiry: make(map[flashKey]time.Time),
 	}
 }
 
@@ -36,7 +36,7 @@ func TestExpiredFlashAbsentFromGetMessagesSnapshot(t *testing.T) {
 		t.Fatal("flash not set")
 	}
 	// Backdate the deadline to simulate expiry having elapsed since SetFlash.
-	cs.flashExpiry["success"] = time.Now().Add(-1 * time.Second)
+	cs.flashExpiry[flashKey("success")] = time.Now().Add(-1 * time.Second)
 
 	// getMessages must NOT include the expired entry — it prunes before
 	// snapshotting so the next render walks a clean slate.
@@ -97,7 +97,7 @@ func TestFlashExpiryPrunesAfterDeadline(t *testing.T) {
 	// Backdate the expiry to force it to be past-due. Direct map write is safe
 	// here: these are sequential single-goroutine tests; setFlash initialized
 	// the map above so the nil-map panic path is excluded.
-	cs.flashExpiry["info"] = time.Now().Add(-1 * time.Second)
+	cs.flashExpiry[flashKey("info")] = time.Now().Add(-1 * time.Second)
 
 	cs.pruneExpiredFlash()
 
@@ -114,7 +114,7 @@ func TestFlashExpiryDoesNotAffectNonExpiredMessages(t *testing.T) {
 	cs.setFlash("info", "Transient...", time.Minute) // expires in 1 minute (not yet)
 
 	// Manually backdate only the "info" key to be past-due (sequential test, safe).
-	cs.flashExpiry["info"] = time.Now().Add(-1 * time.Second)
+	cs.flashExpiry[flashKey("info")] = time.Now().Add(-1 * time.Second)
 
 	cs.pruneExpiredFlash()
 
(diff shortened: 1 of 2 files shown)


adnaan:
Round-2 reply (claude[bot] review):

Good catch on the **defined type vs type alias** terminology. You're right — `type flashKey string` is a defined type, and a true alias (`type flashKey = string`) wouldn't have the property this PR depends on. I've updated the PR description to say "defined type" throughout.

The code itself doesn't change. Squash-merge commit message will pick up the corrected description.

Issue #386, opened 2026-05-08
Tracking: PR #213 Redis pubsub correctness/test follow-ups


adnaan:
Tracking issue to bundle the three unaddressed follow-ups from [PR #213](https://github.com/livetemplate/livetemplate/pull/213) (PubSub broadcaster). All three still apply against current `pubsub/redis.go`; #215 in particular is a real availability concern.

## Sub-issues

| # | Priority | Area | Status |
| --- | --- | --- | --- |
| #215 | **P1** | bug | Release mutex before Redis network calls in `subscribeTo`/`reconnect` — confirmed: `b.mu.Lock()` is held across `pubsub.Subscribe()` Redis call (`pubsub/redis.go:390-409`). Head-of-line blocking under network slowness/partition. |
| #214 | P2 | correctness | Add reference counting for pubsub channel unsubscribe — `subscribedChannels` is `map[string]struct{}` (`pubsub/redis.go:53`), no per-subscriber refcount → channels never unsubscribed in long-running servers. Memory-leak risk. |
| #216 | P3 | testing | Replace `time.Sleep` test synchronization with polling in `pubsub/redis_test.go` (19+ sleep points across lines 135–867). Flake risk under load. |

## Why bundle these

- All three touch the same file (`pubsub/redis.go` + its test) and the lock-discipline change in #215 will likely interact with the refcount change in #214.
- Doing them as one PR avoids redoing locking analysis twice.
- The test sleep migration (#216) is a natural cleanup pass once #215 lands and races become visible/diagnosable.

## Recommended sequencing

1. **#215** first — fix the mutex-across-Redis-call bug; this is the only one that affects production.
2. **#214** — add refcount with the new locking discipline in mind.
3. **#216** — migrate sleeps to `require.Eventually` polling once locking is stable.

Close this tracker when all three are closed.

adnaan:
All three sub-issues are now closed:

- #214 — pubsub channel reference counting (closed via #394)
- #215 — release mutex before Redis network calls (closed)
- #216 — replace time.Sleep with polling in pubsub tests (closed via #398, merge SHA 9abe2fe)

This tracking epic can be closed.

adnaan:
All three sub-issues are now closed:

- #215 — release mutex before Redis network calls (closed via #391)
- #214 — refcount channel subscriptions (closed via #394)  
- #216 — replace time.Sleep with polling in pubsub tests (closed via #398)

[...]

Issue #214, opened 2026-03-19
FOLLOW-UP: Add reference counting for pubsub channel unsubscribe (from PR #213)


adnaan:
## Context

This follow-up task was identified during the review of PR #213.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/213
**PR Title:** fix: track dynamic pubsub subscriptions for reconnect and wire into mount
**Suggested by:** @claude[bot], @copilot

## Task Description

`subscribedChannels` in `RedisBroadcaster` grows unboundedly — channels are added when connections join groups/users but never removed when connections disconnect. Over the lifetime of a long-running server with high session churn, this accumulates stale subscriptions in both the in-process map and on the Redis server.

**Current behavior:** No unsubscribe. Stale subscriptions are harmless (GetByGroup returns empty slice → no-op fan-out) but waste memory and Redis resources.

**Proposed:** Add a reference-counted `map[string]int` that tracks how many local connections use each channel. When the count reaches zero on disconnect, call `pubsub.Unsubscribe()` and remove the entry.

**Considerations:**
- Need to hook into `registry.Unregister()` or the WebSocket disconnect path in `mount.go`
- Must be thread-safe (unsubscribe under mutex)
- The `reconnect()` replay should still work — it replays whatever is in the map at reconnect time

---
*This issue was automatically created from PR review comments.*

PR #398, opened 2026-05-09
test(pubsub): migrate time.Sleep to require.Eventually polling (#216)

### LINKED ISSUE #216: FOLLOW-UP: Replace time.Sleep test synchronization with polling in pubsub tests (from PR #213)
## Context

This follow-up task was identified during the review of PR #213.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/213
**PR Title:** fix: track dynamic pubsub subscriptions for reconnect and wire into mount

## Task Description

Pubsub tests use `time.Sleep(100-500ms)` to wait for goroutines to start, subscriptions to register, reconnects to complete, and messages to arrive. This is fragile under CI load where timing varies.

**Affected tests:** Most tests in `pubsub/redis_test.go`, particularly:
- `TestRedisBroadcaster_ReconnectPreservesDynamicSubscriptions` (200ms reconnect wait + 500ms message wait)
- All tests using `time.Sleep(100ms)` to wait for `Subscribe()` goroutine startup

**Proposed:** Replace sleeps with:
- Polling loops with short deadlines (e.g., `require.Eventually` pattern)
- `pubsub.Receive()` confirmation before proceeding (already used in `Subscribe()` itself)
- `WithReconnectDelay(0)` in reconnect tests to remove timing variable

---

adnaan:
## Summary

Replaces 23 of 30 `time.Sleep` test-synchronization sites in `pubsub/redis_test.go` with deterministic alternatives, dropping suite runtime ~12% (27.6s → 24.3s) and removing CI flake risk.

This is the last sub-issue of tracking epic #386 — **closes #216 and resolves the epic** (#214 and #215 are already closed).

## Replacement patterns

**1. Goroutine-wrapped Subscribe + 100ms sleep → synchronous Subscribe.** `Subscribe()` already blocks on `pubsub.Receive()` until Redis ACKs the SUBSCRIBE (see `pubsub/redis.go` line 234), so the goroutine + sleep was redundant. Synchronous Subscribe also lets us use `t.Fatalf` (correct from main goroutine) instead of `t.Errorf` from a child goroutine. Applies to ~15 sites.

**2. Wait-for-state-change → `waitFor()` polling helper.** Used for:
- *Reconnect completion*: poll until `b.pubsub` pointer has been swapped to a new connection AND `reconnecting` flag clears. Capturing the old pubsub pointer before forcing the close avoids the trap that `b.pubsub != nil` is true even with a dead connection.
- *Message arrival after publish*: poll until the WaitGroup/counter reaches the expected value.

The helper polls every 10ms up to a 2s deadline (matching the test budget).

**3. Wait-for-message-after-publish that has a deterministic sibling → poll on the sibling's arrival.** `TestRedisBroadcaster_UnsubscribedGroupNotReceived` uses group-A's arrival as the synchronization point that proves group-B would also have arrived if not filtered. Eliminates the 500ms blanket sleep.

## Sleeps deliberately retained

- **5x `subscribeAckWindow` (50ms each)**: `subscribeTo()` writes SUBSCRIBE bytes but, unlike top-level `Subscribe()`, does **not** call `Receive()` to wait for Redis to ACK. Tests that publish on the new channel immediately can race the SUBSCRIBE round-trip on a separate connection and silently drop the message. The retained sleeps are gated by an exported constant with explanatory comment. **Worth a follow-up to consider adding `Receive()` to `subscribeTo()` for symmetry with `Subscribe()`** — would let us eliminate these too. (Out of scope for this PR; library behavior change should be its own issue.)

- **1x LocalOptimization (200ms)**: negative assertion ("message must NOT arrive within N ms"). A bounded sleep is the honest way — there is no state change to poll for.

- **1x inside the `waitFor` helper itself (10ms polling interval)**.

## Test plan

- [x] `GOWORK=off go test -race -count=20 ./pubsub/...` — passed across ~490s of runtime, no flakes
- [x] `GOWORK=off go test ./...` — full repo suite passed
- [x] `golangci-lint run --enable-only=errcheck,govet,ineffassign,staticcheck,unparam,unused` — 0 issues
- [x] Pre-commit hook passed (formatting, lint, full test suite with 300s timeout)

## Sleep accounting

| | Before | After |
|---|---|---|
| `time.Sleep` call sites | 30 | 7 (1 helper interval + 1 negative assert + 5 ACK gates) |
| Net removed | — | 23 |
| Suite runtime (single run) | 27.6s | 24.3s |

### DIFF
--- a/pubsub/redis_test.go
+++ b/pubsub/redis_test.go
@@ -17,6 +17,37 @@ import (
 // them as always-nil-result).
 var errTestHandlerNilMsg = errors.New("test handler received nil message")
 
+// waitForTimeout is the default deadline for waitFor. 2s is generous enough
+// for slow CI runners while still failing fast on a real bug; all callers
+// today share the same budget so it lives in one place.
+const waitForTimeout = 2 * time.Second
+
+// subscribeAckWindow is the brief delay after SubscribeToGroup/SubscribeToUser/
+// SubscribeToServerAction/SubscribeToGroupAction returns. Unlike top-level
+// Subscribe() (which calls pubsub.Receive() to wait for Redis to ACK the
+// SUBSCRIBE), subscribeTo() only writes the SUBSCRIBE bytes — Redis processes
+// them async on its end. A test that publishes on the new channel without
+// this gap can race the SUBSCRIBE round-trip on a separate connection and
+// silently drop the message. 50ms is well above Redis's sub-ms processing
+// time and matches the original test cadence. (Worth a follow-up to consider
+// adding Receive() to subscribeTo for symmetry with Subscribe.)
+const subscribeAckWindow = 50 * time.Millisecond
+
+// waitFor polls cond at 10ms intervals up to waitForTimeout. Returns true if
+// cond became true within the deadline, false otherwise. Used to replace
+// time.Sleep-based test synchronization where we're waiting for an
+// observable state change (e.g. reconnect completion, message arrival).
+func waitFor(cond func() bool) bool {
+	deadline := time.Now().Add(waitForTimeout)
+	for time.Now().Before(deadline) {
+		if cond() {
+			return true
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	return cond()
+}
+
 // getTestRedisClient returns a Redis client for testing using testcontainers.
 func getTestRedisClient(t *testing.T) redis.UniversalClient {
 	return testutil.GetTestRedisClient(t)
@@ -124,15 +155,11 @@ func TestRedisBroadcaster_SubscribeAndReceive(t *testing.T) {
 		return nil
 	}
 
-	// Start subscriber
-	go func() {
-		if err := broadcaster2.Subscribe(handler); err != nil {
-			t.Errorf("Subscribe failed: %v", err)
-		}
-	}()
-
-	// Give subscriber time to start
-	time.Sleep(100 * time.Millisecond)
+	// Subscribe synchronously — Subscribe() blocks until the Redis SUBSCRIBE
+	// is confirmed, so no sleep is needed afterward.
+	if err := broadcaster2.Subscribe(handler); err != nil {
+		t.Fatalf("Subscribe failed: %v", err)
+	}
 
 	// Publish from broadcaster1
 	payload := []byte(`{"test": "data"}`)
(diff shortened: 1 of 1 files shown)


adnaan:
Thanks for the review.

- **Title vs `require.Eventually`**: the orchestrator-issued title in #216 used `require.Eventually` as shorthand for "polling instead of fixed sleep". Adding `testify` as a direct dep just for this one helper would create more churn than it saves; the project has no other testify usages. I went with the bespoke `waitFor()` for that reason. Leaving the title as-is to keep traceability with the issue, but acknowledged here.
- **`waitFor` comment trim**: applied in 56aa502a — kept the WHY, dropped the WHAT.
- **`waitForTimeout` "generous enough" phrasing**: leaving as-is per "your call" — the rationale clarifies why 2s is in the budget rather than something tighter.

adnaan:
Addressed all 4 Copilot inline comments in 218015d5:

1. **`subscribeAckWindow` doc** (line 33): reworded to remove the "matches original cadence" claim — the originals were 100ms, this trims to 50ms with the margin documented.
2. **SubscribeToUser comment** (line 386 → now around 384): replaced the misleading "blocks until ACK" claim with an accurate "writes SUBSCRIBE bytes synchronously; the ACK is gated by the subscribeAckWindow sleep below."
3. **Hard-coded "2s" in failure messages** (lines 636, 895): both now use `t.Fatalf(... %v, waitForTimeout)` so they stay accurate if the timeout is tuned.

adnaan:
Thanks for the round-3 review.

The verdict says "ready to merge" with the only remaining note being comment verbosity on `waitForTimeout` and `subscribeAckWindow`. Per CLAUDE.md "AI Code Review Workflow" convergence rule (round 3 is purely cosmetic style on already-accepted content), I'm declining the further trim and merging.

Rationale for the longer forms staying:
- `waitForTimeout`: the "all callers today share the same budget" line documents the intentional shape of the constant (single shared budget, not a tuning knob) — that's a design statement, not redundant WHAT.
- `subscribeAckWindow`: explicitly listing the four `SubscribeTo*` functions makes this discoverable to readers landing here from any of them via grep. The asymmetry-with-Subscribe rationale is the load-bearing comment; the function list is cheap discoverability.

Both are 1-2 lines longer than the bot's compressed forms but neither is incorrect or stale-prone. Calling this converged.

PR #401, opened 2026-05-10
docs(proposals): collapse explicit-submitter to two-phase rollout (#237)

adnaan:
## Summary

LiveTemplate is alpha with no external users, so the original four-phase migration in `docs/proposals/explicit-submitter.md` solves a back-compat problem that doesn't exist. Collapses to a two-phase rollout matching what's actually shipping:

- **Phase 1** — server accepts `submitter` (JSON) and `lvt-submitter` (form key) — already shipped in #396.
- **Phase 2** — client emits both fields — active follow-up in `livetemplate/client`.
- **No-JS fallback** — empty-value heuristic stays permanently (apps under strict CSP / no-script can't inject `lvt-submitter` from JS at all).

## What changes in the proposal

- Status: `Proposed` → `Accepted (Phase 1 shipped in #396; Phase 2 client work is the active follow-up)`. Adds a "Why two phases instead of four?" paragraph at the start of the Implementation Plan section recording the alpha/no-users posture (revised in round 2 from a header-level `Stance:` field, then `Rationale:` field, both flagged as non-standard by claude-bot).
- **Cuts entirely:** Phase 3 (deprecation log), Phase 4 (heuristic removal target v0.9.0), Risks Q3 (no-JS support sub-options + Phase 4 ADR gating), Risks Q4 (deprecation-log suppression mechanism).
- **Tightens Phase 2 spec:** concrete file/function targets in the client repo (`dom/event-delegation.ts` WS message envelope insertion, Tier 1 multipart `tier1FormData.set("lvt-submitter", …)`, the `lvt-form:emit-submitter` directive at the wrapper-delegated `submit` listener). Round-2 review verified that the client never produces URL-encoded request bodies (`sendHTTP` uses JSON, `sendHTTPMultipart` uses multipart), so that bullet is a definitive no-op rather than a "verify if" conditional.
- **Adds No-JS forms subsection** explicitly stating the heuristic is the permanent fallback and listing the four failure modes that audience inherits.
- **Phase 1 reframed** as "Complete (#396)" with a one-paragraph summary plus a small Implementation reference subsection containing the `resolveSubmitterFallback` snippet.
- **References use grep anchors** instead of line numbers, per project guideline that line numbers silently drift.

## Cross-reference touch-up

`docs/proposals/lvt-hook-scope-and-elevation.md` had four references to the old structure of the explicit-submitter proposal — three citing "the form-submit heuristic" as part of the v0.9.0 breaking-change cut, and one citing the "no-JS support ADR pattern" that this revision deletes. All four updated to remove the stale references; the form-submit heuristic is no longer scheduled for v0.9.0 removal.

## What this PR does NOT do

The Phase 2 client implementation (`dom/event-delegation.ts` edits, `lvt-form:emit-submitter` directive, tests, version bump) is **not** in this PR. It's a separate plan in the `livetemplate/client` repo, with this revised proposal as its source of truth. The cross-repo handoff is pinned by a comment on the tracking issue (#237) referencing Verification item 3 of the revised proposal.

## Test plan

- [x] `grep -cE "^### Phase [34]" docs/proposals/explicit-submitter.md` returns `0` (no Phase 3 or Phase 4 section headings remain; "Phase 3" / "Phase 4" still appear once in the historical-context rationale paragraph, which is intentional)
- [x] `grep -cE "LVT_FORM_SUBMITTER_COMPAT|WithFormSubmitterMode|adr-no-js" docs/proposals/explicit-submitter.md` returns `0`
- [x] No remaining "explicit-submitter Phase 4" references in non-archived docs
- [x] `applyDefaultAction("submit")` syntax error fixed (round-3) — the function takes `*message`, not a string parameter
- [x] Pre-commit hook passes (lint + full test suite, including pubsub Docker tests)
- [x] All four review rounds addressed: round 1 (3 Copilot inconsistencies + 2 claude-bot style), round 2 (URL-encoded bullet definitive answer + Rationale field folded + e2e tracking pointer + double-click race wording), round 3 (`applyDefaultAction` signature fix + handoff comment on #237), round 4 (this update — tightened Test-plan grep patterns)
- [ ] Reviewer reads top-to-bottom for internal consistency between Status block, Implementation Plan, and Verification sections
- [ ] PR preview renders cleanly on GitHub


adnaan:
Round-1 review responses:

**Accepted (pushed in 25c4d7fb):**
- Copilot lines 83 / 89 / 108 — all three flagged the same conflation in different spots: "no-JS" was being used loosely to mean three different things (no `<script>` bundle / no JS at submit-time / no inline `onclick`). Reworded to distinguish the three audiences: fetch-submitted forms, JS-enabled native submits with the directive, and pure no-JS forms.
- claude-bot suggestion 1 — `Stance:` → `Rationale:` (more self-explanatory; Stance isn't used elsewhere in `docs/proposals/`).
- claude-bot suggestion 4 — "silently drop `onclick` handlers" → "the browser silently ignores `onclick` handlers" (more accurate for a reader debugging CSP).

**Declined with reasoning:**
- claude-bot suggestion 2 (Phase 2 milestone marker) — Phase 2's status is already trackable via the issue link at the top of the proposal (`Tracking issue: #237 (P1)`). No client-repo issue exists yet, and inventing a target version (v0.8.x / v0.9.x) without a real release plan would be drift bait — the proposal would claim a milestone that nobody is actually committed to, and it would silently rot once the real release sequence diverges. Project is alpha; the issue link is the right abstraction for status.
- claude-bot suggestion 3 (Q5 / file-upload ergonomics one-liner) — adding "Verified in #396 examples smoke test" would be a documentation claim I can't substantiate; there's no smoke-test artifact from #396 I can cite. Per the project's "factual claims in docs get bot-verified" rule, I'd rather drop the assertion than ship one I can't back up. The current Q5 text already addresses the substantive concern (zero read overhead for `r.FormValue("lvt-submitter")` on the multipart path); the ergonomic question gets re-raised for real when Phase 2's client implementation lands and exercises file-upload flows end-to-end.

adnaan:
Round-2 review responses:

**All four points addressed (pushed in 78a07670):**

1. **"Verify if" in Phase 2 URL-encoded bullet** — went and verified in the client repo. \`sendHTTP()\` (livetemplate-client.ts) sends \`application/json\`; \`sendHTTPMultipart()\` sends \`multipart/form-data\`. The client never produces URL-encoded bodies. The server's URL-encoded parser is exercised only by native browser POSTs, which the \`lvt-form:emit-submitter\` directive covers by injecting the hidden input pre-serialization. The bullet now says this definitively — no more "verify if."

2. **\`Rationale:\` field** — dropped the standalone header field; folded the alpha/no-users context into a short "Why two phases?" paragraph at the start of Implementation Plan. Resolves the "non-standard field" concern from this round AND from the previous round (which originally suggested \`Rationale\` as the rename for \`Stance\`); the underlying observation both rounds was "the field doesn't match other proposals," and integrating the content into the body section addresses that root cause rather than picking another novel field name.

3. **Verification item 3 e2e test deferral** — added "Tracked under [#237] until the Phase 2 PR opens; that PR will cite this verification item explicitly so the cross-repo handoff doesn't drop the requirement." Closes the loop without fabricating an issue number.

4. **Double-click race wording** — fixed. The old text asserted "both submits carry the same \`submitter\`," which is only true for same-button double-clicks; cross-button rapid races would carry different values. Reworded to "each request carries whatever \`submitter\` its own click resolved (the same name for a same-button double-click, different names for a cross-button race). No mutual-exclusion guarantee is needed or provided."

This round's feedback was substantive (the URL-encoded bullet was a real contradiction with the PR description). Subsequent rounds focused on cosmetic alternatives or convergence checks should be addressed by reply rather than push, per the project's bot-review-loop convergence rule.

adnaan:
Round-3 review responses:

**Verdict received: Approve.** Two minor fixable items addressed:

1. **`applyDefaultAction("submit")` signature mismatch** — verified against \`action.go:52\` (\`func applyDefaultAction(msg *message)\`). The string isn't a parameter; the function defaults \`msg.Action\` to \`"submit"\` internally. Corrected to "falls through to \`applyDefaultAction\`, which defaults the action to \`"submit"\`" — pushed in $(git rev-parse --short HEAD).

2. **e2e handoff strengthening** — dropped a tracking comment on the underlying issue (#237) noting that the Phase 2 PR must include the e2e test from Verification item 3 of this proposal: livetemplate/livetemplate#237 (comment). The proposal text already references #237; this comment closes the loop on the GitHub side so a future Phase 2 author finds the requirement when they pick up the issue.

**Acknowledged but not changed:**

- The "removed implementation detail about \`key == "action"\` location" observation. The grep anchor for \`detectSubmitButtonName\` covers the function, and someone modifying it will see the in-helper exclusion when they read the source. The diagnostic note added length without commensurate value once Phase 4's removal context (which made the note relevant) was stripped.

This was a substantive round (signature fix is a real correctness improvement for the doc). If the next round only surfaces style/wording alternatives, I'll address by reply per the bot-review-loop convergence rule rather than push.

adnaan:
Round-4 review responses:

**Caught a real regression — addressed:**

The two grep checks I claimed pass in the original PR description (`Phase [34]` returns 0; `deprecat|...` returns 0) actually returned 1 each after round 2's "Why two phases instead of four?" rationale paragraph was added. That paragraph legitimately mentions "Phase 3 rate-limited compat warning, Phase 4 ADR-gated heuristic removal" as historical context for the simplification, so the prose is right but the test-plan greps were too loose to distinguish that prose from leftover scaffolding.

Took option 1 (tighten the patterns) per claude-bot's preference and mine. The proposal text stays unchanged; the PR description now uses:
- \`grep -cE "^### Phase [34]" docs/proposals/explicit-submitter.md\` — matches Phase 3/4 *section headings* only, not prose
- \`grep -cE "LVT_FORM_SUBMITTER_COMPAT|WithFormSubmitterMode|adr-no-js"\` — drops the loose \`deprecat\` term, since "deprecation" appears legitimately in the rationale paragraph now

Both return 0. PR description updated. The Test plan checkboxes are now honestly marked.

**Acknowledged but not changed:**

- \`Stance\` label observation — the round-2 "fold into TL;DR / put before Implementation Plan" was the right call; the content is now in a \`**Why two phases instead of four?**\` bold-lead paragraph at the top of the Implementation Plan section (where the bot suggested it could go), and there's no separate \`Stance:\`/\`Rationale:\` header field anymore. Consistent with what other proposals in \`docs/proposals/\` do.

**Convergence call:**

This round's substantive issue (the failing grep checks) is now resolved. Across four rounds the bot has surfaced: 3 Copilot inconsistencies (round 1), 4 substantive items (round 2), 1 signature fix + 1 handoff (round 3), 1 verification regression (round 4). Each round caught something real. If round 5 only surfaces style/wording alternatives on settled content, I'll address by reply rather than push, per the project's bot-review-loop convergence rule.

PR #411, opened 2026-05-14
docs: propose BroadcastAction redesign with implicit peer sync and topics

adnaan:
## Summary

Adds a design proposal at `docs/proposals/broadcast-action-redesign-proposal.md` that addresses three problems with today's `ctx.BroadcastAction`:

- **Boilerplate at every mutation site.** The dominant call pattern (visible in `e2e/docker/app/main.go`, `broadcast_test.go`) is `mutate → BroadcastAction("RefreshFoo", nil) → controller method that just re-reads state`. The action call exists only to fan-out a re-render.
- **No same-user, cross-device default.** Whether a broadcast reaches a user's other devices depends entirely on how `Authenticator.GetSessionGroup` happens to map `userID → groupID`. Silently broken for custom per-device groupIDs.
- **No cross-user topic broadcast in the public API.** PubSub already implements `PublishToUser` / `PublishGlobal` / `PublishGroupAction`, but `Context` exposes only group-scoped broadcasts. Use cases like chat rooms or auction bids require coercing the Authenticator into a contrived groupID.

The proposal separates three concerns currently entangled in `BroadcastAction`:

1. **State scope** — `Authenticator.GetSessionGroup`, unchanged.
2. **Auto-sync scope** — new implicit peer sync after any successful action. Follows `userID` when present (use case A), `groupID` when anonymous (use case B). `ctx.SkipPeerSync()` opts out per-action. Render-only path on peers — no controller invocation, so no double-counted side effects.
3. **Topic fan-out** — new explicit `ctx.SubscribeTopic` / `ctx.BroadcastToTopic` with a single `WithTopicACL` hook for auth gating. Also exposes out-of-band `handler.BroadcastToUser` / `BroadcastToTopic` / `BroadcastGlobal` for webhooks and cron.

`ctx.BroadcastAction` stays in the API for the rare case where a peer needs to react to a sender-chosen payload, not just re-render with the new state.

The proposal enumerates 11 use cases (A–K) with today's support status, points to existing primitives that can be reused (`registry.GetByUserExcept`, `Connection.EnqueueDispatch`, the pubsub channel patterns), lists the files that would change, and includes a 12-test verification plan mirroring `broadcast_test.go`.

**Hard pre-implementation gate (per round-3 review):** `lvt` and `client` repos must be audited against this design before the implementation PR opens — `lvt`'s e2e app uses exactly the `BroadcastAction` pattern the proposal recommends deleting.

## Test plan

- [ ] Reviewer reads `docs/proposals/broadcast-action-redesign-proposal.md` end-to-end and validates the use-case matrix is exhaustive
- [ ] Confirm the three-concern split (state / sync / topics) lands the way it reads
- [ ] Spot-check the grep anchors in the "Current architecture" and "Existing utilities to reuse" sections
- [ ] Audit `lvt` and `client` against the proposal (specifically: `BroadcastAction` call sites the proposal marks redundant; any usage shapes outside use cases A–K)
- [ ] Confirm benchmark requirements (§ "Benchmarks required before implementation finalizes") are acceptable as the gating criteria for the implementation PR

https://claude.ai/code/session_01TNnXaGTX9aY5QLa62dEzki


adnaan:
Thanks for the review. Addressed in ce0c224:

**Factual fixes**
- #1 `GetByUserExcept`: corrected. The function doesn't exist today; the doc now says it's a small new function to add, modeled after `GetByGroupExcept` (line 468). Both the "Current architecture" and "Critical files to modify" sections updated.
- #2 `pubsub/redis.go:20-31`: corrected. Reference now points to the interface at `pubsub/types.go:54-95` plus the actual implementation range `redis.go:104-323`, with a grep anchor (`^func (b *RedisBroadcaster) Publish`) for drift resilience.

**Design clarifications added**
- #3 Cross-instance state-consistency: new constraint paragraph requiring `persistState` to commit before `RenderInvalidation` is published. Two options discussed; option (a) — synchronous persist-then-publish — recommended for consistency with current `processBroadcasts` ordering.
- #4 Anonymous fallback: clarified explicitly. The proposal now states "Else (anonymous, `userID == ""`)" uses `GetByGroupExcept` and explicitly **not** `GetByUser("")`, with the reason (cross-browser state leak) called out.
- #5 Breaking-default behavior: new "Behavior-change warning" paragraph in the Migration section. Documents both opt-out paths (`ctx.SkipPeerSync()`, `WithImplicitSyncDisabled()`) and flags the inversion (default-off via `WithImplicitSyncEnabled()`) as an open decision for the reviewer. The user-facing decision to default-on was explicit in pre-design Q&A, but the migration impact was understated; that's now front and center.
- #6 Topic ACL timing: new explicit paragraph stating the ACL is evaluated at subscribe time, the trade-off (cheap fan-out vs. revocation latency), and the workaround (server-side `UnsubscribeTopic` when permissions change).
- #7 Double-update edge case: new "deduplication semantics" paragraph. Stance: the framework does not dedupe; the two paths serve different purposes (state diff vs. controller invocation with payload). `ctx.SkipPeerSync()` is the escape hatch for developers who want strictly one update.

**Minor notes**
- Render fan-out coalescing: promoted from "Open Questions" to a v1 design constraint with a starting implementation sketch (10ms debounce timer, reset on `EnqueueRender`).
- Topic GC: also promoted to a constraint; explicitly wired into the existing `Connection.Close()` cleanup path (same code site as `byUser` / `byGroup` cleanup) rather than a separate goroutine.
- `mount.go:1626-1631` line ref: replaced with grep anchor `BroadcastAction calls inside a dispatched action are ignored` per CLAUDE.md guidance on grep-anchors-over-line-numbers.

The bot's correction on `GetByUserExcept` was the most important — it was a real factual error that would have misled implementers. Good catch.

---
_Generated by [Claude Code](https://claude.ai/code)_

adnaan:
Round-2 review fixes pushed in `34ec6d5`. The inline-reply API errored on pending-review constraints, so consolidating responses here.

### Maintainer inline comments

**[Line 17 — "Need snippets in the design"](https://github.com/livetemplate/livetemplate/pull/411#discussion_r3243700423):** added an "At a glance" snippet block right below the Context section. Four code blocks for the same `Send` action: today's verbose `BroadcastAction + RefreshMessages` pair, proposed implicit sync (zero broadcast code), proposed topic broadcast with Mount-time subscription, and the new handler-level entry points. The intent is to make the three-concern split tangible before the reader hits the use-case matrix.

**[Line 200 — "Need benchmarks"](https://github.com/livetemplate/livetemplate/pull/411#discussion_r3243705523):** new "Benchmarks required before implementation finalizes" section. Five measurements with an explicit acceptance criterion (< 1ms post-action hook overhead at 100 connections/user under 10 actions/sec). The 10ms / 50ms coalescer defaults are now flagged as placeholders pending the benchmark numbers, not final values.

### Bot review feedback (rounds 2 + 3)

| Finding | Resolution |
|---|---|
| Line numbers drift (CLAUDE.md grep-anchor guidance) | Converted refs throughout Current Architecture, Critical Files, and Existing Utilities to grep anchors. |
| `pubsub/types.go:54-95` misses `DynamicSubscriber` | Replaced with three grep anchors per interface (`Broadcaster`, `GroupActionBroadcaster`, `DynamicSubscriber`). |
| `RenderChan` vs discriminated dispatch left open | Decided: discriminated `DispatchRequest{Kind: KindRender}` on the existing channel. Rationale documented (single select, simpler coalescer). |
| Local vs remote state source unclear | Now explicit: local peers read in-memory from the originator's post-mutation state pointer (no store round-trip); remote peers re-read from the store after `RenderInvalidation`, gated by the write-ordering constraint. |
| Implicit sync recursion guard | New explicit statement: implicit sync is suppressed inside `handleDispatchedAction`. Prevents N-deep cascade where each topic-broadcast recipient triggers its own peer sync. |
| `SubscribeTopic` on HTTP GET | Documented as a no-op on HTTP GET — the call records intent on the Context but only attaches to a Connection on WS connect. ACL still evaluated eagerly. Same `Mount` body works for both transports without `if ctx.IsInitialMount()` guards. |
| `WithTopicACL` `*http.Request` semantics | New paragraph: it's the WS upgrade request, not per-action. Implications for stateless (JWT) vs database-backed checks documented. |
| `GetByUserExcept("")` invariant | Called out as a hard precondition. Implementation note now says guard with `ctx.UserID() != ""` at the call site; the registry function should panic in dev / log+empty in prod on empty userID. |
| `BroadcastToUser` wraps `PublishServerAction` | Noted explicitly so reviewers see no new PubSub protocol is needed for use case F. |
| Double-update user-visible behavior | Added: "two consecutive WebSocket frames in rapid succession" with a note that this is benign for most UI, disruptive for focus/animation-sensitive UI. |
| Coalescing: pure debounce starves under sustained load | Updated to hybrid bounds: `IdleDelay` (10ms, resets per enqueue) **AND** `MaxDelay` (50ms, since first pending) — whichever expires first fires the dispatch. |
| `TestBroadcastAction_NoAutomaticPeerDispatch` inversion | New test #11 in verification plan: the existing test must be deleted or repurposed under `WithImplicitSyncDisabled()`. Also added tests #10 (suppress-in-dispatched) and #12 (SubscribeTopic-on-HTTP-GET). |
| "byte-for-byte" idiom | Changed to "following the same pattern as". |

### Decline: bot's reinstated "major version bump" recommendation

The bot's round-2 review (comment `4453687304`, "Breaking-default sign-off") still recommends a major version bump, a prominent changelog entry, and an upgrade-guide first section for the implicit-sync default. **Declined** per maintainer's explicit guidance in this thread that the library has no production users outside the ecosystem repos (`lvt`, `client`) at the time of writing. The proposal already documents the change as the new default with two opt-out paths (`ctx.SkipPeerSync()` per-action, `WithImplicitSyncDisabled()` per-handler) for the rare sovereign-tab pattern. No migration ceremony is warranted at this stage. Flagging the decline here so the next review round doesn't reopen it.

---
_Generated by [Claude Code](https://claude.ai/code)_

adnaan:
Round-3 review fixes pushed in `54fcc53`. The bot's round-4 review (comment `4454253361`) caught real implementation-gating gaps — most of which I'd left implicit.

| Finding | Resolution |
|---|---|
| `DispatchRequest` has no `Kind` field today; not listed in Critical Files | Added as bullet (a) in the `internal/session/registry.go` section. New `RequestKind` enum with `KindAction` as the zero value, so all existing `EnqueueDispatch` call sites stay backward-compatible. |
| Handler-level `BroadcastGlobal`/`BroadcastToTopic` wire format unspecified | Documented: reuse the existing `GroupActionMessage` envelope shape (`{"type", "action", "data", "timestamp", "instanceID"}`), marshalled to bytes before calling the byte-oriented `PublishGlobal`. No change to existing PubSub signatures. |
| `RenderInvalidation` channel pattern undecided | Settled: new dedicated `livetemplate:render:{groupID}` channel. Rationale documented (keeps the channel schema regular alongside `broadcast:`, `groupaction:`, `user:`, `topic:`; receivers can subscribe selectively without parsing envelopes). |
| ACL execution on HTTP GET is a hot path | New paragraph in §3 Topic ACL. Two mitigations: prefer stateless ACL (JWT claims) as default; database-backed ACL can return early on `r.Method == "GET"` as a deliberate trade-off that loses the pre-upgrade rejection property. |
| byTopic cleanup target: `Connection.Close()` is wrong; should be `Unregister()` | Corrected. The proposal now targets `Unregister()` explicitly with a note that it's the canonical index-cleanup site, and `Close()` delegates to it. |
| Empty-userID guard: panic vs. log left as a "pick one" | Decided: `slog.Error` + return empty slice. Rationale: matches `EnqueueDispatch`'s drop-and-log pattern; panic is hard to recover from in production if the guard fires unexpectedly. |
| Broken anchor regex `(Publish\|Subscribe)GroupAction` misses `SubscribeGroupActions` (trailing `s`) | Replaced with explicit alternation `(PublishGroupAction\|SubscribeGroupActions)`. |
| Sibling-repo re-validation should be a hard gate, not a soft note | Upgraded the caveat to "hard gate before implementation" with a concrete audit task: `lvt`'s e2e app uses the `BroadcastAction` pattern the proposal recommends deleting; confirm before any implementation PR opens. Also added to the PR's test plan. |
| PR summary count (9-test → 12-test) | PR body updated. |

### Minor items not actioned

- "`BroadcastAction calls inside a dispatched action are ignored` is a prefix match, not exact." Correct, but intentional — grep prefix matching is a feature (it finds the line even if the parenthetical drifts), so leaving as-is.
- "Use case F test (test 6) doesn't validate cross-instance." Test 7 covers cross-instance broadly including the same paths; adding a redundant cross-instance variant of test 6 doesn't add coverage commensurate with the maintenance cost. Will revisit if the implementation surfaces a cross-instance-specific bug for `BroadcastToUser` specifically.

Three points from the round-4 review are now hard pre-implementation gates that need maintainer sign-off before code lands:
1. `lvt` / `client` audit (concrete, blocking).
2. Benchmark numbers (§ "Benchmarks required") — defaults like `IdleDelay = 10ms`, `MaxDelay = 50ms` are placeholders pending measurement.
3. The three round-4 implementation decisions documented above (`Kind` field, render channel pattern, empty-userID guard strategy).

Still subscribed for next round.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #412, opened 2026-05-15
docs: enrich BroadcastAction redesign proposal with cross-repo audit

adnaan:
## Summary

Reworks the `BroadcastAction` redesign proposal (`docs/proposals/broadcast-action-redesign-proposal.md`) from a two-target design into a **single Publish/Subscribe topic model**, Phoenix-faithful. **Docs-only; no code or behavior change** (the proposal is a design doc; the implementation is a separate future PR).

This PR evolved through three phases (all in the proposal doc):

1. **Cross-repo audit** — closed #411's "sibling-repo hard gate": verified `lvt` has **zero** `BroadcastAction` call sites; `client` needs only an error-envelope guard; added the §"Impacted repositories" table + release order. Surfaced that `Sync()` was removed in #406 (root cause of today's boilerplate).
2. **Resolved the open questions** — deny-all topic ACL + `WithOpenTopics()`; `lvt` `Broadcast()` → `ReloadClients()`; wildcard topics in v1; (initially) per-connection `lvt:"local"`.
3. **Topics-only rewrite** (maintainer pivot) — collapse implicit peer-sync + topics into ONE primitive: `ctx.Subscribe`/`ctx.Publish`/`ctx.SelfTopic()` + `livetemplate.UserTopic/SessionTopic/GlobalTopic` + `handler.Publish`. Per-identity targeting is a derived topic string (like `Phoenix.PubSub`, verified). Self-sync = explicit `Subscribe(SelfTopic())` + `Publish` + a selective reconciler (Phoenix `handle_info` role). **Removed:** `BroadcastAction`, `SkipPeerSync`, `WithImplicitSyncDisabled`, `lvt:"local"`+merge, render-only mode, `GetByUserExcept`+empty-userID footgun, `RenderInvalidation`/`livetemplate:render:` channel. Added a 7-phase implementation tracker (gated by V1–V18). A prereview session resolved the dispatch model (named-action, not a `handle_info` inbox).

**Status of the former "open questions": all resolved** — see the proposal's §"Resolved decisions". Nothing in this PR is awaiting a maintainer decision; it awaits **sign-off review** of the resolved design.

Diff is the proposal doc plus a one-line `.gitignore` entry for the prereview tool's runtime scratch (`.prereview/`). The GitHub gross line count reflects the full multi-commit rewrite, not a small patch.

## Test plan

- [x] Pre-commit hook green on every commit (golangci-lint + full Go suite incl. Redis container tests) — docs-only, but the hook runs the suite regardless
[...]

PR #415, opened 2026-05-17
docs: restructure BroadcastAction→pub/sub proposal as a self-contained design+impl spec

adnaan:
## Summary

Post-merge audit + prereview restructure of the merged Publish/Subscribe topic
spec (`docs/proposals/broadcast-action-redesign-proposal.md`), before the
implementation PR begins. **Docs-only — no code changes.**

### Audit (code-grounded)
Verified the load-bearing claims against the codebase (they hold:
`handleDispatchedAction` runs vs the receiver's own state; topic & user actions
share `DispatchWithState`; the recursion guard; the single `processMessages`
pump; `Sync()` removed in #406; cross-repo site counts). Fixed 3 spec-blocking
defects + precision items, recorded in **Appendix C**:
- developer-grammar vs reserved-`lvt:` validator **order** (`Subscribe(SelfTopic())` would otherwise fail its own validator);
- the false "the parser already tracks wired action names" justification for the severe symmetry guard — guard kept, the small new parse-layer pass is now scoped honestly;
- the canonical `_ = ctx.Subscribe(ctx.SelfTopic())` idiom could silently swallow the empty-identity programmer error — now fail-closed + loud.

### Restructure (LLM-executable spec)
- Clean, present-tense, self-contained body; **History / Alternatives / Audit moved to an Appendix** (no prior-version or audit-process language in the body — verified by a cold-read pass).
- New **"The reconciler method"** section, defined before first use, incl. authenticated-vs-anonymous topic identities.
- All previously-deferred items resolved: **multi-segment wildcards promoted into v1**; no trie/debounce by design.
- Added **V20** removed-API cross-repo sweep and a **"Test tiers & harness"** subsection (Tier 1 Go integration, Tier 2 chromedp browser-e2e with the four mandatory captures, Tier 3 cross-repo CI) so the verification plan is concretely implementable.

Reviewed interactively over multiple prereview rounds; converged.

### Scope
This is the design+implementation **specification**. The implementation is a
separate future PR, scoped by the spec's "Phased implementation plan" and
"Impacted repositories".


adnaan:
Addressed the three "Issues worth discussing" from the review (`14e37cc2`):

1. **V15 grep defect** — removed the post-filter `| grep -vE 'sync\.[A-Za-z]|"sync"|\bsync\b'`. You're right: `\bsync\b` is dead (lowercase can't occur on a line selected by case-sensitive `\bSync\b`) and `sync\.[A-Za-z]` was unsafe (could mask a real stale-`Sync` line that also mentions `sync.Something`). The case-sensitive grep needs no filter; the spec now says so and explains why.
2. **V19 "wired to a client element"** — pinned a precise definition (form/`button` `name=`, `lvt-on:`, other `lvt-*` action attributes) and instructed the implementation to pin the exact `internal/parse/` node types so the V19 test asserts a fixed set, not a moving target.
3. **Multi-segment wildcard / V17 scope** — added a documented "minimum-viable subset": single trailing `*` is the only shape the v1 use cases strictly need; multi-segment stays the v1 target but Phase 3 can be trimmed by tightening *only* the grammar validator, no spec reopening. Matcher/registry/cross-instance stay general-case.

The Nits were acknowledgements of correct changes — no action needed. Strengths/Conclusion noted, thanks.

adnaan:
**Convergence boundary.** Round 2 confirmed all three round-1 issues resolved with "no new functional issues remain." `db46ab24` folds in the single round-2 minor note (the V15 `--include='*.md'` / Go-scaffold-coverage clarification) — the bot marked it "no push needed," but it prevents a real implementer question, so it's worth the one sentence.

This is the **final substantive push**. Per the project's AI-review anti-churn rule, any further review rounds that only reword or re-raise already-resolved points will be handled by reply, not by push. The spec is converged and ready to drive the implementation PR.

adnaan:
Round-3 was a fuller re-review (not a reword pass) and surfaced **three genuinely-new functional items** — the anti-churn rule's explicit carve-out ("a single round of clearly-substantive new feedback is not a convergence signal"), so the convergence boundary correctly yields to a substantive push (`da5328ad`):

1. **GET-phase `Publish` was unspecified** (the item you flagged as the real gap). Now explicit in §2: `Publish` is **not** a no-op on GET (unlike `Subscribe`) — it's transport-agnostic like `handler.Publish` and use case J; an unguarded `Publish` in `Mount` fans out on every page load, so side-effecting publishes MUST be guarded with `ctx.IsInitialMount()`/`IsReconnect()`. Made the rationale explicit (no-op-on-GET would be the surprising behavior — it'd break J's HTTP-POST path).
2. **Escape-valve precondition** — added the forward reference: the multi-segment risk valve composes *only because* Phase 0 implements the general-case `segmentMatch`/`byTopicPattern`; do not narrow them.
3. **Fan-out backpressure** — new Design-constraints bullet: `EnqueueDispatch` drop-on-overflow under broad/wildcard fan-out is the *identical pre-existing* `BroadcastAction` behavior (not new), surfaced by `wsBufferFull`/`wsSlowClientCloses` — documented as the accepted model so implementors don't treat it as a new problem.

Items 4–5 + Minor were acknowledgements of already-correct changes (item 4's split-coverage note was already added in `db46ab24`). **The convergence boundary now re-applies**: remaining rounds that only reword or re-raise resolved points will be handled by reply, not push.

adnaan:
**Termination boundary (final substantive push: `1db92656`).**

Round 4's three new items were genuine and are fixed:
1. **Reconnect** — reconnect-durable subscriptions must be in `Mount`; a `Subscribe` only in an action is lost after a WS drop (actions don't replay). Now explicit in §2.
2. **Dedup key** — `(instanceID, timestamp)` is **not** collision-free for same-instance same-tick publishes (the second would be wrongly dropped). The spec now mandates a per-instance monotonic `Seq uint64` envelope field; dedup id is `(instanceID, seq)`. Updated consistently in §2, §"Critical files", and §"Design constraints". *(This was the most important — a real correctness bug the implementation would otherwise have shipped.)*
3. **`ctx.Publish(GlobalTopic())` blast radius** — extended the Global-fan-out constraint: `ctx.Publish` from any action has the same all-users reach as `handler.Publish` (neither runs the ACL); send-side gating is required for both.
Plus two cheap clarity nits (§1 cross-ref to the Mount/`Publish` GET asymmetry; Tier-1 V13/14/15 parenthetical).

**This is the termination boundary.** Across R1–R4 the audit caught the spec-blocking defects (validator order, symmetry-guard justification, fail-closed `SelfTopic`, and now the dedup-key collision); every round the review has independently concluded the spec is *implementable and ready to drive the implementation PR*. Further full re-reviews of a doc this dense will keep surfacing ever-finer enhancement suggestions indefinitely — that is the anti-churn pattern CLAUDE.md addresses. From here, any remaining items that are **enhancement-tier (not a spec-blocking correctness defect)** will be acknowledged by reply and carried as *implementation-PR considerations*, **not** as further spec pushes. A genuinely new spec-blocking correctness defect (if any) would still be fixed; a reword/re-raise/finer-polish will not. The spec is converged.

adnaan:
**Round 5 confirms convergence — boundary holds, no push.**

Round 5 verified every audit fix as a real correctness fix and found **no new spec-blocking defect**. The remaining items are enhancement-tier, so per the termination boundary they're acknowledged here and carried as **implementation-PR / docs-rewrite considerations**, not a further spec push (another push only re-triggers the full-review cycle and its next batch of finer "minor observations" — the exact anti-churn pattern CLAUDE.md addresses):

- **§5 `ctx.Publish` severity cross-ref** — agreed and reasonable; the substance is already in §"Design constraints → Global fan-out". Logged for the docs rewrite (Phase 6) to mirror the one-sentence caution into the §5 API table — both are reference landing points. Not a correctness gap (the spec already states it).
- **Troubleshooting entry for subscribe-in-action-lost-on-reconnect** — good idea; folded into the Phase 6 docs scope (the spec body already states the rule in §2).
- **§B stronger no-debounce rationale** ("the framework can't know which publishes should coalesce") — a nicer rationale; reply-level, not spec-blocking.

These are tracked for the implementation PR (its §6 docs-rewrite step is the right home — the spec body itself is converged and already carries the load-bearing statements). The spec is **ready to drive implementation**. Closing the review loop here per the project anti-churn rule; thanks for the thorough multi-round pass — R4's `(instanceID, timestamp)` → `(instanceID, seq)` catch in particular was a real pre-implementation correctness save.

adnaan:
A full 3-way internal-consistency scan was run on the whole doc (not just the diff). The GlobalTopic-removal / ACL / anti-spoof area scanned **fully consistent**. This round (`8fd56eb9`) fixes every genuine internal inconsistency / factual defect found — including the verified ones from this review:

**Fixed (genuine consistency / factual defects):**
- **Correctness #1 (verified vs code):** §5 "Removed" falsely listed never-existent public APIs. `grep` confirms `BroadcastToUser`/`BroadcastGlobal`/`SkipPeerSync`/`WithImplicitSyncDisabled`/`RenderInvalidation`/`GetByUserExcept` are **absent in the tree** — only `ctx.BroadcastAction` exists. Reframed as "Not in this design" (one real removed API + design-alternatives with no codebase symbol); **V20 sweep rescoped to `ctx.BroadcastAction` only** (sweeping for never-existent names is vacuous, exactly your point).
- **Minor (verified):** `TestBroadcastAction_NoAutomaticPeerDispatch` is in `broadcast_test.go`, not the implied `context_broadcast_test.go` — corrected.
- **#12:** "relay" was undefined jargon — now defined at first use (the cross-instance step in `dispatchToTopic`).
- Scan-found: dangling `§6` numeric refs → `§"Impacted repositories"` (named-section convention, 8 sites); Tier-1 over-claimed V20 → `V16–V19` + V20 listed Tier-3; "Phases 0–5 = release steps 1–4" → "1, 2, 4"; Phase-2 envelope now enumerates **both** `Topic`+`Seq` (was Topic-only, contradicting §Critical-files); 3 body-purity archaeology phrases removed.
- Also: `SessionTopic` dropped per maintainer prereview decision (UserTopic kept — concrete use case F).

**Carried as implementation-PR considerations (acknowledged, not pushed — per the termination boundary):** these are spec-*completeness* gaps, not internal contradictions; the spec is implementable and the impl PR is their correct home: #2 parse-pass hook/lifecycle detail, #3 `Unsubscribe` per-connection scope, #4 `handler.Publish` signature/error, #5 a distinct `ErrInvalidTopic` vs `ErrTopicForbidden`, #6 reconnect exact-vs-pattern replay-window semantics, #7 `atomic.Uint64` for `Seq`, #9 `LiveHandler` interface-break note, #10 Mount-error HTTP behavior, #11 the `../docs/content` cross-repo grep command, #13 HTTP-POST `excludeConn=nil` asymmetry, RFC-2119 disclaimer, benchmark latency target. These are logged for the implementation PR; continuing to expand the spec each full-re-review round is the anti-churn pattern CLAUDE.md addresses. **`8fbd...`/`8fd56eb9` is the consistency-pass boundary**: further rounds that re-raise these completeness items will be handled by reply, not push.

adnaan:
Triage of the consistency-commit review (`3d98d26d`):

**Fixed (genuine):**
- **#1 V20 `--exclude-dir=proposals`.** Correct and important — V20's `BroadcastAction` sweep would fail its own gate against this very document (the §"Current architecture" grep anchor `^func \(c \*Context\) BroadcastAction`, the recursion-guard string, §4 migration text, and Appendix B/C all legitimately contain `BroadcastAction`). The grep now mandates `--exclude-dir=proposals` with that rationale spelled out, parallel to V15.
- **#3 `Seq` scope for non-topic `GroupActionMessage`.** Real ambiguity in a field this PR introduces on a shared envelope. Resolved per your suggested resolution: one per-instance counter incremented for *every* `GroupActionMessage` (topic and group-action); group-action messages consuming `seq` values is intentional and harmless because the double-fire dedup fires only on the topic `SUBSCRIBE`+`PSUBSCRIBE` double-delivery path (never group-action). One sentence added.

**Declined (already addressed — re-raise, per the anti-churn boundary):**
- **#2 "make the valve mechanism explicit in §2".** It already is. §2 "Grammar" states verbatim: *"…they can be deferred by tightening **only** this grammar validator to reject non-trailing or multiple `*` — no other spec change… Matcher, registry, and cross-instance machinery are written for the general case regardless (see Phase 0) — **do not narrow them to trailing-`*`**."* The valve = validator-only is explicit in §2 itself, not just inferable from Phase 0. No change; flagged here so the next round doesn't re-raise it.

**Minor notes:** all three are explicitly "no action required" / Phase-6 docs-PR items (pattern-scaffold cross-reference, metric names, `SessionTopic` rationale) — acknowledged, carried to the Phase-6 docs rewrite, not spec changes.

The spec is converged. Substantive new functional/correctness/consistency defects continue to be fixed (this round: 2); completeness/clarity re-raises are handled by reply per the established boundary. Conditional-approval items 1 & 3 are now addressed.

adnaan:
Triage of the latest review (`0ab0b6f1`):

**Fixed (genuine — self-defeating gate / coverage gap / load-bearing invariant):**
- **#1 V20 changelog.** Correct — `--exclude-dir=proposals` alone still self-defeats on `changelog.md` history (literal `BroadcastAction`). Added `--exclude='CHANGELOG*' --exclude='changelog.md'` and documented *both* exclusions as load-bearing for the gate to be runnable.
- **#2 reconnect-durability V-coverage.** Real gap: §2 states "subscribe-in-action is not reconnect-durable" as a MUST contract with no V-item. Added **V21** (Tier-1 Go integration: action-only subscribe → WS drop/reconnect → assert *not* re-delivered, vs Mount-subscribe re-delivered), wired consistently into the V-count (`V1`–`V21`), Tier-1 coverage, and the Phase-1 gate.
- **#5 `Seq` shared-counter.** Documented the load-bearing invariant: the shared `(instanceID, seq)` safety *assumes group-action stays exact-`SUBSCRIBE`-only*; if wildcard group-action is ever added, the dedup scope must be revisited before that ships. This converts your "future fragility" note into an explicit invariant in the envelope spec.

**Declined (already covered / boundary):**
- **#3 At-a-glance guarded-`Publish` example.** §1's "Mount asymmetry (footgun)" paragraph already states the rule explicitly with the `ctx.IsInitialMount()` guard; At-a-glance is deliberately minimal ("one concern, one code shape") — adding a Mount-`Publish` block there would dilute that. Covered, not added.
- **#4 V15 human-confirm labeling.** V15 already says "any residual hit is a literal capital-`Sync` token to **eyeball-confirm**" — the human-classification step is stated; the CI-assertable part is "grep returns no *lifecycle* hits". Labeling preference on already-conveyed content → reply, not push.
- **Minors** (V19 mock-vs-real-template; Appendix C praise) — implementor's call / acknowledgement; carried, no spec change.

Pattern holds: genuine new functional/correctness/consistency/coverage defects fixed each round; clarity/pedagogy re-raises declined by reply per the anti-churn boundary. The spec remains converged and implementable.

adnaan:
Triage of the latest review (`d8616546`):

**Fixed (genuine):**
- **#1 seen-set strategy (the one you flagged as the main pre-impl gap).** Pinned: a **bounded fixed-size most-recent-N ring** (small, e.g. N=64), **never an unbounded `map`** (which leaks one entry per cross-instance `Publish` forever). Also pinned **N>1**, not a single last-seen slot: concurrent publishes interleave the `SUBSCRIBE`/`PSUBSCRIBE` copies on the one multiplexed channel, so the duplicate can arrive a few messages later — the small ring absorbs that window. This removes the "implementation silently makes a wrong choice" risk.
- **#3 V8 cross-instance leg.** Correct — `handler.Publish(UserTopic(...))`'s primary value (use case F) is cross-instance. Added the required leg (publish on instance A → alice's connections on instance B, via the `lvt:user:` channel through Redis) and listed it in Tier-1's cross-instance set.
- **#2 GET-skipping ACL.** Added the security/UX consequence explicitly: an unauthorized user gets the initial render — a flash of the gated page's default state — before the WS refusal. It's not just perf-vs-early-rejection; it changes what an unauthorized visitor sees.
- **#4 / Phase-0:** V21 now states it reuses the existing `broadcast_test.go` WS-lifecycle harness (no new harness); Phase-0 gate gained a concrete failing case (`room/*` ∤ `room/42/log`) aligned with V17's wording.

**Declined (project guidance / convention — per the boundary):**
- **#5 restore "the `Sync()` half is stale post-#406" to the `sync-and-broadcast.md` docs-scope entry.** Declined: this directly conflicts with the standing **body-purity directive** (the body must not narrate the doc's own #406/Sync archaeology — the user spent multiple prereview rounds removing exactly this). Per CLAUDE.md, project guidance trumps a bot suggestion. The #406 rationale is preserved where it belongs: the single advisor-sanctioned §4 migration sentence (explains the non-uniform echo-method names) and Appendix A/C. The docs-scope list entry stays terse by design.
- **Nits** (`§"Name"` vs numbered-§ format; Appendix-C resolution-pointer): the `§"Name"` form is the deliberate, now-consistent convention for named sections (numbered §1–§5 only for the Proposed-design subsections — this was itself a consistency fix). Polish, not a defect. No change; flagged so it isn't re-raised.

Pattern holds: genuine new correctness/coverage/security-guidance defects fixed each round (this round: 4, incl. the memory-leak-prevention one you called the main gap); body-purity/convention re-raises declined with rationale.

adnaan:
Triage of round 3+ (`4d289801`):

**Fixed — genuine functional defect (security-relevant, good catch):**
The ACL hot-path mitigation was wrong. You're correct: a WebSocket upgrade handshake **is** an HTTP `GET` (RFC 6455), so the spec's `early-return on r.Method == "GET"` guidance would skip the ACL on the WS connection too — silently **disabling** it (skip-forever), not deferring it. The spec was actively prescribing a security-disabling pattern. Replaced with the correct distinguisher — the `Upgrade` header:
```go
if !strings.EqualFold(r.Header.Get("Upgrade"), "websocket") { return true, nil }
// real check — now runs only at WS connect/reconnect
```
This makes "defer the real check to the WS path" literally accurate (the ACL still enforces at WS connect/reconnect; only the plain page-render GET is skipped). The security/UX paragraph stays accurate — it describes this pattern's correct behavior.

**Nit — resolved honestly:** you flagged that V21's "no new harness needed" was an unverified assertion. Confirmed against `broadcast_test.go`: it has the `websocket.DefaultDialer.Dial` + `ws.Close()` primitives but **no** re-dial-same-session reconnect helper. Reworded V21: reconnect = re-dial with the **same session cookie** (so the server re-runs `Mount` for the same identity); if `broadcast_test.go` lacks that helper, scope a small one in Phase 1 alongside V21, not after.

**Everything else this round was validated correct** by the review (the `(instanceID, seq)` dedup, the bounded ring N>1 reasoning, the single-`*redis.PubSub` precondition, `GlobalTopic()` removal, `Publish`-on-GET footgun, V21 contract, `PublishGlobal` disambiguation, V8 cross-instance leg) — no action, acknowledged.

This was the highest-value catch of the loop after the dedup-key fix: the spec would otherwise have shipped guidance that disables the very ACL it mandates. Pattern holds — genuine functional/correctness/security defects fixed each round; the spec is converging.

adnaan:
**Convergence reached — stopping the push loop here (no commit for this round).**

This round's own verdict — *"Ready for the implementation PR. The convergence signal is met: this round surfaces no new functional issue"* — is the CLAUDE.md anti-churn stop signal. The two functional defects from the prior rounds (ACL `Upgrade`-header, V21 reconnect-helper scoping) are confirmed fixed in-diff; V15 grep simplification confirmed correct. The remaining items are explicitly non-blocking and are handled by reply, not another push (a 15th commit would only re-trigger a full re-review for items the review itself says don't block — the exact treadmill the project's anti-churn rule and the PR #412 precedent exist to stop):

- **Phase 3 → §2 valve cross-ref.** Fair navigability point; directional linkage already exists (§2's valve text explicitly says *"if Phase 3 pressure mounts…"*). Logged as a one-line addition for the implementation PR's Phase-3 checklist rather than a spec re-push. Self-consistent as-is (your words).
- **V19 enumeration pending.** You confirmed this is the right call — the spec must not enumerate parse node types the parser doesn't expose yet; the implementation PR closes the definition and makes V19 assert a fixed set. No spec change by design; dependency noted.
- **V8 Phase-2 gate cross-instance leg.** Phase 2 is the Redis phase, so the gate implicitly covers V8's cross-instance leg; the explicit phrasing is a clarity nicety for the impl PR, logged with the above.

These three are carried as implementation-PR considerations (the right home — they're impl-time concerns by the review's own framing). **This is the spec convergence boundary**: further rounds that re-raise non-blocking navigability/enumeration items will be handled by reply. The spec is internally consistent, audit-backed, and — per this review — ready to drive the implementation PR. Thank you for the multi-round pass; it caught a cross-instance dedup-collision bug, a memory-leak-prone underspec, and an ACL-self-disabling guidance defect, all pre-implementation.

PR #417, opened 2026-05-17
Phase 0: pub/sub topic foundations (registry indexes + topics.go)

adnaan:
## Context

**Phase 0** of the BroadcastAction → pub/sub topic-model redesign
(`docs/proposals/broadcast-action-redesign-proposal.md`). The proposal is built
as 7 dependency-ordered phases, **one phase per implementation session**; each
phase carries its own close-out record. This PR contains Phase 0 plus its
`learnings/phase-0.md` (the durable handoff Phase 1's audit reads) and the
`progress.md` tracker update.

## What's here (pure additions — no behavior wired)

No existing dispatch/broadcast path is touched. `ctx.BroadcastAction` and its
group-dispatch path remain **fully functional**; removal is atomic with
migration in Phase 5.

- **`internal/session/registry.go`** — `byTopic` + `byTopicPattern` indexes;
  `Connection.subscribedTopics` membership set (the GC root `Unregister()`
  walks); `SubscribeConnectionToTopic` / `UnsubscribeConnectionFromTopic` /
  `GetByTopicExcept` (deduped exact∪pattern union, idempotent **set** semantics
  — deliberately *not* the Phase-2 Redis ref-count); topic-GC wired into the
  existing `Unregister()` cleanup. Reuses `removeConnection` and the
  `GetByGroupExcept` copy contract. Pattern scan is flat O(P) — no trie, by
  design (proposal §2 "Matcher").
- **`topics.go`** (new) — `UserTopic`, `isReservedTopic`,
  `validateDeveloperTopic`, `segmentMatch` (general-case multi-`*`; Phase 3's
  risk valve narrows only the validator, never the matcher).
- **Tests** — `registry_topic_test.go` (6 tests, `-race`-clean) and
  `topics_test.go` (`TestSegmentMatch` = a 27-row executable spec, plus the
  validators/constructor).

## Gate status

Pre-commit hook ran in full (never `--no-verify`): `go fmt` + `golangci-lint`
(`errcheck,govet,ineffassign,staticcheck,unparam,unused`) **0 issues** +
non-race `go test -v ./... -timeout=300s` (incl. the Redis testcontainers
suite) — all green. `internal/session` also passed standalone `-race` (the only
concurrency-relevant Phase 0 code).

## Known pre-existing — NOT introduced here

A *full* `go test -race ./...` fails one root-package test,
`TestRangeBuildLatency_PostPhase7` (`range_build_latency_test.go`): a hard
wall-clock latency ceiling (50 ms / 250 ms) with no `-race`/`-short` guard, in
unrelated prior streaming-range work. **Verified pre-existing** by reproducing
it identically on a clean `main` checkout (no Phase 0 changes):
`median 109 ms` vs the `50 ms` ceiling — i.e. the race detector's
instrumentation overhead, not a regression, with zero code path from
`topics.go`/the registry. The pre-commit gate is non-race and passes it.
Out of Phase 0's scope guard to fix; recorded in the Ledger as a candidate
standalone follow-up (guard the latency test with `testing.Short()`/race-skip).

## Intentional handoffs to Phase 1 (recorded in `phase-0.md`, not omissions)

- **⚠ Latent subscribe-after-`Unregister` race.**
  `SubscribeConnectionToTopic` has no connection-liveness guard — *none is
  needed in Phase 0 (no caller)*. Phase 1's `ctx.Subscribe` exposes it; Phase 1
  must pick policy (a) registry `<-conn.done` short-circuit (mirrors
  `EnqueueDispatch`) or (b) `ctx.Subscribe` error return.
- **`DispatchRequest.Kind` deferred** — the proposal lists it *optional*; a
  zero-valued field with no consumer is dead code until Phase 1's dispatch path
  branches on it.
- **`GetByTopicExcept` takes an injected `match` parameter** — `internal/session`
  cannot import the root package's `segmentMatch` (import cycle), so the matcher
  is dependency-injected; Phase 1's `dispatchToTopic` passes `segmentMatch`.

## Scope

Phase 0 only (registry/topics substrate). Phases 1–6 land as separate PRs per
the proposal's protocol. `segmentMatch` was authored by the assistant at the
maintainer's explicit request (the learning-mode contribution was offered).

### DIFF
--- a/docs/proposals/broadcast-action-redesign-proposal/learnings/phase-0.md
+++ b/docs/proposals/broadcast-action-redesign-proposal/learnings/phase-0.md
@@ -153,10 +153,30 @@ exactly as the spec's contingency intends).
   other `lvt:` string. (2) else `validateDeveloperTopic(topic)` (segment
   grammar). (3) else the ACL (deny-all default; only `SelfTopic()` exempt). The
   grammar is **never** applied to `lvt:` topics (it excludes `:`).
+  **Layering (must hold):** validation runs **in `ctx.Subscribe`, before** the
+  `registry.SubscribeConnectionToTopic` call — the registry methods deliberately
+  do **not** validate (reserved identity topics like `UserTopic`/`SelfTopic`
+  legitimately bypass `validateDeveloperTopic`; validating inside the registry
+  would wrongly reject them and double-validate developer topics). The registry
+  is the index; `ctx.Subscribe` is the gate.
+- **Bare `*` is spec-permitted (confirmed against the proposal — not a gap).**
+  A single-segment `*` pattern passes `validateDeveloperTopic` and matches any
+  *single-segment* concrete topic. Proposal §2 "Grammar": segments may be `*`
+  "any number of times" — one `*` segment is valid. It is **bounded** (does not
+  match multi-segment topics — count mismatch) and **ACL-gated** (deny-all
+  default; a bare-`*` developer subscribe still needs `WithTopicACL`/
+  `WithOpenTopics`). Adding a validator rejection would *contradict the spec
+  grammar* and pre-empt Phase 3's valve (which narrows the validator only for
+  non-trailing/multiple `*`, never bare `*`). No change; recorded so Phase 1
+  and reviewers don't re-litigate.
 - **`dispatchToTopic` (Phase 1) must inject the matcher:**
   `registry.GetByTopicExcept(concrete, excludeConn, segmentMatch)`. The registry
   cannot reach `segmentMatch` itself (import cycle) — passing it is the caller's
-  responsibility; nil + pattern subscribers = panic by design.
+  responsibility; nil + pattern subscribers = panic by design. *Optional
+  (Phase 1's call):* a named wrapper `func topicMatch(p, c string) bool { return
+  segmentMatch(p, c) }` at the single call site gives audits a stable grep
+  anchor; passing `segmentMatch` directly is also fine (it is itself a named,
+  greppable symbol). Phase 1 decides — recorded so the option isn't lost.
 - **⚠ Latent race Phase 1 MUST resolve: subscribe-after-Unregister leak.**
   `Unregister()` sets `conn.subscribedTopics = nil` and drops the conn from all
   indexes. `SubscribeConnectionToTopic` has **no** connection-liveness guard
--- a/docs/proposals/broadcast-action-redesign-proposal/learnings/phase-0.md
+++ b/docs/proposals/broadcast-action-redesign-proposal/learnings/phase-0.md
@@ -169,6 +169,15 @@ exactly as the spec's contingency intends).
   grammar* and pre-empt Phase 3's valve (which narrows the validator only for
   non-trailing/multiple `*`, never bare `*`). No change; recorded so Phase 1
   and reviewers don't re-litigate.
+- **⚠ Phase 1: guard empty-`UserID` reserved topics.** `UserTopic("")` →
+  `"lvt:user:"` (the constructor is intentionally pure — proposal §5; no Phase 0
+  change). Phase 1's `ctx.Subscribe`/`SelfTopic()` MUST reject an empty-`UserID`
+  reserved subscribe before it reaches the registry, or `"lvt:user:"` could
+  match across anonymous connections. The spec already has the adjacent
+  invariant (proposal §1: anonymous `SelfTopic()` is `lvt:session:<GroupID>`,
+  never empty; an empty `SelfTopic()` from a misimplemented Authenticator is
+  logged `slog.Error`) — Phase 1 must additionally cover the out-of-band
+  `UserTopic("")` vector at the `Subscribe` gate.
 - **`dispatchToTopic` (Phase 1) must inject the matcher:**
   `registry.GetByTopicExcept(concrete, excludeConn, segmentMatch)`. The registry
   cannot reach `segmentMatch` itself (import cycle) — passing it is the caller's
--- a/docs/proposals/broadcast-action-redesign-proposal/learnings/phase-0.md
+++ b/docs/proposals/broadcast-action-redesign-proposal/learnings/phase-0.md
@@ -1,0 +1,226 @@
+# Phase 0 — Learnings
+
+**Session:** 2026-05-17 / claude-code   **Status at exit:** in progress
+**Plan reference:** Phase 0 in docs/proposals/broadcast-action-redesign-proposal.md
+
+## Audit (start) — findings
+
+Read-only verification against live code (worktree `broadcast-redesign-phase-0`,
+branched from `main` @ `19774be3`). No prior phase to read.
+
+- **No symbol collision.** `grep -rn 'byTopic\|byTopicPattern\|segmentMatch' internal/session/`
+  → none. `topics.go` does not exist at repo root. Safe to add.
+- **`ConnectionRegistry`** (`internal/session/registry.go:319-325`): fields
+  `byGroup`, `byUser`, `mu sync.RWMutex`, `metrics`, `dispatchBufferSize`.
+  `NewConnectionRegistry()` (`:330-335`) initializes `byGroup`/`byUser`.
+- **`Unregister()`** (`registry.go:392-426`) is the **sole** index-cleanup path.
+  Removes from `byGroup`/`byUser` under `r.mu.Lock()` via `removeConnection`,
+  deletes emptied map keys, then calls `conn.Close()`.
+- **Risky assumption — confirmed.** `Connection.Close()` (`registry.go:182`) only
+  signals writePump shutdown + closes the socket; it does **not** mutate registry
+  indices and does **not** call `Unregister()`. The slow-client path
+  (`Send()` → `go c.Close()`, ~`registry.go:146`) closes the socket but index
+  removal still flows through `Unregister()` later. **No third path bypasses
+  `Unregister()`** → topic-map cleanup wired there is necessary and sufficient.
+- **`removeConnection(conns, target)`** (`registry.go:572`): reused verbatim for
+  the topic-map slice removal.
+- **`DispatchRequest`** (`registry.go:69-72`): `{ Action string; Data map[...] }`.
+  No `Kind` field today (see Deviations — deferred).
+- **Surface is `UserTopic` only.** No `SessionTopic`/`GlobalTopic` references in
+  the tree to mirror (spec §5 / Appendix B forbid them).
+
+The audit did **not** reshape the task list — the plan's predicted reuse targets
+and grep anchors all held against live code.
+
+## What shipped
+
+Pure-additive infrastructure; **no existing dispatch/broadcast path touched**
+(`BroadcastAction` fully functional, as required through Phases 0–4).
+
+- **`internal/session/registry.go`**
+  - `ConnectionRegistry.byTopic` + `byTopicPattern map[string][]*Connection`;
+    initialized in `NewConnectionRegistry()`.
+  - `Connection.subscribedTopics map[string]struct{}` — per-connection
+    membership set; the GC root `Unregister()` walks. Guarded by the existing
+    `ConnectionRegistry.mu` (no new lock).
+  - `SubscribeConnectionToTopic` / `UnsubscribeConnectionFromTopic` /
+    `GetByTopicExcept` + unexported `isPatternTopic` helper. Reuses
+    `removeConnection`; mirrors the `GetByGroupExcept` copy contract and the
+    `byGroup`/`byUser` empty-slice-delete idiom.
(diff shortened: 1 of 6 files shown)


adnaan:
Thanks for the thorough review. Resolution per point (pushed in `ad34d977`):

**Addressed (committed):**
- **Test gap — `GetByTopicExcept` zero subscribers:** added `TestGetByTopicExcept_EmptyRegistry` (fresh registry + "pattern present, no match" — the path the first Phase 1 publish hits).
- **Test gap — unsubscribe non-nil-map path:** added `TestUnsubscribeConnectionFromTopic_DifferentTopicKeepsOthers` (subscribed to A, unsubscribe unrelated B, A survives). The existing `…_NoOpWhenNotSubscribed` already covers the pure nil-map path (`&Connection{}` ⇒ `subscribedTopics == nil`); this adds the non-nil sibling.
- **Validation layering (Phase 1 handoff):** `phase-0.md` now states explicitly that `ctx.Subscribe` validates **before** the registry call; registry methods deliberately do not validate (reserved identity topics bypass `validateDeveloperTopic` by design).
- **Injected-matcher named-wrapper option:** recorded as a Phase 1 call in `phase-0.md`.

**Already covered (no change — verifiable):**
- **Map initialization:** `byTopic`/`byTopicPattern` are initialized in `NewConnectionRegistry()` alongside `byGroup`/`byUser` (and `phase-0.md` "What shipped" states it). `-race` green confirms.
- **Trailing/leading slash:** `TestValidateDeveloperTopic`'s invalid list already includes `"room/"`, `"/room"`, `"room//x"` (empty-segment rejection).
- **Subscribe-after-`Unregister` race:** flagged in `phase-0.md` exactly as you describe (mirror `EnqueueDispatch`'s `conn.done` check, MUST land in Phase 1 before `ctx.Subscribe` is public). No Phase 0 caller exists, so adding the guard now would be behavior with no exerciser — deliberately deferred, not missed.

**Declined, with rationale:**
- **Reject bare `"*"` in the validator:** declining — this would *contradict the proposal*. §2 "Grammar": a segment is `[a-zA-Z0-9_-]+` **or** `*`, and `*` may appear "any number of times" — a single `*` segment is grammatically valid. It is bounded (matches only single-segment concretes — count mismatch otherwise) and ACL-gated (deny-all default). A rejection would also pre-empt Phase 3's documented risk valve, which narrows the validator only for non-trailing/multiple `*`, never bare `*`. Rationale recorded in `phase-0.md` so it isn't re-litigated.
- **Pre-allocate `cap = len(byTopic)+len(byTopicPattern)`:** declining for Phase 0 — you correctly flagged this as out-of-Phase-0-scope future work. `len(byTopicPattern)` counts distinct *patterns*, not connections, so it's a loose (often over-) bound; the current `len(byTopic[concrete])` lower-bound hint is appropriate for the substrate. Revisit under the Phase 2 benchmarks if profiling warrants.

**Out of scope (recorded, not fixed here):** the pre-existing `-race`-only `TestRangeBuildLatency_PostPhase7` flake — reproduced identically on clean `main`, unrelated to topics, non-race gate (= pre-commit + CI Unit Tests) passes. Ledger-tracked as a candidate standalone follow-up.

Net: 2 real coverage gaps closed; everything else either already covered, a correctly-deferred Phase 1 handoff, or a spec-grounded decline.

adnaan:
Round 2 addressed (`b5428e57`):

- **Cap hint widened** to `len(byTopic[concrete]) + len(byTopicPattern)` — accepted; you're right that a pattern-heavy publish would realloc on the first pattern hit.
- **Comment trim pass** per CLAUDE.md "no unnecessary comments": `subscribedTopics` 7→3 lines (kept the non-obvious WHY: GC-root + `r.mu`-guarded), `SubscribeConnectionToTopic` doc drops the ref-count paragraph (it lives in `phase-0.md`), `segmentMatch` CONTRACT 17→~8 lines pointing at `TestSegmentMatch` as the exhaustive spec.
- **Nil-match contract**: added `TestGetByTopicExcept_NilMatchSafeWhenNoPatterns` documenting that `nil` match is safe iff no pattern subscribers (panic-by-design otherwise).
- **`UserTopic("")` empty-UserID**: agreed the constructor stays pure (proposal §5); recorded as a Phase 1 `ctx.Subscribe`-gate handoff in `phase-0.md` (reject empty-`UserID` reserved topics before the registry; complements the spec §1 anonymous-`SelfTopic` invariant).
- **subscribe-after-`Unregister`**: confirmed — already the headline Phase 1 must-fix in `phase-0.md`/`progress.md`; no Phase 0 caller exists so it's correctly deferred, not missed.

Correctness was confirmed bug-free; remaining items were style/handoff and are now resolved or recorded.

adnaan:
Round 3 addressed (`359347a8`). Notably, you caught that **round 2's own cap suggestion was wrong** — good catch:

- **Cap hint**: reverted to the honest `len(byTopic[concrete])` and dropped the false "covers exact + pattern" comment. `len(byTopicPattern)` is a pattern-key count, not a connection count, as you note. An accurate pattern cap would need a second O(P) pre-count loop (the function already scans patterns once) — not worth it for a hint when `seen` dominates allocation. Took your "simply drop it" option.
- **nil `match`**: added an explicit guard with a named, actionable message (points Phase 1 at `dispatchToTopic`/`segmentMatch`). Placed **before** the loop, `len(byTopicPattern) > 0`-guarded — one check instead of per-iteration, and the nil-safe-when-no-patterns contract (+ its test) still holds.
- **`isPatternTopic` comment**: 3 sentences → 1.
- **`slices.Equal`**: replaced the hand-rolled `equalStrings` (go 1.26).
- **Concurrent test**: added `TestConcurrentTopicSubscription` — 8 workers × 200 iters of subscribe/unsubscribe on disjoint (conn, topic) pairs + a concurrent `GetByTopicExcept` reader, deterministic empty-registry assertion, `-race`-clean. Validates Phase 0's lock discipline (it deliberately does **not** exercise the Phase-1-deferred subscribe-after-`Unregister` race — no `Unregister` caller here).

Thanks for the detailed positives call-out too — the import-cycle DI, Topic-GC completeness, and idempotent-semantics testing were the design decisions I most wanted a second pair of eyes on.

adnaan:
Round 4 — both documentation items fixed (`f4126d83`); agree this is the convergence point (no new functional issue across the last two rounds):

- **`phase-0.md` status** → `complete` (was inconsistent with `progress.md`; since Phase 1's audit reads `phase-0.md` first, a stale "in progress" would have falsely signalled an unfinished phase — good catch).
- **Panic message de-rotted** → `"session: GetByTopicExcept requires a non-nil match when pattern subscribers are indexed (callers must pass the segment matcher)"`. Dropped the `Phase 1 dispatchToTopic` forward-reference (would have rotted once Phase 1 lands); states the stable invariant + caller obligation, no phase coupling. Per CLAUDE.md's own anti-rot guidance, did it now rather than defer — it's a one-liner.

No functional change; no test asserts the panic text; CI green, `MERGEABLE`/`CLEAN`. From the PR side this is converged — remaining "must-fix" is the Phase 1 subscribe-after-`Unregister` guard, which is correctly out of Phase 0 (no caller) and is the headline handoff in `phase-0.md`/`progress.md`.

adnaan:
Round 5 — all four trivial items fixed (`4067a4b2`):

- `TestUnregister_TopicGC` now comments the bare-`Connection`/`Close()`-nil-guard dependency (the dependency is real and verified in the Phase 0 audit; the comment protects the next reader).
- `diff` → `ids` (it's sorted IDs, not a diff).
- `slices.Sort` replaces `sort.Strings`, `sort` import dropped (go 1.26).
- `isValidSegmentChar` comment trimmed to the non-obvious WHY (`:`/`*` exclusions); dropped the `[a-zA-Z0-9_-]` line that just restated the `switch`.

**Convergence reached.** Two consecutive rounds (R4, R5) with no new functional issue, across 5 review rounds; correctness confirmed bug-free; CI green and `MERGEABLE`/`CLEAN` throughout; benchmarks show no regressions. Per the project's review-loop guidance I'm treating this as converged and will not push further cosmetic-only rounds — any remaining nits I'll address by reply.

Outstanding by design (not Phase 0): the subscribe-after-`Unregister` liveness guard is Phase 1's must-fix (no caller exists in Phase 0); `phase-0.md` records both policy options and instructs Phase 1 to record its chosen one in `phase-1.md`. Phase 0 is the substrate; this PR is ready.

adnaan:
Round 6 — comprehensive pass done (`3accab9d`):

- **Comment de-rot (CLAUDE.md "no task/caller refs"):** swept *all* Phase 0 code comments — `topics.go`, `registry.go`, `registry_topic_test.go` — and stripped every `Phase N` / `phase-0.md` / `Phase 0 audit` / caller-prediction reference, **including the "Verified in the Phase 0 audit" line I introduced in round 5** (good catch — that was my regression). Kept only stable invariants and durable spec anchors (`proposal §2/§5`), per your "keep the invariant statements" guidance.
- **nil-match contract:** `GetByTopicExcept` doc now states it self-contained — nil is safe **only** with zero pattern subscribers, a time-of-call property (safe until the first pattern subscriber, then panics).
- **`isPatternTopic`:** comment now describes what the function does; dropped the upstream-validator precondition (you're right — correctness doesn't depend on the validator having run).
- **`Unregister`:** comment trimmed to the one non-obvious point (many-to-many → no O(1) reverse lookup).
- **Test gap:** added `TestGetByTopicExcept_ExcludesSenderViaPattern` (publisher matched via pattern only → still excluded).
- `testSegmentMatch` "contributed" → "root-package"; `slices.Sort`/var-rename from round 5 retained.

Correctness has been confirmed bug-free for three consecutive rounds (R4–R6); the only items have been documentation/contract phrasing, now comprehensively resolved. Per the project's review-loop convergence rule this is the closeout: I won't push further cosmetic-only rounds — any residual wording nits I'll handle by reply. **Out of Phase 0 by design:** the subscribe-after-`Unregister` liveness guard (Phase 1 must-fix; no caller here) and the `UserTopic("")` subscribe-gate — both recorded in `phase-0.md` as hard Phase 1 handoffs. Phase 0 substrate is correct, green, and ready.

adnaan:
Round 7 — took the one actionable item (`ca639476`):

- **`GetByTopicExcept` `concrete`-wildcard guard:** added a leading panic so the documented "concrete, no `*`" precondition is now *enforced*, symmetric with the existing nil-match guard — **both** of the method's documented preconditions are now self-enforcing. Did it now rather than deferring (unlike the subscribe-after-`Unregister` guard, this has zero design decision — a trivial completing 1-liner — and it makes Phase 1's `dispatchToTopic` self-validating from the first call, exactly as you noted). Added `TestGetByTopicExcept_PanicsOnWildcardConcrete`; no behavior change for valid concrete inputs (existing tests unaffected).
- Findings 2–4 (subscribe-after-`Unregister`, unexported validators, `testSegmentMatch` duplication): confirmed as you noted — deferred-by-design / intentional / correct-by-necessity, no action.

This is the closeout. Four consecutive correctness-✅ rounds (R4–R7), CI green and `MERGEABLE`/`CLEAN` throughout, comments de-rotted, both `GetByTopicExcept` preconditions guarded, the one bot-flagged test gap closed. Per the project's convergence rule I'm not pushing further rounds — any residual wording nit gets a reply. Phase 0 substrate is correct, green, bot-approved ("Ready to merge"), and the Phase 1 hard handoffs (subscribe-after-`Unregister` policy choice; `UserTopic("")` subscribe-gate) are recorded in `phase-0.md`/`progress.md`.

adnaan:
Round 8 — **Approved, no functional issues** (5th consecutive correctness-clean round, R4–R8). This is the definitive closeout.

Per the project's review-loop convergence rule (recognize when successive rounds surface no new *functional* issue and stop pushing): the one remaining item — `seen` allocated unconditionally — is explicitly **low-priority / "follow-up if it shows up in profiles"** by your own framing, and the proposed lazy variant trades a negligible map alloc for per-iteration nil-branches in a hot loop. **Declining for Phase 0**, recorded here in the PR for the record: the proposal already mandates Phase 2 fan-out/pattern-scan benchmarks — that is the correct, data-driven place to revisit `seen` allocation if it actually appears in a profile. Optimizing it now would be speculative (no wildcard subscribers exist until Phase 1+).

All other round-8 notes are either "✓ correct" confirmations (nil-map reads, `Unregister` nil-range, lock ordering of the two `GetByTopicExcept` guards) or reiterations of the **known, documented Phase 1 obligations** — subscribe-after-`Unregister` policy choice and the `UserTopic("")` subscribe-gate — both already hard handoffs in `phase-0.md`/`progress.md`.

**Closeout state:** 8 commits, CI green + `MERGEABLE`/`CLEAN` throughout, lint 0, `internal/session` `-race`-clean, both `GetByTopicExcept` preconditions self-enforcing, comments de-rotted, bot-approved. No further pushes — Phase 0 substrate is done; the Phase 1 audit has its hard handoffs. Thanks for the thorough multi-round review.

PR #421, opened 2026-05-18
fix(pubsub): synchronize the reconnectHook test seam (data race, #420)

adnaan:
Closes #420.

## Problem

`go test -race ./...` (the cross-repo "Test LVT against Core Changes" CI job) reliably reports a real `WARNING: DATA RACE` → `--- FAIL: TestReconnect_DoesNotBlockConcurrentPublish` (`pubsub`). It is **pre-existing on `main`** and surfaces on any PR whose CI runs `-race ./...` (e.g. #419), independent of that PR's content.

## Root cause

`reconnectHook` was a bare package-global test seam:

- **Production read** — `redis.go` `reconnect()`: `if reconnectHook != nil { reconnectHook() }`, executed on the `processMessages` goroutine.
- **Test write** — the three `TestReconnect*` tests install a hook and `defer` restoring it from the test goroutine.

Nothing synchronized the two. Under CI's full-suite `-race ./...` contention (a live `processMessages→reconnect()` goroutine concurrent with a test's deferred hook restore) the detector catches the unsynchronized global. It does not reproduce in isolated local runs (green 15× under `-race`) because the race window only widens under whole-suite contention.

## Fix

Replace the bare `var reconnectHook func()` with a mutex-guarded accessor pair:

- `currentReconnectHook() func()` — production reads via this.
- `setReconnectHook(h) (prev func())` — tests install via this; returns the previous hook for deferred restore.

The variable is now only ever read/written under `reconnectHookMu`, so the race is eliminated by construction. No reconnect-logic change; purely synchronizes the existing test seam.

## Verification

- `TestReconnect*` pass `go test -race -count=10 ./pubsub/`.
- `go build ./...`, `go vet ./pubsub/`, and the full pre-commit gate (`go fmt` + `golangci-lint` enable-only set 0 issues + full `go test -v ./... -timeout=300s` incl. Redis testcontainers) all green.

Standalone fix off `main`, deliberately scoped to the `reconnectHook` seam only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #419, opened 2026-05-18
feat(topics): Phase 1 — Context API + topic ACL (single-instance)

adnaan:
Implements **Phase 1** of the Publish/Subscribe topic-model spec (`docs/proposals/broadcast-action-redesign-proposal.md`, #415). Phase 0 (registry + `topics.go` substrate) landed in #417; this builds the single-instance behavior on top. `ctx.BroadcastAction` is **untouched** (removed in Phase 5).

## What's in this PR

- **`ctx.Subscribe / Unsubscribe / Publish / SelfTopic`** (+ `WithGroupID`) — exact validation order: reserved-`lvt:` anti-spoof (exact-self only) → developer-topic segment grammar → ACL.
- **`WithTopicACL` / `WithOpenTopics`** — deny-all default; both-set is an **order-independent hard error at `New()`**.
- **`dispatchToTopic`** — local fan-out, structural twin of `dispatchBroadcastToGroup`; deduped exact∪pattern union via the injected `segmentMatch`; drained at the post-`persistState` `processBroadcasts` site (persist-before-publish).
- **`Publish` recursion guard** — mirrored at both the dispatched-action and server-initiated-action sites (post-hoc drop+`slog.Error`, matching the existing `BroadcastAction` guard).
- **Symmetry-collision warning** — `Publish` emits `slog.Warn` when the action name is wired to a client element (statics scan parallel to `ExtractFormSchema`).
- **Server-side WS error envelope** — `{"type":"error","code":"topic_forbidden","topic":…}` on the ACL-denied WS-connect path (`*TopicForbiddenError` carries the topic; `errors.Is(…, ErrTopicForbidden)` still holds). The TS client consumer is Phase 4.
- **Registry** — `<-conn.done` subscribe-after-`Unregister` liveness guard (silent-drop policy); `DispatchRequest.Kind` placeholder.

## Verification

Gate (Phase 1 = **V1–V7, V10–V12, V16, V19, V21**): green.

- 11 new Tier-1 integration tests in `topic_test.go` — all pass.
- Pre-commit hook (`go fmt` + `golangci-lint` enforced set **0 issues** + full `go test -v ./... -timeout=300s` incl. Redis testcontainers) — passes.
- `-race` scoped to `internal/session` + topic/broadcast tests — clean.
- Existing suite (incl. all `BroadcastAction` tests) — regression-free.

[...]

PR #424, opened 2026-05-19
feat(topics): Phase 2 — cross-instance Redis (Subscribe/Publish over livetemplate:topic:{name})

adnaan:
Phase 2 of the Publish/Subscribe topic model (#415) — `docs/proposals/broadcast-action-redesign-proposal.md`. Builds on Phase 0 (#417) + Phase 1 (#419), atop the #421/#423 pubsub race fixes already on `main`. Per the proposal's per-phase **Audit → implement → Learnings** protocol; `learnings/phase-2.md` ships in this PR.

`ctx.BroadcastAction` and its group-dispatch path are **untouched** (still fully functional, as required through Phases 0–4).

## What shipped

- **`pubsub/types.go`** — `GroupActionMessage` gains `Topic` + `Seq` (per-instance atomic dedup counter, incremented on **every** emit — group-action *and* topic, per the spec's "Seq scope"; the shared-counter load-bearing invariant is documented on the struct). The existing `groupID` json tag is unchanged (BroadcastAction wire bytes preserved). Two new optional interfaces: `TopicActionBroadcaster`, `TopicChannelSubscriber`.
- **`pubsub/redis.go`** — one `livetemplate:topic:{name}` channel (exact `SUBSCRIBE`; `PSUBSCRIBE`/wildcards are Phase 3); `PublishToTopic`/`SubscribeToTopicActions`/`SubscribeToTopicChannel`/`UnsubscribeFromTopicChannel` reusing `subscribeTo`/`unsubscribeFrom` → **free reconnect-replay** via `subscribedChannels`; `"topic_action"` routing in the single pump.
- **`mount.go` / `topic_runtime.go` / `template.go`** — `dispatchToTopic` cross-instance leg; `handleTopicActionMessage`; `LiveHandler.Publish` out-of-band entry (no Context/ACL/recursion-guard); exactly-once Redis relay via a per-connection registry transition-bool + a LIFO-ordered disconnect sweep (`internal/session` can't import `pubsub`, so teardown is root-package-driven).
- **`internal/session/registry.go`** — `SubscribeConnectionToTopic`/`UnsubscribeConnectionFromTopic` return the per-connection 0→1 / 1→0 transition `bool`; new `SubscribedTopics` snapshot accessor.
- **`topic_wired_actions.go`** — landed the deferred Phase-1 round-4 fix: the symmetry-collision warning canonicalizes via `toSnakeCase` (the form `methodNameToActions` reduces every style to), so a style-mismatched collision (`Publish "Save"` vs `<button name="save">`) is now caught. No false positives; V19 stays green.

## Verification

- **Gate (V8, V9): GREEN** — `TestTopic_V8_OutOfBandHandlerPublish_{Announcements,UserTopicCrossInstance}` + `TestTopic_V9_CrossInstancePublish` against two real `RedisBroadcaster`s on one Redis testcontainer.
- 7 deferred-coverage tests (invalid-topic, cap, Unsubscribe, nil-r deny, upload-complete drain, topic_forbidden envelope, round-4 normalization) — pass.
- Full `scripts/pre-commit.sh` green: `go fmt` + `golangci-lint --enable-only=errcheck,govet,ineffassign,staticcheck,unparam,unused` **0 issues** + full `go test ./...` incl. Redis testcontainers. Scoped `-race` green (pubsub / internal/session / topic). Phase-1 V-items regression-green.

## Benchmarks (§"Benchmarks required" items 1 & 3; item 2 = Phase 3)

| Topic fan-out (local enqueue) | ns/op | B/op | allocs/op |
|---|---|---|---|
| N=1 | 922 | 232 | 6 |
| N=5 | 2 392 | 400 | 10 |
| N=10 | 4 999 | 920 | 18 |
| N=50 | 21 917 | 4 328 | 62 |
| N=100 | 99 037 | 8 744 | 114 |

**Cross-instance round-trip (single Redis, 50 iters):** topic **250.7µs** vs `PublishGroupAction` baseline **259.9µs**, ratio **0.96×** — no measurable overhead (the topic path is the same envelope/pump/`publishJSON` machinery with a different `Type` + routing key).

## Scope notes

- **Seen-ring `(instanceID,seq)` dedup is Phase 3**, not here (advisor-confirmed): it is the *consumer* of the `SUBSCRIBE`+`PSUBSCRIBE` double-fire, which only exists once Phase 3 adds `PSUBSCRIBE` (gated by V17). Phase 2 ships the wire plumbing + counter, cross-instance-proven by V8/V9; Phase 3 inherits a stable, correctly-keyed envelope.
- The `Seq` addition to `PublishGroupAction` is **spec-directed** (§"Critical files"), an additive backward-compatible JSON field — "BroadcastAction untouched" governs the API/semantics/dispatch path, all unchanged.


adnaan:
Thanks for the review — addressed in `036c1a22`:

**Fixed**
1. **Misleading "no extra unmarshal" comment** — corrected on both `handleGroupActionMessage` and `handleTopicActionMessage`. The comment now states accurately: `handleMessage` does a partial `Type`+`InstanceID` parse to route and drop own-instance messages; the handler does a fresh *full* unmarshal; the `msg.InstanceID` re-check is a cheap redundant guard against the already-decoded struct (no separate unmarshal *for the guard itself*).
2. **Shared `Seq` Phase-3 constraint** — flagged at the `seq` field: values are monotonic **per-instance only**, never per-Type (group + topic emits interleave through one counter, so a topic stream sees gaps); the Phase-3 `(instanceID, seq)` ring must not assume per-stream/contiguous `seq`.
3. **`GroupActionMessage` tagged-union doc** — added an explicit field-population-by-`Type` note (`"group_action"` → `GroupID`; `"topic_action"` → `Topic`).
4. **Bench `time.Sleep(300ms)` fragility** — replaced with a blocking warmup round-trip (retry publish until first delivery confirmed, 15s deadline). The fixed sleep is no longer load-bearing; topic pub/sub has no replay, so re-publishing is the correct readiness probe. (The V8/V9 *correctness* tests already used the retry pattern; only this *latency* test had the sleep, since a retry loop can't be inside the timed section without distorting the measurement — the warmup runs before timing.)
5. **`BenchmarkTopicFanoutByN` cleanup** — explicit panic-safe `b.Cleanup` that unregisters connections and stops the drain goroutines.

**Declining one item with rationale: the double-unmarshal *refactor*** (decode-once in `handleMessage`, route on the decoded struct).

This is a **pre-existing pattern**, not introduced by this PR — `handleMessage`'s partial-parse-then-full-decode is the established shape for `handleServerActionMessage`/`handleGroupActionMessage`/`handleBroadcastMessage`; Phase 2's topic handler mirrors it deliberately for consistency. Collapsing it to a single decode would rewrite the routing for the **`BroadcastAction` group-action and server-action dispatch paths**, which the pub/sub-topic rollout's governing constraint keeps **untouched through Phases 0–4** (the proposal's atomic-removal/no-intermediate-build requirement; `BroadcastAction` is migrated and its internals are fair game only in Phase 5). A cross-path hot-path refactor is out of scope for an additive Phase-2 PR and would broaden the blast radius well past topics. I took the review's explicit fallback instead — the double-decode is now documented at the call sites as a deliberate, known cross-instance cost with the Phase-5 rationale. (Ref: `docs/proposals/broadcast-action-redesign-proposal.md` §"Phase 5 — Removal", and CLAUDE.md "project guidance trumps bot suggestions when they conflict — cite the rule and decline".) Happy to file a standalone follow-up issue for the decode-once optimization so it isn't lost.

adnaan:
Round 2 follow-up (`9895e8e9`):

The full review comment didn't post — only the summary placeholder landed ("One functional issue (slotKey unused in `awaitTreeSlot`), two style nits, and one Phase 3 handoff note"; the "see next comment" follow-up never arrived). Acting on what was delivered:

- **Functional issue — `slotKey` unused in `awaitTreeSlot`: fixed.** Renamed `awaitTreeSlot` → `awaitWSContains` and removed the dead parameter. The helper intentionally substring-matches `want` against the raw WS frame rather than asserting `tree[slotKey]==want`: a tree update carries only the *changed* dynamics and the value may be nested, so a raw-body contains check is the robust "value landed end-to-end" signal these cross-instance retry-loop tests need. Documented that rationale on the helper. No behavior change; V8/V9 green; full pre-commit gate green.
- **Two style nits / Phase 3 handoff note:** not actionable — the detail comment never posted (only the placeholder). If the next review re-raises them with specifics I'll address them; the Phase 3 handoff note is informational and is already captured in `learnings/phase-2.md` "Adjustments recommended for the next phase" + the seeded Phase 3 row in `progress.md`.

adnaan:
Round 3 (`e7c8ca25`). This review surfaced **no new functional issues** — confirmations, "Minor" nits, and forward-looking Phase 3 notes — so per the CLAUDE.md bot-convergence rule this is the convergence point: the 3 genuinely valuable items are pushed below; the rest are addressed by reply (not churned).

**Fixed (pushed):**
- **`seq=0` / pre-upgrade-sender invariant (the real Phase 3 footgun you spotted — good catch):** documented at the `seq` field. A rolling-upgrade instance on pre-`Seq` code omits the field (JSON→0), so *every* message from it has seq=0; the Phase 3 `(instanceID, seq)` ring **MUST bypass dedup when seq==0** (process unconditionally — a pre-Phase-2 instance has no topic `PSUBSCRIBE`, hence no double-fire). Also noted `Add(1)` ⇒ live instances never emit seq=0. Carried into `learnings/phase-2.md` "Adjustments for Phase 3" + the seeded `progress.md` Phase 3 row so the Phase 3 audit can't miss it.
- **Latency test:** now `t.Skip`s under `-short` (timing-sensitive informational measurement, not a hard gate — exactly your point about CI noise).
- **Log-level asymmetry:** commented — subscribe failure silently breaks cross-instance delivery (Error); unsubscribe failure only leaks a recoverable refcount (Warn).

**Addressed by reply (no churn):**
- **Deferred-ordering comment "missing":** it's already there — `mount.go` directly above `defer h.releaseRelayedTopics(connection)`: *"Registered after the Unregister defer so LIFO runs this first, while conn.subscribedTopics is still live (Unregister nils it)..."*. That comment exists precisely to stop the "future maintainer inserts a defer between them" hazard. No change needed.
- **Double unmarshal:** you concurred it's the intentional Phase-0–4 BroadcastAction-untouched tradeoff; documented at the call sites. No action (the round-1 decline holds).
- **Drop the redundant own-instance guard in `handleTopicActionMessage`:** declining — it's deliberately symmetric with the pre-existing `handleGroupActionMessage` guard (same defense-in-depth, same shape). Dropping it from only the topic handler would make the two siblings inconsistent for no functional gain.
- **`awaitWSContains` could mask duplicate delivery:** correct, and correctly out of scope for Phase 2 — exact `SUBSCRIBE` only means no double-fire path exists until Phase 3 adds `PSUBSCRIBE`. Phase 3's **V17** explicitly asserts exactly-once for a dual-subscribed connection; the count assertion belongs there, with the dedup ring.
- **`canonicalActionKey` store name→canonical:** "not needed today" per your note — the warning logs the raw action arg (original form preserved in the message); only detection/dedup keys on the canonical form. Deferred unless a future need arises.
- **`SubscribedTopics` returns `nil` for empty:** intentional — `range nil` is a no-op, so the disconnect sweep needs no empty-slice allocation.
- **`isReservedTopic` skips developer-grammar in `liveHandler.Publish`:** intentional and not a validation gap — reserved `lvt:` topics have a different grammar (contain `:`, which the developer segment grammar excludes) and are validated by the reserved-namespace rule, not the segment grammar. It exactly mirrors `ctx.Publish` (send-side is permissive on `lvt:`; anti-spoof is a Subscribe-side rule — proposal §2/§3).

Every commit this cycle passed the full `scripts/pre-commit.sh` gate (fmt + golangci-lint 0 issues + `go test ./...` incl. Redis testcontainers); CI has had zero failures; PR stays `MERGEABLE`. Treating subsequent rounds as converged unless a new functional issue appears.

adnaan:
Round 4 — **converged.** This review identified **no new functional issue** (rounds 3 and 4 are both no-new-functional), CI is **12/12 green**, and the PR is **MERGEABLE / CLEAN**. Per the CLAUDE.md bot-convergence rule I'm not churning the remaining cosmetic/observability nits by push; they're addressed by reply below.

**Captured durably (pushed, docs-only `09fbdbe9`):**
- **`SubscribeToTopicActions` init-failure contract** — you correctly framed this as an architectural awareness note, not a defect: it's logged-only-and-continue, *identical to the pre-existing `SubscribeGroupActions`/`SubscribeServerActions` pattern at that exact `Handle()` site* (deliberately consistent; propagating it would alter the BroadcastAction init path, which is untouched through Phases 0–4). I recorded the operator contract — *"a logged init failure here ⇒ the instance's entire cross-instance topic-receive leg is dead"* — in `learnings/phase-2.md` "Open questions" and the `progress.md` Phase 6 row so the Phase 6 docs/troubleshooting pass surfaces it for operators. That's the right home for it (a Phase 6 deliverable, not a Phase 2 code change).

**Addressed by reply (no churn — all bot-marked "no action / minor / not a blocker / correct"):**
- **Double unmarshal:** you concurred "no action needed"; the round-1 decline holds (documented trade-off, BroadcastAction path untouched).
- **`goto` in the `measure` closure:** declining a refactor — the two forward `goto`s (`timed`, `run`) only skip past the warmup/drain blocks; they share the closure's `t`/`hits`/`publish` captures, so extracting a helper would fragment that for zero functional gain. golangci-lint (gate) accepts it.
- **`awaitWSContains` swallows non-timeout errors:** fair observability point, explicitly "not a correctness issue in happy-path testing." It's test-diagnostic polish on a converged PR; logging it here for the record rather than re-spinning CI for a non-functional nit (CLAUDE.md: address cosmetic nits by reply once converged). Reasonable Phase 3 cleanup when V17 adds delivery-count assertions to these same helpers.
- **`Publish` error-string inconsistency** (`"livetemplate: cannot Publish to an empty topic"` vs pubsub-layer `"topic cannot be empty"`): the root guard fires first so users only ever see the `livetemplate:`-prefixed message; the pubsub check is an unreachable-from-here backstop. "Fine as-is" per your note; not aligning strings on a converged PR.
- **`seq` no `omitempty`:** confirmed intentional and load-bearing (seq==0 ⇒ pre-upgrade sender — the Phase 3 ring relies on it); the struct comment + the new field comment both call this out so a Phase 3 cleanup pass won't add `omitempty`.

Thanks for the thorough multi-round review — the `seq=0` mixed-version-cluster catch in round 3 was genuinely valuable and is now a documented Phase 3 ring constraint. Treating the PR as converged: green, CLEAN, no outstanding functional items. Leaving the merge decision to the maintainer per the signoff gate.

adnaan:
Round 5 (`503caa2e`) — **converged** (rounds 3, 4, and 5 all surfaced no new functional issue; CI 12/12 green; PR MERGEABLE/CLEAN).

**Closed at the code site (pushed, comment-only):** the two items that recurred across rounds 4–5 *because the invariant lived only in `phase-2.md`*, not at the code — addressing them where the reviewer/maintainer actually looks is the move that closes the thread:
- **`template.go` `Handle()`** — added a comment: one `topicActionHandler` per broadcaster; a second `Handle()` sharing a `*RedisBroadcaster` log-errors here and *silently receives no cross-instance topic actions* for its lifetime (consistent with the `SubscribeGroupActions`/`SubscribeServerActions` pattern directly above; full operator contract in `phase-2.md` "Open questions" for Phase 6 docs). No code change — the pre-existing log-and-continue init pattern is deliberately preserved (BroadcastAction init path untouched, Phases 0–4).
- **`pubsub/types.go`** — inline note on the `Seq` json tag that the absent `omitempty` is intentional (`seq==0` ⇒ "pre-upgrade sender", not "absent"; the Phase 3 ring relies on it), so a future cleanup pass can't cargo-cult `omitempty`.

**Addressed by reply (no churn — bot-marked "fine as documented" / "minor" / "low blast radius"):**
- **`Publish` always returns nil on cross-instance failure:** already explicit in the `LiveHandler.Publish` interface doc — *"Returns an error only for an invalid topic/action argument …; a transport failure to a remote instance is logged, not returned (local fan-out still succeeds)."* The contract is documented at the signature; a caller reading the godoc cannot miss it. Returning relay errors would also diverge from the established `BroadcastAction` fan-out contract. Keeping as-is per your "fine as documented behavior."
- **`goto` in the latency-test `measure` closure:** reaffirming the round-4 decline — the two forward `goto`s only skip the warmup/drain blocks and share the closure's `t`/`hits`/`publish` captures; a helper would fragment that for no gain. Test-only, gate-green.
- **`canonicalActionKey` one-line wrapper:** keeping it — you noted "the naming is good (documents *why* `toSnakeCase` is called here)"; that self-documenting call site is the point, and it's the option you offered as acceptable.
- **Phase 3 constraints "should be mirrored in `phase-3.md` Adjustments":** they already are in the right place — `phase-3.md` doesn't exist until the Phase 3 session creates it from the skeleton; the Phase 3 audit-start protocol reads **`phase-2.md` "Adjustments recommended for the next phase"** (both `seq` ring constraints are there) **and** the seeded `progress.md` Phase 3 row (also has them). That's exactly the audit-start reader's input, and it persists in-repo post-merge.

Five rounds, every commit through the full `scripts/pre-commit.sh` gate (fmt + golangci-lint 0 issues + `go test ./...` incl. Redis testcontainers), zero CI failures throughout, PR CLEAN. No outstanding functional items. Leaving the merge to the maintainer per the signoff gate.

adnaan:
Acknowledged — and agreed this is the convergence terminus (your verdict: "no new functional issues… recommend merge"; CI 12/12; MERGEABLE/CLEAN).

**On the one new observation — stale `dispatchTypedMessage` doc comment:** valid catch, and as you note it's **pre-existing and cosmetic** (the comment claims `handleGroupActionMessage` uses the helper; it has its own unmarshal block — true before this PR). Declining to fix it *in this PR*, deliberately:
- It documents a helper on the **BroadcastAction / server-action dispatch path**, which the rollout's governing constraint keeps **untouched through Phases 0–4**. A doc-only edit there is low-risk, but it's still outside Phase 2's additive scope, and re-spinning the full CI + Redis suite + another review round for a pre-existing one-line comment nit at an explicit merge-recommend point is the exact churn the convergence rule warns against (CLAUDE.md: "address remaining cosmetic nits by reply rather than by push").
- The inconsistency is now *more* visible than before, not less: my new `handleTopicActionMessage` / `handleGroupActionMessage` comments explicitly describe the route-then-decode pattern and why those handlers don't use `dispatchTypedMessage`. A reader hitting the stale helper comment has the accurate explanation right next to it.
- Correct home: **Phase 5** (Removal + in-repo migration) is the phase that legitimately touches the BroadcastAction/server-action dispatch internals — a one-line `dispatchTypedMessage` comment correction belongs in that pass (or a trivial standalone nit PR). I'll note it for the Phase 5 audit.

**`LiveHandler.Publish` on the public interface** — agreed, expected and intended: it's the documented out-of-band entry point (use cases F/G/H); consumers via `Handle()` are unaffected; only a hand-rolled `LiveHandler` implementer adds a method, which is the normal cost of a new public capability.

Thanks for the five-round review — the `seq=0` mixed-version-cluster catch (round 3) materially de-risked Phase 3 and is now a documented ring constraint. Converged: 6 commits, every one through the full pre-commit gate, zero CI failures, no outstanding functional items. Merge decision left to the maintainer per the signoff gate.

Issue #420, opened 2026-05-18
Data race in pubsub TestReconnect_DoesNotBlockConcurrentPublish (reliably caught by CI -race ./...)


adnaan:
### Summary

`go test -race ./...` (the cross-repo "Test LVT against Core Changes" CI job) reliably reports a real data race:

```
WARNING: DATA RACE
testing.go:1712: race detected during execution of test
--- FAIL: TestReconnect_DoesNotBlockConcurrentPublish (~1.4s)
FAIL  github.com/livetemplate/livetemplate/pubsub
```

`TestReconnect_DoesNotBlockConcurrentPublish` lives at `pubsub/redis_test.go:1156` and exercises the reconnect-vs-concurrent-publish path of `RedisBroadcaster`.

### Pre-existing — not introduced by PR #419

- PR #419 (Phase 1 of the pub/sub topic model) changes **0** `pubsub/` files: `git diff --name-only main...broadcast-redesign-phase-1 -- pubsub/` is empty.
- The failing test was last modified by #389 / #394 / #398 (unrelated, by other authors).
- Reproduced **green 15× locally** under `-race` on clean `main` (`61a44c32`) and on the PR branch — it only manifests under CI's full-suite `-race ./...` contention (slower/contended runners widen the race window).
- Failed in CI on **two** independent PR #419 commits: run `26018626027` (`2a0968bf`) and run `26019615643` (`a7ad4731`). It is not an intermittent "flaky test" — it is a genuine race that CI's environment catches consistently while isolated local runs do not.

### Impact

This is a real unsynchronized-access bug in the `pubsub` reconnect path, not a CI artifact. It blocks `-race ./...`-gated CI for any PR (it surfaces regardless of the PR's content). It is recorded in `docs/proposals/broadcast-action-redesign-proposal/learnings/phase-1.md` and flagged on the Phase 2 tracker row — Phase 2 owns `pubsub` and is the natural place for the real fix, but it is not Phase-2-specific (it affects `main` today).

### Suggested handling

The real fix (synchronize the reconnect-vs-publish access) is `pubsub`-owned and out of scope for the Phase 1 feature PR (#419), which deliberately changes no `pubsub` code. Options for maintainers, in priority order:

1. Fix the race in `pubsub` directly (separate PR).
[...]

PR #426, opened 2026-05-19
feat(topics): Phase 3 — multi-segment wildcards (PSUBSCRIBE + seen-ring dedup)

adnaan:
Phase 3 of the Publish/Subscribe topic-model redesign (proposal #415). Cross-instance multi-segment wildcards on Phase 2's exact-`SUBSCRIBE` seam. Gated by **V17** + **V18**. `BroadcastAction` and its group-dispatch path remain **untouched** (additive through Phases 0–4).

## What shipped

- **`pubsub/redis.go`** — `subscribedPatterns` map parallel to `subscribedChannels`. The race-critical retry / check-lock-check loop is **single-sourced** via a `subKind` param (`subExact`/`subPattern`) selecting the refcount map + Redis verb; the 10 exact callers pass `subExact` (group-action wire byte-for-byte unchanged). `SubscribeToTopicPattern` / `UnsubscribeFromTopicPattern` issue `PSUBSCRIBE`/`PUNSUBSCRIBE` on the **same `b.pubsub`** the single `processMessages` pump reads. `reconnect()` replays `subscribedPatterns` via `newPubSub.PSubscribe` on the new PubSub.
- **`pubsub/redis.go`** — lock-free `(instanceID, seq)` seen-ring (bounded N=64) folded into `handleTopicActionMessage` before registry resolution, on the single serialized pump. `seq==0` bypasses the ring **entirely** (process unconditionally **and** do not record — the pre-upgrade-sender two-halves discipline).
- **`pubsub/types.go`** — `TopicPatternSubscriber` as a **separate** optional interface (no backward-incompatible widening of `TopicChannelSubscriber`).
- **`topics.go`** — root-package `isPatternTopic` (cross-boundary twin of `session.isPatternTopic`).
- **`topic_runtime.go`** — relay branches on `isPatternTopic`: a pattern relays as exactly **one `PSUBSCRIBE`**, never expanded into per-concrete `SUBSCRIBE`s (the relay invariant). Centralized so the per-action relay and the disconnect sweep share one source; fixes a latent pattern-mis-release in `releaseRelayedTopics`.
- **`topic_context.go` + `mount.go`** — `ctx.Publish` / `handler.Publish` reject a wildcard pattern with a clean error (previously a `GetByTopicExcept` panic) — patterns are Subscribe-only.

Local strict re-match needed **no code**: `GetByTopicExcept` + `segmentMatch` already drop Redis-`*`-spans-`/` over-delivery via the equal-segment-count rule (audit-confirmed; phase-2.md predicted this). Multi-segment wildcards **HOLD in v1** — no reduction.

## Tests / gate

- **V17** single-instance (deterministic): trailing/multi/leading patterns + segment-count valve + exact∪pattern exactly-one dedup. **V17** cross-instance (Redis testcontainers): PSUBSCRIBE delivery, over-delivery rejection, SUBSCRIBE+PSUBSCRIBE double-fire exactly-once. **V18**: ACL receives the literal pattern; first-ever concrete delivered with ACL not re-invoked at Publish.
- Pubsub units: `TestSeenRing`, `TestHandleTopicActionMessage_DedupAndSeqZeroBypass` (the `seq==0` two-halves discipline), `TestRedisBroadcaster_ReconnectPreservesPatternSubscriptions` (the `subscribedPatterns` PSubscribe-replay seam). `TestIsPatternTopic` / `TestSegmentMatch` regression-green.
- Full `scripts/pre-commit.sh` GREEN (go fmt + `golangci-lint --enable-only=errcheck,govet,ineffassign,staticcheck,unparam,unused` 0 issues + full `go test ./...` incl. Redis testcontainers). `-race` scoped green: `pubsub` 25.9s, `internal/session` 32.3s, root `TestTopic|TestSegmentMatch|TestIsPatternTopic` 11.2s. Cross-instance 3×-repeat, no flake.

### Benchmark (§"Benchmarks required" item 2 — wildcard pattern-scan; items 1 & 3 were Phase 2)

Worst case (concrete matches none of P distinct patterns ⇒ all P scanned):

| P | ns/op | B/op | allocs/op |
|---|---|---|---|
| 1 | 836 | 256 | 6 |
| 10 | 3 407 | 1000 | 24 |
| 100 | 22 224 | 8193 | 204 |

Clean linear **O(P)** (~220 ns/pattern) — the flat `segmentMatch` scan is adequate; **no trie/radix index** is warranted (proposal §2 / Appendix B — this validates the linear scan, not whether to add an index). A future zero-allocation optimization (pre-split patterns at Subscribe time, no trie) is logged in `phase-3.md` as parked future-work.

## Learnings

`docs/proposals/broadcast-action-redesign-proposal/learnings/phase-3.md` (full Audit + deviations + adjustments) and `progress.md` (Phase 3 → complete, Surfaced-Scope Ledger rolled, the multi-segment-holds-in-v1 decision journaled, Phase 4/5 seams seeded).


adnaan:
Thanks for the review. Addressed in `3e86a537`:

1. **Duplicate `## File / commit / V-item pointers` in `phase-3.md`** — removed the trailing empty leftover skeleton header (real content stays under the earlier occurrence).
2. **`trySubscribe` refcounts re-read** — dropped the no-op `refcounts = b.refcounts(kind)` re-assignment (Go map is a reference type; the local already points at the live map). Comment now states the lock-held guarantee comes from `b.mu.Lock()`, not a re-fetch — your exact point.
3. **Rot-prone phase/PR tokens** — trimmed `"(Phase 3 wildcards)"` (×2) and `"#420/#422"` (×2) from code comments per CLAUDE.md (phase/PR refs drift).
4. **`awaitWSContainsRejecting`** — added the note that the retry is a SUBSCRIBE/PSUBSCRIBE-propagation guard, not a load driver (normally one iteration).

**Deliberately kept (declining the broader comment-density trim):** the invariant-explaining comments — the `seq` two-ring-constraint doc, the relay-invariant / single-PubSub-instance / seen-ring rationale, and the proposal-section citations that pin *why* (e.g. `N=64`, the `seq==0` two-halves discipline). These are load-bearing for a distributed-systems dedup/reconnect seam and you flagged them as worth keeping; the proposal-section names are stable, not drift-prone like phase/PR numbers. Per CLAUDE.md this is the documented examples-vs-core distinction inverted: here density is defensible because the seam is subtle. The remaining inline phase references (if any) are a deliberate Phase 6 docs-pass item, logged in `phase-3.md`.

adnaan:
Round 2 addressed in `7149f163`. Functional correctness was declared sound; this round was style + minors:

**Comment blocks** — distilled every block you enumerated to one tight line (interface: two): the `seenRing` field/const/type/method cluster, the `ctx.Publish`/`handler.Publish` pattern guards, `TopicPatternSubscriber`, `SubscribeToTopicPattern`, the `seen_ring_test.go` header, and `relayTopicSubscribeOne`. Net −124/+37 lines of comment. The full reasoning is preserved verbatim in `learnings/phase-3.md`, which is the durable record.

**Minors** — all done:
- `reconnect()`: pinned the `go-redis v9.17.2` assumption next to the no-second-`Receive` `PSubscribe` so a version bump prompts re-verification.
- `TestHandleTopicActionMessage_DedupAndSeqZeroBypass`: `atomic.Int64` → plain `int` (synchronous, no concurrency) + dropped the unused `sync/atomic` import.
- `awaitWSContainsRejecting`: 150ms → 400ms per-read deadline to narrow the stale-frame window.

**Confirm-only items — confirmed:**
- *`seenThenRecord` eviction edge:* yes, an id at the `r.next` slot is found and returns `true` without advancing, so it can outlive the nominal 64-window if it keeps matching. Understood and acceptable: the only producer of a repeat key is the SUBSCRIBE+PSUBSCRIBE double-fire whose two copies arrive back-to-back on the one pump, so the window never matters in practice (pinned by `TestSeenRing` wrap-around).
- *`reconnect()` no second `Receive`:* confirmed correct for go-redis v9 (the `.Channel()` pump drains psubscribe confirmations); now version-pinned per your suggestion.

**Declining the blanket "every multi-line block → one line":** the remaining multi-line comments in the diff are (a) the load-bearing invariant docs you yourself flagged as the warranted-WHY case (relay invariant, `seq` two-ring-constraint, single-PubSub), and (b) **pre-existing Phase 0/1/2 blocks in these same files that I did not introduce** (`reconnect()`'s 5-step flow, `Subscribe`'s validation-order doc, the `GroupActionMessage` envelope doc, etc.). Per CLAUDE.md "Follow existing patterns — check neighboring code for conventions": the merged neighboring code in `pubsub/redis.go` / `topic_context.go` / `mount.go` uses this exact block style and passed prior review; reformatting it is out of this PR's scope and would make the diff unreviewable. The AI-Code-Review-Workflow section's own note ("comment density is higher than CLAUDE.md defaults but defensible") and your Round 1 ("defensible for a distributed-systems dedup/reconnect seam") apply. The new code now matches — or is tighter than — that established convention.

adnaan:
Round 3 addressed in `e2fcccd2`. Good catch — the crash-prevention guard genuinely had only indirect coverage.

Added `TestTopic_PublishToWildcardPatternRejected` covering **both** boundaries with the suite's capture-the-real-error convention (more robust than asserting a WS error-frame shape):
- **`ctx.Publish`** via an action — `room/*`, `*/alice`, `org/*/room/*`, `*` all return the `Subscribe-only` error; plus a concrete-topic (`room/42`) still-accepted control so the guard is provably pattern-only, not a regression.
- **`handler.Publish`** (the `LiveHandler` path) — pattern rejected, concrete (`org/42`) accepted, called directly (no Context, no Redis).

Full pre-commit gate green (lint 0 issues + full `go test ./...` incl. Redis). This was the one remaining functional item; the rest of Round 3 confirmed the functional surface correct.

adnaan:
Round 4: no new functional issues — everything re-confirmed correct and the Round-3 guard test accepted ("coverage is strong"). Convergence reached (R3 was the last substantive item; R4 is a clean re-affirmation).

The two "areas to watch" are explicitly non-blocking and already mitigated — addressing by reply rather than another push (per CLAUDE.md, pushing optional polish only re-triggers the review loop):

1. **`reconnect()` go-redis version dependency** — already guarded: the in-code comment pins `v9.17.2` and says re-verify on a go-redis bump (your Round-2 ask), and it's parked as explicit future-work in `phase-3.md`. The bot itself notes the code comment "is a sufficient guard for now." A `go.mod` inline comment is optional polish; declining to push it solo to avoid an extra full review/CI cycle for a non-blocking nit.
2. **`awaitWSContainsRejecting` flakiness surface** — acknowledged in-code (the propagation-guard comment + 400ms deadline) and empirically clean (cross-instance suite 3×-repeat, no flake). No actionable defect; noted as a future surface, not a regression.

No code change this round. Holding here for CI to finish; not merging (prcommentsfix scope only — awaiting maintainer signoff).

PR #433, opened 2026-05-22
fix(#413): cache range-item statics across full content swaps

### LINKED ISSUE #413: {{range}} statics not cached across full content swap — resends per-row scaffolding
## Summary

livetemplate's statics/dynamics split correctly caches static HTML for **in-place updates** (selecting an item, toggling a class — tiny payloads, statics omitted). But when a `{{range}}`'s backing slice is **fully replaced** (every item differs), the framework resends the per-row *static scaffolding* for all N rows rather than reusing the cached range-item template and sending only the new per-row *dynamics*.

For large lists this dominates the wire payload and is the bottleneck on slow/mobile links.

## Reproduction

The `prereview` app (github.com/livetemplate/prereview) renders a file's diff as:

```html
<div class="code">
  {{range .Lines}}
  <div class="line-row">
    <button class="line {{.Kind}}" data-key="L{{.OldNum}}-{{.NewNum}}">
      <span class="gutter">…{{$ln}}</span>
      <span class="content">{{.HighlightedContent}}</span>
    </button>
  </div>
  {{end}}
</div>
```

Switching the selected file replaces `.Lines` entirely (different file, different count). The per-row **static structure** is identical for every row and every file — only the dynamics (`.Kind`, line number, content) differ.

## Evidence

Captured WebSocket frames (post-decompression payload) on a file switch:

| Switch | Payload | `"s":[` (statics) present? |
|---|---|---|
| → controller.go (757 rows) | **434 KB** | yes |
| → csv/reader.go (57 rows) | 61 KB | yes |

Payload scales linearly with row count and the per-frame data contains the statics key. If the range-item statics were cached, switching to a 757-row file should send ~the dynamic line text only (tens of KB), not 434 KB of repeated `<div class="line-row"><button…` scaffolding.

Also tested: presence/absence of a path-scoped `data-key` on the container made no difference to the statics payload (434322 B vs 434628 B) — so this is inherent to the range content-swap path, independent of client-side morphdom keying.

## Expected

On a full `{{range}}` content swap where the item template is unchanged, the server should send only the per-row dynamics (and add/remove ops for count delta), reusing the already-cached range-item statics — same as it does for in-place single-item updates today.

## Impact

- Bandwidth: O(rows × static-scaffold-size) per navigation on any range-backed list (diff viewers, tables, log views).
- Most visible on mobile / high-latency links where transfer time, not render, is the bottleneck.

## Current mitigations (app side, not a real fix)

- `WithWebSocketCompression()` — permessage-deflate gets ~10–20× on the repetitive scaffolding. Strongly recommended regardless, but it's compressing redundant data.
- Path-scoped `data-key` on the container to force a clean subtree replace (avoids a separate per-row-key collision issue; doesn't change the statics resend).

The real fix is framework-side: cache range-item statics across content swaps.

🤖 Filed from prereview perf investigation.

adnaan:
Fixes #413.

## Summary

The reported symptom was "per-row static scaffolding re-emitted on `{{range}}` full content swaps." Investigation found:

1. **For the simple template shape in the issue body, per-row statics are already correctly stripped** by the existing fingerprint-based op-envelope path. Verified across full content swap, grow, shrink, single-item swap, empty↔non-empty, and repeated A↔B swaps. The 434KB / 61KB payloads cited in the issue are dominated by per-row *dynamic* content (highlighted source), not redundant statics.

2. **The real bug is narrower**: `PrepareTreeForClient` was dropping `AutoKey` (the `_k` field) when stripping statics. Auto-keyed ranges (slices without an explicit `data-key`) hit this on stream-mode inserts — the new row arrived without `_k`, so the client couldn't track it for later remove/update operations. This was the root cause of the downstream regressions in `examples/flash-messages` (remove-by-id failed) and tinkerdown auto-tables (new rows never appeared).

## The fix

One-line preservation of `AutoKey` in `PrepareTreeForClient`'s strip path:

```go
case *TreeNode:
    result := &TreeNode{AutoKey: v.AutoKey}
```

`Metadata` is preserved by existing code further down the same function.

## What this does NOT do

An earlier attempt at this PR threaded `clientHasItemStatics` through the insertion helpers to also strip nested item statics on stream-mode inserts (mirroring the update path). That was **reverted**: unlike updates (which patch existing DOM in place), inserts render new DOM and the client has no per-position branch-statics cache. Stripping nested statics on insert lost the HTML wrapper for any item containing a nested `{{if}}` with dynamics. The reverted approach is now locked out by regression tests; do not re-introduce it without solving the client-side cache problem.

## Tests added

- `range_statics_swap_test.go` (11 tests, public API via `ExecuteUpdates`) — locks in the no-per-row-statics contract for the simple-template case and the must-carry-statics contract for auto-key + nested-branch inserts.
- `internal/diff/range_insert_strip_test.go` (3 tests) — exercises the dispatch directly:
  - `TestStreamInsert_PreservesNestedDynamicBranchStatics` — locks out the reverted approach
  - `TestStreamInsert_PreservesStaticOnlyBranchIdentity` — non-regression for the `{s:["+"]}` / `{s:["-"]}` special case
  - `TestStreamInsert_PreservesAutoKey` — direct regression test for the landed fix

## Benchmarks

`range_statics_swap_bench_test.go` measures wire-payload bytes per full content swap across N = 10 / 100 / 1000 for two shapes:
- **Simple**: no nested conditionals — bytes/op scales with per-row dynamic content size (zero per-row statics).
- **DynamicBranch**: nested `{{if}}` with a dynamic. Branch statics MUST stay in the wire so the client can render new rows — bytes/op reflects that cost (intentional, not a regression to optimize away).

Run with `go test -bench=BenchmarkRangeFullSwap -benchtime=20x -run=^$ .`

## Risks & compatibility

- **No wire-format change.** `PrepareTreeForClient`'s output now includes `_k` for auto-keyed items on the strip path — clients already handle this field on the no-strip path.
- **No protocol additions, no schema change.**
- **Insertion path unchanged.** Nested statics still flow through on stream-mode inserts.

## Verification

- `go test ./... -timeout 300s`: all packages pass
- `go vet ./...`: clean
- Pre-commit hook (`scripts/pre-commit.sh`): passes

---
_Generated by [Claude Code](https://claude.ai/code/session_014nUsSxrv8yH7tYWRnmaqnY)_

### DIFF
--- a/internal/diff/prepare.go
+++ b/internal/diff/prepare.go
@@ -24,8 +24,10 @@ func PrepareTreeForClient(node interface{}, clientHasStatics bool) interface{} {
 	// Client has statics - remove them to reduce wire size
 	switch v := node.(type) {
 	case *TreeNode:
-		// Create new TreeNode without statics or fingerprint
-		result := &TreeNode{}
+		// Create new TreeNode without statics or fingerprint. AutoKey must
+		// be preserved because it is the "_k" identifier the client uses to
+		// track range items by key (issue #413).
+		result := &TreeNode{AutoKey: v.AutoKey}
 		// Pre-allocate to avoid repeated growth in SetDynamic
 		result.GrowDynamics(len(v.Dynamics))
 		// Recursively prepare dynamics, filtering out empty values while preserving static-only conditional blocks
--- a/internal/diff/prepare.go
+++ b/internal/diff/prepare.go
@@ -24,9 +24,7 @@ func PrepareTreeForClient(node interface{}, clientHasStatics bool) interface{} {
 	// Client has statics - remove them to reduce wire size
 	switch v := node.(type) {
 	case *TreeNode:
-		// Create new TreeNode without statics or fingerprint. AutoKey must
-		// be preserved because it is the "_k" identifier the client uses to
-		// track range items by key (issue #413).
+		// AutoKey is the "_k" identifier the client uses to track range items by key.
 		result := &TreeNode{AutoKey: v.AutoKey}
 		// Pre-allocate to avoid repeated growth in SetDynamic
 		result.GrowDynamics(len(v.Dynamics))
--- a/internal/diff/range_insert_strip_test.go
+++ b/internal/diff/range_insert_strip_test.go
@@ -1,0 +1,175 @@
+package diff
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/livetemplate/livetemplate/internal/build"
+)
+
+// TestStreamInsert_StripsNestedDynamicBranchStatics — issue #413.
+//
+// The stream-mode update path (`dynamicsToUpdatePayload`) already strips
+// nested statics via `PrepareTreeForClient(v, true)` because the §5b
+// homogeneity check guarantees the client has them. Insertions were
+// inconsistent: they kept all nested statics via `PrepareTreeForClient(item,
+// false)`, re-emitting per-row scaffolding the client had cached.
+//
+// After the fix in #413, stream-mode insertions also pass the
+// homogeneity-guaranteed flag, so nested branches containing dynamics drop
+// their statics on the wire. Static-only branches (whose entire content is
+// HTML, no dynamics) still retain statics — the special case in
+// `PrepareTreeForClient` preserves branch identity for those, since the
+// client otherwise has no way to distinguish two static-only branches at the
+// same item position.
+func TestStreamInsert_StripsNestedDynamicBranchStatics(t *testing.T) {
+	// Build a homogeneous range "old side" with items shaped like:
+	//   {data-key, marker-branch-with-dynamic}
+	// where the marker is `{{if .HasMarker}}<m>{{.MarkerText}}</m>{{end}}`.
+	// All items share the same statics-fingerprint shape, so
+	// TransitionToStreamMode succeeds.
+	itemStatics := []string{`<li data-key="`, `">`, `</li>`}
+	makeItem := func(key, markerText string) *TreeNode {
+		marker := &TreeNode{
+			Statics:  []string{"<m>", "</m>"},
+			Dynamics: []interface{}{markerText},
+		}
+		return &TreeNode{
+			Statics:  itemStatics, // before extractItemDynamics, item carries statics
+			Dynamics: []interface{}{key, marker},
+		}
+	}
+
+	oldItems := []interface{}{
+		extractDynamics(makeItem("k1", "old-marker-1")),
+		extractDynamics(makeItem("k2", "old-marker-2")),
+	}
+	oldTree := &TreeNode{
+		Statics: itemStatics,
+		Range:   &build.RangeData{Items: oldItems, Statics: itemStatics},
+	}
+	TransitionToStreamMode(oldTree)
+	if oldTree.Range.StreamState == nil {
+		t.Fatalf("TransitionToStreamMode left StreamState nil — fixture lost homogeneity?")
+	}
+
+	// New side: all items replaced with new keys (full content swap).
+	// Items have the SAME structure (marker with dynamic), but different
+	// dynamic values.
+	newItems := []interface{}{
+		extractDynamics(makeItem("n1", "new-marker-1")),
+		extractDynamics(makeItem("n2", "new-marker-2")),
+	}
+	newTree := &TreeNode{
+		Statics: itemStatics,
+		Range:   &build.RangeData{Items: newItems, Statics: itemStatics},
+	}
+
+	// Drive the dispatch path so stripStatics == true (fingerprint match).
+	changes := &TreeNode{}
+	if !handleStreamModeRange(oldTree, newTree, changes) {
+		t.Fatalf("handleStreamModeRange did not dispatch to stream mode")
+	}
+
+	if changes.Range == nil {
+		t.Fatalf("expected changes.Range to be populated; got: %+v", changes)
+	}
+
+	payload, err := json.Marshal(changes.Range.Items)
+	if err != nil {
+		t.Fatalf("json.Marshal failed: %v", err)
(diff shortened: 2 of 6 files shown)


adnaan:
## CI green ✅ — downstream regressions fixed

All 12 checks pass on the latest push ([commit e3f0df44](https://github.com/livetemplate/livetemplate/pull/433/commits/e3f0df44), [workflow run](https://github.com/livetemplate/livetemplate/actions/runs/26304679668)), including the two suites that were failing on the previous push:

- ✅ Test Examples against Core Changes (was failing on `TestFlashMessagesE2E/RemoveItemWorks`)
- ✅ Test Tinkerdown against Core Changes (was failing on `TestAutoTables_SQLiteAdd`)
- ✅ Test LVT, Unit Tests, Fuzz Tests, benchmark, claude-review, all 5 docs suites

### What was wrong

The previous commit (b4a8c7af) symmetrized the stream-mode INSERT path with the UPDATE path by stripping nested per-item statics under the §5b homogeneity guarantee. **The symmetry assumption was wrong.** Updates patch existing DOM in place; inserts have to *render* new DOM from data, and the client has no per-position branch-statics cache. When the client received an inserted item shaped like `{"0": {"0": "marker text"}}` with no `"s"` on the nested object, its `renderValue` path joined the numeric-key values with empty string and emitted just the dynamic text — no HTML wrapper.

Two regressions came from that:

| Test | Why it broke |
| --- | --- |
| `examples/flash-messages` `RemoveItemWorks` | `PrepareTreeForClient` was also dropping `AutoKey` when stripping. Items in a `{{range}}` over `[]string` (no explicit `data-key`) get an auto-generated `_k` hash. Without `_k` on the appended row, the client couldn't track the new item — and the subsequent "click the last row's Remove" couldn't address it. |
| `tinkerdown` `TestAutoTables_SQLiteAdd` | Template is `{{range .Data}}{{if eq .Id $.EditingId}}<tr edit>{{else}}<tr view>{{end}}{{end}}`. Each item carries a nested `*TreeNode` (the if/else branch) with per-row `<tr><td>…</td></tr>` statics. Stripping those on append produced items like `{"0":{"0":"Widget","1":"9.99"}}` — the client rendered "Widget9.99" with no row structure, so the new `<tr>` never appeared in the table. |

### What changed

- **Reverted the parameter-threading in `generateInsertionOps`** and its helpers. The insert path always passes `clientHasItemStatics=false` to `PrepareTreeForClient`, so per-item nested statics ship intact. Op-envelope statics (`["a"/"p"/"i", items, statics]` third element) still get stripped by `stripStaticsFromOperations` — that strip was correct and is unchanged.
- **Defensively preserved `AutoKey` in `PrepareTreeForClient`'s `*TreeNode` branch.** The strip path now copies `v.AutoKey` to the result. Future callers can't accidentally drop the `_k` identifier.

The simple-range case from the original issue (homogeneous range with primitive dynamics — the literal repro in #413) is unaffected: those items have no nested `*TreeNode`s to strip, so the existing op-envelope strip still drives bytes/op down. The 9 `TestRangeStatics_*` swap-shape tests from b4a8c7af continue to pass without modification.

### Tests

Augmented (not deleted) the regression tests added in b4a8c7af:

- `TestStreamInsert_PreservesNestedDynamicBranchStatics` — renamed from `…StripsNestedDynamicBranchStatics`; assertion inverted to lock in the correct contract (nested branch statics MUST remain on the wire so the client can render new rows).
- `TestStreamInsert_PreservesAutoKey` (new) — exercises the flash-messages flow: `{{range}}` over `[]string`, auto-keys, stream append. Asserts `_k` is present on the new item.
- `TestStreamInsert_PreservesStaticOnlyBranchIdentity` — unchanged; the `PrepareTreeForClient` special case for static-only branches was always correct.
- `TestRangeStatics_AutoKey_AppendIncludesAutoKey` and `TestRangeStatics_NestedBranch_AppendIncludesBranchStatics` (new) — public-API regression tests mirroring the failing downstream cases.

### Caveats

- The wire payload reduction the original PR description advertised for the "DynamicBranch" benchmark shape (-33% across N=10/100/1000) is no longer claimed. Those savings only materialized by stripping branch statics the client needed — they weren't real wins, they broke rendering. The benchmark file is kept (with updated comments) so any future stripping work can measure against it.
- The simple-range payload reduction (the actual case from the issue body) is unchanged and still in effect. `BenchmarkRangeFullSwap_Simple_N1000` reports the same ~114 KB/op as before.
- PR stays draft per the task brief; not marking ready-for-review.

---
_Generated by [Claude Code](https://claude.ai/code)_

adnaan:
Thanks for the review — calling convergence on this round per our [bot-review-loop convergence rule](https://github.com/livetemplate/livetemplate/blob/main/CLAUDE.md#ai-code-review-workflow): "successive rounds aren't identifying any new functional issue, only style/wording alternatives on content that was already accepted." Both nits in this review are explicitly marked optional/non-blocking and are cosmetic. Addressing them by reply rather than by push to avoid the rephrase-forever loop.

**On the `AutoKey` comment wording**: keeping the current `// AutoKey is the "_k" identifier the client uses to track range items by key.` — it's already a one-line WHY, and the alternative phrasing is a marginal improvement, not a fix. Both communicate the same constraint.

**On the `TestRangeStatics_EmptyToNonEmpty` count==1 comment**: declining for now. The test's name (`_IncludesItemTemplateOnce`) already telegraphs the assertion's intent. Adding the same kind of "statics live at the op envelope" explainer comment on every range-statics test would inflate the suite without adding new information — `TestRangeStatics_GrowCountDelta` already carries that explanation as a representative anchor for the test family. Happy to revisit if a future contributor actually hits the confusion.

The fix itself (one-line `AutoKey` preservation in `PrepareTreeForClient`) and the test/benchmark coverage stand as approved in this review. CI is green (12/12), PR is mergeable. Ready for human review/merge.

PR #435, opened 2026-05-23
fix(build): tokenizer-based body/script detection (closes #414)

### LINKED ISSUE #414: Wrapper injection mis-targets when <head> contains literal "<body"/"<script" text — silently breaks all lvt-on:* delegation
## Summary

`injectWrapperDivStringBased` in `internal/build/wrapper.go` locates the
document body with naive substring scans:

- `bodyStart := strings.Index(htmlDoc, "<body")`  (≈ line 118)
- `bodyEnd := strings.LastIndex(htmlDoc, "</body>")`  (≈ line 132)
- `scriptStart := strings.Index(bodyContent, "<script")`  (≈ line 141)

These match the **first/last literal occurrence anywhere in the document
string**, including inside `<head>` — in CSS, inline JS, `<meta content="…">`,
or **HTML/CSS comments**. If any such text appears before the real
`<body>` tag (e.g. a CSS comment that mentions `<body>`), the
`<div data-lvt-id>` wrapper is injected at the wrong byte offset.

The browser then receives markup where the page content is **not** inside
`[data-lvt-id]`. Client-side event delegation's `inWrapper` guard walks
ancestors for `[data-lvt-id]` and finds none, so **every `lvt-on:*`
action is silently rejected**. Tier-1 form submits still work, which
makes the failure look element-specific/local when it is global — a very
expensive misdiagnosis trap.

This path is taken whenever the template contains `{{`/`}}` (the common
case for full-HTML templates), so it is not an edge path.

## Reproduction

A full-HTML livetemplate template whose `<head>` `<style>` contains a
comment like:

```css
/* the wrapper sits between <body> and our layout … */
```

`strings.Index(htmlDoc, "<body")` matches the comment text in `<head>`,
not the real `<body>` element. All `lvt-on:click` etc. stop firing; no
error is logged.

## Impact

- Silent, total loss of `lvt-on:*` delegation.
- No diagnostic; only Tier-1 (form-submit) actions keep working, which
  misleads debugging toward "this one element/handler is broken".

Fast diagnostic for anyone who hits this:
`document.querySelector('[lvt-on\\:click]').closest('[data-lvt-id]') === null`
(true ⇒ wrapper was mis-injected).

## Suggested fix

Resolve `<head>` first and start the `<body>`/`<script>` scans **after
`</head>`** (or parse only the post-head region). Similarly bound the
`<script>` split to body content that actually follows the real `<body>`
tag. The HTML-parsing path (`html.Parse`) is already correct; only the
string-based fallback used for templates containing `{{ }}` has this bug.

## Found via

Diagnosed 2026-05-15 in the downstream `prereview` app. Worked around
there template-side by keeping the literal substrings out of `<head>`,
but the framework should not be substring-fragile.

adnaan:
## Summary

- Closes #414. `injectWrapperDivStringBased` and `ExtractTemplateBodyContent` used naive `strings.Index` to locate `<body>`, `</body>`, and `<script>` — matching any literal substring including those inside `<head>` CSS comments, inline JS, meta `content=`, HTML comments, or `<title>` RCDATA. The wrapper `<div data-lvt-id>` got injected at the wrong byte offset → page content not inside `[data-lvt-id]` → client-side delegation's `inWrapper` ancestor check silently rejected every `lvt-on:*` handler while Tier-1 form submits still worked (a misleading partial failure that cost a multi-hour misdiagnosis when first encountered downstream).
- Replaced with a single `html.Tokenizer` walk that correctly skips matches inside comments, RAWTEXT (`<script>`/`<style>`), RCDATA (`<title>`/`<textarea>`), and attribute values. Tolerant of `{{...}}` directives in text + attribute-value positions (the common full-HTML case that `html.Parse` would mangle, which is exactly why the string-fallback path exists).
- Both call sites share one helper (`locateBodyAndFirstScript`) so a future substring-based regression in either spot would have to actively bypass the helper.

## Performance

`BenchmarkWrapperInjection/string-fallback-tokenizer` is **34% faster (7.8µs vs 11.8µs) and uses 62% fewer allocs** than the html.Parse path on the same input — so this is a net perf improvement on templates that contain `{{...}}`.

## Test plan

- [x] Unit: 11 tricky-head fixtures mirrored across `InjectWrapperDiv` and `ExtractTemplateBodyContent` (CSS comment, inline JS string with `<body`, meta content, HTML comment, title text, multi-decoy combinations)
- [x] Property fuzz: 1000-iter decoy permutation pinning the regression
- [x] Offset-coverage sanity: `sum(len(z.Raw())) == len(input)` over a fixture corpus
- [x] Behavior pins: nested-body `strings.LastIndex` semantic preserved; safe template-directive attribute-value case; documented limitations (`{{/* <body> */}}` comment and `{{if}}` mid-tag) pinned with deliberate-change-only comments
- [x] `go test -race ./internal/build/...` PASS
- [x] `go test ./...` at 300s timeout PASS (no golden-file drift)
- [x] `scripts/pre-commit.sh` PASS (gofmt + golangci-lint with project linter set + full test suite)
- [x] Local two-way E2E: chromedp test passes against this branch via local replace, FAILS against v0.10.1 (page doesn't even bootstrap — wrapper mis-positioned so badly that the client.js script tag is mangled). The chromedp test itself will land in lvt as a follow-up PR after this releases.

## Out of scope

`internal/build/html_segmentation.go` has a related naive `strings.Index` scan for block-level structural tags. Different bug class (segmentation/diff heuristic, not event delegation) and would risk golden-file churn if bundled — filed separately as a follow-up.

### DIFF
--- a/internal/build/build_bench_test.go
+++ b/internal/build/build_bench_test.go
@@ -89,6 +89,23 @@ func BenchmarkWrapperInjection(b *testing.B) {
 
 	fragment := `<div><h1>Hello</h1><p>World</p></div>`
 
+	// String-fallback path (triggered by {{...}} directives) — exercises
+	// the html.Tokenizer-based locateBodyAndFirstScript helper added for
+	// livetemplate#414. Compared against the cleaner full-html path to
+	// guard against regressions from the previous naive strings.Index.
+	stringFallbackHTML := `<!DOCTYPE html>
+<html>
+<head>
+<title>Test</title>
+<style>/* layout sits between <body> and footer */</style>
+<meta property="og:description" content="contains <body literally">
+</head>
+<body>
+<div><h1>{{.Title}}</h1><p>{{.Body}}</p></div>
+<script>console.log("loaded");</script>
+</body>
+</html>`
+
 	b.Run("full-html", func(b *testing.B) {
 		b.ReportAllocs()
 		for i := 0; i < b.N; i++ {
@@ -100,6 +117,13 @@ func BenchmarkWrapperInjection(b *testing.B) {
 		b.ReportAllocs()
 		for i := 0; i < b.N; i++ {
 			_ = InjectWrapperDiv(fragment, "test-wrapper", false)
+		}
+	})
+
+	b.Run("string-fallback-tokenizer", func(b *testing.B) {
+		b.ReportAllocs()
+		for i := 0; i < b.N; i++ {
+			_ = InjectWrapperDiv(stringFallbackHTML, "test-wrapper", false)
 		}
 	})
 }
--- a/internal/build/build_bench_test.go
+++ b/internal/build/build_bench_test.go
@@ -89,10 +89,6 @@ func BenchmarkWrapperInjection(b *testing.B) {
 
 	fragment := `<div><h1>Hello</h1><p>World</p></div>`
 
-	// String-fallback path (triggered by {{...}} directives) — exercises
-	// the html.Tokenizer-based locateBodyAndFirstScript helper added for
-	// livetemplate#414. Compared against the cleaner full-html path to
-	// guard against regressions from the previous naive strings.Index.
 	stringFallbackHTML := `<!DOCTYPE html>
 <html>
 <head>
--- a/internal/build/wrapper.go
+++ b/internal/build/wrapper.go
@@ -116,7 +116,7 @@ func FindBodyNode(n *html.Node) *html.Node {
 // document body and the first <script> inside it. Unlike strings.Index, the
 // tokenizer correctly skips '<body' / '<script' substrings appearing in HTML
 // comments, RAWTEXT (<script>/<style> content), RCDATA (<title>/<textarea>
-// content), and attribute values — the bug behind livetemplate#414.
+// content), and attribute values.
 //
 // Returns four byte offsets into htmlDoc, or -1 when not found:
 //   - bodyOpenStart:    '<' of the FIRST <body...> start tag
@@ -174,7 +174,7 @@ func locateBodyAndFirstScript(htmlDoc string) (bodyOpenStart, bodyOpenEnd, bodyC
 // htmlDoc contains Go template directives (which would be mangled by
 // html.Parse + html.Render). It uses html.Tokenizer for tag-boundary
 // detection so that '<body' / '<script' substrings inside head content
-// don't fool it (livetemplate#414).
+// don't fool it.
 func injectWrapperDivStringBased(htmlDoc string, wrapperID string, loadingDisabled bool) string {
 	bodyOpenStart, bodyOpenEnd, bodyCloseStart, scriptStart := locateBodyAndFirstScript(htmlDoc)
 	if bodyOpenStart < 0 || bodyCloseStart < 0 || bodyOpenEnd > bodyCloseStart {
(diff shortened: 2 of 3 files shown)


adnaan:
Thanks for the review — addressed in 2nd push:

**Item #2 (issue number in names) — accepted.** Renamed:
- `trickyHeadIssue414Cases` → `trickyHeadDecoyCases`
- 4 `TestInjectWrapperDiv_Issue414_*` / `TestExtractTemplateBodyContent_Issue414_*` → dropped the `Issue414_` infix
- Stripped `livetemplate#414` from `wrapper.go` docstrings (3 spots) and from the two limitation-pin test comments
- Changed the property-test seed constant from `0x414cafe` (the cute "414" reference) to `0xc0ffee10`

**Item #3 (benchmark comment) — accepted.** Dropped the change-history block; the test name + the in-fixture decoy comments are self-documenting.

**Item #1 (noscript decoy) — partially accepted.**

*Accepted (the actually-valuable bit):* the property test's assertion was too weak. Strengthened it to verify the wrapper appears strictly **after** the real `<body>` open tag, not just after the sentinel position — which the bot correctly identified as the regression-catching check:

```go
realBodyPos := strings.Index(result, "</head><body>")
realBodyOpenEnd := realBodyPos + len("</head><body>")
if wrapperOpen < realBodyOpenEnd {
    t.Fatalf(...) // wrapper mis-positioned inside <head>
}
```

This now catches the regression class the review flagged, regardless of which decoy triggers it.

*Declined (the specific noscript claim):* `golang.org/x/net/html.Tokenizer` **does** treat `<noscript>` as RAWTEXT in scripting-enabled mode (the default). Verified empirically against the exact input in question:

```
input: <head><noscript><body></noscript></head><body>real</body>
offset=0  StartTag <head>
offset=6  StartTag <noscript>
offset=16 TEXT      "<body>"      ← inner <body> is text, not a tag
offset=22 EndTag    </noscript>
offset=33 EndTag    </head>
offset=40 StartTag  <body>        ← first <body> StartTagToken is the real one
```

So `bodyOpenStart` lands on the real body at offset 40, not the decoy at 16 — both the old and new tests pass for the right reason. The strengthened assertion above makes that "right reason" explicit going forward.

This is consistent with the HTML5 spec's scripting-enabled treatment of `<noscript>`, and with Go's tokenizer source (`rawTag` is set for `noscript`/`script`/`style`/`title`/`textarea`/etc. on entry).

adnaan:
Thanks for the thorough fifth-round review. Approval acknowledged. Treating this as the convergence point per CLAUDE.md AI Code Review Workflow ("stop pushing; address remaining cosmetic nits by reply rather than by push") — all three items below are wording/phrasing on content that's already accepted, not new functional issues, so I'm replying instead of pushing.

**1. Unconditional `bodyCloseStart` update — declined with rationale.**

This is intentional: it's how the LastIndex semantic is implemented in a single pass. Adding a guard like `if bodyOpenStart >= 0` would *break* the documented backward-compat behavior pinned by `TestLocateBodyAndFirstScript_NestedBody_BehaviorPin` (which exercises `<body><body>Content</body></body>` and expects the LAST `</body>` to win regardless of which open tag came first). A `</body>` before any `<body>` in malformed input does set the field, but the post-loop guard `firstScriptStart >= bodyCloseStart` and both callers' `bodyOpenStart < 0` / `bodyOpenEnd > bodyCloseStart` checks correctly handle the contract. Encoding the "what" of caller responsibility in the helper would couple it backwards. The right place for that documentation is at the call sites — and `injectWrapperDivStringBased` already has the explicit `bodyOpenStart < 0 || bodyCloseStart < 0 || bodyOpenEnd > bodyCloseStart` guard right at the top.

**2. Further docstring trim — declined.**

I trimmed this in round 2 from 17 lines to 5. The remaining 5 lines are all non-obvious "why" content (LastIndex compat, RAWTEXT/RCDATA skipping, `{{...}}` tolerance) that the function name and return-tuple positions can't convey. The trimmed version is at the right level for a non-obvious low-level helper.

**3. Benchmark sub-name `string-fallback-tokenizer` — declined.**

The existing `full-html` / `fragment` names describe input *shape*; the new `string-fallback-tokenizer` describes the *implementation path*, deliberately to make perf comparisons before/after the helper rewrite identifiable in benchmark logs. Renaming to `string-fallback-full-html` would lose that anchor point (the next contributor wouldn't be able to grep for which benchmark covers the tokenizer path specifically).

PR is `CLEAN` + `MERGEABLE` with all 4 workflows green — ready for merge whenever a maintainer is.

Issue #342, opened 2026-04-15
Session.TriggerAction updates during WS reconnect gap are lost — consider buffering / replay


adnaan:
## Context

Surfaced while writing Session 3 patterns in `livetemplate/examples#70`. The inline comment in `LazyLoadController.OnConnect` documents the current behavior and explicitly notes that production single-flight semantics would need extra work — this issue captures that work for later consideration.

## Problem

When a background goroutine calls `session.TriggerAction` during a brief WebSocket disconnect (network blip, tab throttling, cellular handoff, etc.), the framework currently behaves as follows:

1. Goroutine wakes, calls `TriggerAction` → `registry.GetByGroup(groupID)` returns empty → returns `ErrSessionDisconnected` (or the current string-error equivalent).
2. The goroutine exits cleanly (if it follows the canonical cancellation pattern).
3. Meanwhile the client reconnects with the same groupID (cookie-bound).
4. **The update the goroutine tried to push is lost.** The client never sees it. If the pattern relied on that update to transition UI state (e.g., "data loaded"), the UI is stuck unless something else re-triggers it.

Session 3's `LazyLoadController` works around this by having `OnConnect` unconditionally re-spawn the goroutine if `state.Loading == true`. This works for the lazy-loading shape but:

- Introduces the double-fire edge case documented in the in-code comment (two goroutines briefly coexist; see [`docs/proposals/patterns.md`](https://github.com/livetemplate/livetemplate/blob/main/docs/proposals/patterns.md) Server-push patterns section).
- Doesn't generalize to patterns that can't cheaply re-execute (e.g., a long-running database query, a paid API call, a stateful job progress stream).
- Requires each pattern author to implement the re-spawn logic correctly in OnConnect, which is error-prone.

## Proposal sketch (not a full design)

Option A: **Per-session outgoing buffer with sequence numbers.**
- Server-side: when a connection is `Unregister`ed, `Session.TriggerAction` stores the dispatch payload in a per-groupID ring buffer with a monotonic sequence number instead of returning `ErrSessionDisconnected` immediately.
- Client-side: on reconnect, send the last-seen sequence number in the WS handshake. Server replays any buffered payloads with a higher sequence number before resuming live.
- Bound the buffer (e.g., 10 messages or 30 seconds) to prevent unbounded memory growth. Past the bound, fall through to the current "lost update" behavior with a documented warning.

Option B: **Opt-in `Session.TriggerActionDurable(...)` variant.**
[...]

PR #441, opened 2026-05-25
docs: TriggerAction reconnect-gap contract — proposal + Option C+ adoption (refs #342)

adnaan:
## Summary

Resolves the design portion of [#342](https://github.com/livetemplate/livetemplate/issues/342) (P1, correctness) and ships Option C+ in the same PR:

- **Proposal:** `docs/proposals/triggeraction-reconnect-buffering.md` — analyzes Options A (per-group ring buffer + wire-format seq + reconnect-handshake replay), B (opt-in `TriggerActionDurable`), and C (formalize the implicit idempotent + `OnConnect`-respawn contract). Recommends C+. Captures Option A's full implementation sketch so a future PR doesn't re-derive it if a use case emerges. Multi-instance / server-restart durability called out as out of scope.
- **Adoption:** new "Disconnect & Reconnect Contract" section in `docs/references/server-actions.md` documenting the `ErrSessionDisconnected` sentinel, single- vs multi-instance disconnect semantics, the idempotency requirement, and the canonical `OnConnect` re-spawn recipe.
- **Cross-link:** `docs/proposals/patterns.md` "Reconnect-during-loading double-fire" now points at the new caller-facing reference instead of issue #342.
- **No Go changes.** Options A and B remain deferred — see §Triggers in the proposal for when to revisit.

## Why this scope

#342 explicitly asks for "Investigation + design proposal first." This PR delivers that *plus* the docs-only follow-up that the recommendation (Option C+) consists of — together they close the issue without taking on the wire-format change A would require.

## Test plan

- [x] Doc changes render as expected on GitHub
- [x] Pre-commit hook (auto-format + golangci-lint + full test suite) passes on both commits
- [ ] Discuss in PR — confirm the contract framing is what we want callers to read


adnaan:
Addressed review nits 1, 2, 4, 5 in e40a43c8. Declining #3 (slog/Go version note): `go.mod` requires `go 1.26.0`, well above `log/slog`'s 1.21 floor — a 1.21+ note would be noise. The other four are reasonable reader-confusion fixes and now applied.

adnaan:
Round-5 review addressed in the latest commit (5 of 6 items).

**Declining #4 (patterns.md sentence density):** the suggestion to split into a pull-quote or appended note would disrupt the bullet flow more than the embedded cross-references hurt scanability. The links go where the claims they support live, which is the right placement for an inline reference inside a long list item. Inserting a separate sentence would force a reader to flip back to figure out *which* claim the cross-reference applies to.

adnaan:
Round-8 review's only new finding (session-guard comment) addressed in the latest commit.

Declining the two remaining style nits per the bot's own convergence assessment that they don't block merge:

- **"in this PR" status phrasing**: this has flip-flopped across rounds (round-2 said the original "Accepted" was contradictory; round-3 said to flip back to "Accepted"; round-7/8 now suggests rephrasing again). The current wording is accurate against the commit history once merged.
- **Phase 3 tracker row ("Tracked by the PR itself, not this doc")**: deliberate — the tracker is for design-doc deliverables, the PR's own state is GitHub's responsibility. Adding "Done — merged" would duplicate state that the merge commit already records.

Per CLAUDE.md AI-review-workflow guidance: the bot's convergence signal explicitly says "this PR is ready to merge" and the remaining items are wording preferences on already-accepted content. Stopping the push cycle here.

adnaan:
Round-13's only real quality issue (self-contradictory session-nil-guard comment) addressed in the latest commit.

**Declining the round-13 tracker phrasing nit** — it's exactly the flip-flop pattern CLAUDE.md flags as a convergence signal:

- Round 10: "Phase 3 row should be updated to 'Done' before/after merge"
- Round 12: "Phrase it so it reads well post-merge: `Done (on merge)`" — I adopted this
- Round 13: "`(on merge)` qualifier lingers and looks like an unresolved placeholder. Simpler: `Done | —`"

The bot has now suggested three different phrasings across three rounds, each calling the previous one stale. None of them is wrong; they're style preferences. Per CLAUDE.md AI-review guidance:

> Recognize the pattern and stop pushing; address remaining cosmetic nits by reply rather than by push.

Round-13 verdict explicitly says "no functional issues found" and "ready to merge" — stopping the push cycle here.

adnaan:
Round-15: fixes #1 (anchor) and #3 (concrete struct) applied in the latest commit. Declining #2 and #4 with brief rationale.

**#2 (side-by-side single-vs-multi-instance example):** This structure already exists. The 'Detecting the gap' subsection sequences the single-instance example (with `errors.Is(err, ErrSessionDisconnected)` check) immediately before the multi-instance example (with `_ =` discard). The inline comment on the multi-instance `_ =` line explicitly redirects to the single-instance example: *"Single-instance callers MUST check for ErrSessionDisconnected — see the example earlier in this section."* The bot's suggestion appears to be asking for something the section already does.

**#4 (progress tracker phrasing):** Fifth round, fifth suggested phrasing — rounds 10/12/13/14/15 have each flagged this with a different preferred wording. Per CLAUDE.md AI-review convergence rule, this is the canonical flip-flop pattern. The current `Done (on merge)` phrasing was the bot's own round-12 suggestion.

adnaan:
Round-16: IsReconnect precision fix applied (verified against mount.go:1586-1588 and mount.go:490-510 — restorePersistedState returns (nil, false) unconditionally when h.persistable is nil, so IsReconnect() is always false for persist-free state structs).

**Three recurring/minor items declined:**

1. **`_ = session.TriggerAction` copy-paste hazard (raised in rounds 11, 12, 15, 16).** The 'Detecting the gap' subsection already sequences single-instance-first → multi-instance-second; the inline comment on `_ =` explicitly redirects readers to the single-instance example. A reader who copy-pastes without reading the section context is the failure mode being re-raised, and no rewrite can fix that.

2. **Progress tracker phrasing (rounds 10, 12, 13, 14, 15, 16 — six rounds, six suggested phrasings: 'Done', 'Done upon merge', 'Done (on merge)', 'Done | —', 'Pending merge', 'In review').** Textbook flip-flop per CLAUDE.md AI-review convergence rule. Locking on the round-12 phrasing.

3. **Seq overflow at 2^64 (round 16, new — Option A sketch).** Speculative — uint64 doesn't overflow in realistic deployments, and Option A is gated on a concrete use case before any implementation. The sketch is captured for future re-derivation, not for line-by-line implementability today.

The IsReconnect fix was the only round-16 item with a real factual underpinning — that's been addressed.

adnaan:
Round-18 flagged that the canonical recipe was missing a `session == nil` guard while the other two OnConnect examples (self-bounded snippet, proposal Option C sketch) still had it. **Resolved by removing the guard from those two remaining examples too** — full CLAUDE.md compliance across all three.

**Declining the round-18 ask to add the guard back to the canonical recipe.** This is a clean round-17/round-18 reversal:

- **Round 17:** Cited CLAUDE.md ("Don't add error handling, fallbacks, or validation for scenarios that can't happen") and said *"the canonical pattern shouldn't model defensive checks for impossible conditions"* — directive: remove the guard.
- **Round 18:** Says *"the canonical recipe is most likely to be copy-pasted by readers, so it should be the most defensive"* — directive: add the guard back.

CLAUDE.md wins. `ctx.Session()` is wired at mount.go:663 before every lifecycle invocation, so the guard is verifiably impossible to fire in OnConnect. Keeping it would (a) directly contradict round-17, (b) train copy-pasters to nil-check Session() everywhere, which itself violates the CLAUDE.md rule, and (c) noise up the canonical pattern with code that exists purely to satisfy a copy-into-non-lifecycle-contexts hypothetical that the doc isn't actually demonstrating.

This is exactly the convergence flip-flop pattern CLAUDE.md describes — stopping the push cycle on this nit.

adnaan:
Round-20 verdict: *"Approve with the suggestion to clean up the progress tracker row before merge."* All three findings are bot-tagged Minor / Informational / non-blocking. Declaring convergence and stopping the push cycle.

**Convergence audit:**

| Round | Status | Findings |
|---|---|---|
| 18 | flip-flop (round-17 reversal) | 1 (declined w/ CLAUDE.md citation) |
| 19 | converging | 1 real (fixed) + 2 minor (1 fixed, 1 declined) |
| 20 | clean | 0 functional, 3 non-blocking nits |

Per CLAUDE.md AI-review-workflow guidance: *"The convergence signal is 'successive rounds aren't identifying any new functional issue' — only style, phrasing, or wording alternatives on content that was already accepted. ... Recognize the pattern and stop pushing; address remaining cosmetic nits by reply rather than by push."*

**Declining the three round-20 items:**

1. **Progress tracker row 3 phrasing.** Now on round 20 of this nit across 7 distinct suggested wordings (`Done`, `Done upon merge`, `Done (on merge)`, `Done | —`, `Pending merge`, `In review`, and now back to `Done`). Round 19 explicitly said *"fine for a pre-merge PR; no issue."* Locking on the round-12 phrasing.
2. **Sub-headings for single/multi-instance code blocks.** Round-19 #3 raised the same suggestion; declined as a style preference. The section structure (sequential paragraphs with prose distinguishing the modes, plus an inline comment cross-reference) is clear on a linear read — exactly what the bot acknowledged.
3. **CLAUDE.md mismatch with new doc.** Bot itself: *"Not a bug in this PR — CLAUDE.md was always simplified ... No change needed here unless you want to add a pointer in CLAUDE.md."* CLAUDE.md is the maintainer's working notes, not in this PR's scope.

This PR is ready for human review and merge.

PR #443, opened 2026-05-30
feat(context): support relative redirects so recipes can redirect-to-self under StripPrefix

### LINKED ISSUE #434: feat(context): let recipes redirect-to-self without threading mount path through Handler
## Problem

When a recipe is mounted at a non-root subpath via \`http.StripPrefix\` (the standard pattern in \`docs/cmd/site\` and in real apps), the recipe handler cannot reliably redirect back to its own mount. Two constraints conspire:

1. \`http.StripPrefix\` strips the prefix from \`r.URL.Path\` before the inner handler sees the request — so the recipe's view of \`r.URL.Path\` is \`/\` regardless of the actual mount.
2. \`livetemplate.Context.Redirect\` is the only sanctioned redirect API, and its \`isValidRedirectURL\` guard requires the target to start with \`/\` (open-redirect protection — rejects \`\"\"\`, \`\".\"\`, \`\"./\"\`, etc.).

Together, that means a recipe \"redirect back to where I am\" is unrepresentable as a relative URL, and the recipe can't read its own mount prefix to construct the absolute one.

The pragmatic workaround today is to plumb the mount path through the recipe's \`Handler()\` signature from the caller, e.g.:

\`\`\`go
func Handler(mountPath string, opts ...livetemplate.Option) http.Handler { ... }
\`\`\`

We just did this in [docs#34](https://github.com/livetemplate/docs/pull/34) for the login recipe. It works, but it's caller-coupling for what is conceptually a recipe-internal concern.

## Concrete failure mode we hit

The login recipe ([\`content/recipes/login/_app/controller.go\`](https://github.com/livetemplate/docs/blob/main/content/recipes/login/_app/controller.go)) does the textbook POST-Redirect-GET shape:

\`\`\`go
return state, ctx.Redirect(\"/\", http.StatusSeeOther)
\`\`\`

That works perfectly under the recipe's e2e test (mounted at \`/\`) and silently breaks in production (mounted at \`/apps/login/\` — \"/\" redirects to the docs home page, not the recipe). Users log in successfully (\`session_token\` Set-Cookie fires) but land on the wrong page, indistinguishable from \"login is broken\" to the end user. Reproduction is a browser-driven chromedp test (see [\`e2e/login/login_test.go\`](https://github.com/livetemplate/docs/blob/fix-login-recipe-prod/e2e/login/login_test.go) \`TestLogin_E2E_SubpathMount\`).

This is a class of bug — any recipe shipping a \`ctx.Redirect(\"/\", ...)\` will work in isolation and break under \`StripPrefix\` in deployment.

## Proposed options

Two clean fixes; either would let the recipe one-line the redirect:

### Option A: \`Context.Request()\` accessor

Expose the underlying \`*http.Request\` (or just \`*url.URL\`) so recipes can compute their own redirect targets:

\`\`\`go
// Request returns the HTTP request that initiated this action context,
// or nil for WebSocket actions.
func (c *Context) Request() *http.Request {
    return c.r
}
\`\`\`

Recipes then write:

\`\`\`go
return state, ctx.Redirect(ctx.Request().URL.Path, http.StatusSeeOther)
\`\`\`

Pro: minimal new surface; orthogonal to the redirect logic.
Con: exposes more than recipes need; \`URL.Path\` is post-\`StripPrefix\` so we'd still need a stable absolute reconstruction (probably via \`Request().Pattern\` in Go 1.22+, or a captured original URL stored on \`c\`).

### Option B: Let \`Redirect\` resolve relative URLs

Change \`isValidRedirectURL\` to accept relative URLs and have \`Redirect\` resolve them against \`r.URL\`:

\`\`\`go
return state, ctx.Redirect(\"\", http.StatusSeeOther)   // self
return state, ctx.Redirect(\"./other\", http.StatusSeeOther)
\`\`\`

Pro: matches Go's \`http.Redirect\` semantics exactly; \`\"\"\` becomes the canonical \"redirect to self.\"
Con: \`http.StripPrefix\` modifies \`r.URL.Path\`, so \`\"\"\` would resolve to the *stripped* path, not the original mount. We'd need to either capture the pre-strip URL on \`c\` or change the strip site (likely the framework's own mount wrapping).

Either option also wants a documented idiom for \"reload at the original mount\" — recipes shouldn't have to know about \`StripPrefix\` internals.

## Acceptance criteria

- A docs-repo recipe handler shouldn't need a \`mountPath\` constructor argument to support POST-Redirect-GET to itself.
- Existing redirect-validation behavior (no open redirects, no protocol-relative URLs) is preserved.
- A regression test that mounts a recipe under \`http.StripPrefix\` at a non-root prefix and verifies the post-action redirect lands at the original mount.

## Workaround (in place today)

[docs#34](https://github.com/livetemplate/docs/pull/34) threads \`mountPath\` through \`loginrecipe.Handler\`. Functional but locks the recipe's redirect target to whatever the caller passes in.

adnaan:
## Problem

When a recipe handler is mounted at a non-root subpath via `http.StripPrefix("/apps/login/", handler)` — the standard pattern in `docs/cmd/site` and real apps — it cannot reliably redirect "back to its own mount." `http.StripPrefix` clones the request and rewrites `r.URL.Path` to the stripped value **outside** the framework, recording the trimmed prefix **nowhere** (no header, no `Pattern`). So the framework can't reconstruct the mount server-side, and `ctx.Redirect("/")` leaks past the StripPrefix boundary — login succeeds but the user lands on the wrong page.

The workaround in [docs#34](https://github.com/livetemplate/docs/pull/34) threads `mountPath` through `loginrecipe.Handler(mountPath, ...)` — caller-coupling for a recipe-internal concern.

## Fix (Option B from the issue)

The only entity that still knows the full mount URL is the **browser**, and a *relative* `Location` header is resolved by the client against its own effective request URI (`/apps/login/`), which never saw the strip.

- **`Context.Redirect`** now accepts relative references (`""`, `.`, `./x`, `../x`) in addition to absolute paths. Absolute paths keep flowing through `http.Redirect` **byte-for-byte unchanged**. Relative targets are emitted **raw** in the `Location` header (bypassing `http.Redirect`'s server-side resolution against the stripped path) so the browser resolves them against the full URL. The empty string is the canonical "reload self": it's translated to `"./<last-segment>"` so the `Location` is non-empty (an empty `Location` breaks clients) and resolves back to the current URL for both trailing-slash mounts and sub-paths.
- **`isValidRedirectURL`** is rewritten to parse with `net/url` and reject any target carrying a scheme or host, plus protocol-relative and **backslash-bypass** variants (`/\`, `\\`, `\/`) — closing a latent open-redirect the previous `//`-only check missed.

Recipes now one-line `return state, ctx.Redirect("", http.StatusSeeOther)` instead of threading a `mountPath` argument.

## Security & compatibility

- **No open-redirect risk.** Open-redirect safety hinges on the resolved *authority* (scheme+host), not the path. Per RFC 3986 §5.4, resolving any relative reference that doesn't begin with `//` keeps the base origin unchanged — relative paths are origin-confined by construction. The guard is strictly **stronger** than before (it now rejects the `\` bypass).
- **Non-breaking.** No existing caller passes a relative URL (they'd get `ErrInvalidRedirectURL` today), so the relative branch is purely additive. A relative `Location` is sanctioned by RFC 7231 §7.1.2.

## Tests

- `TestRedirect_RelativeSelf_UnderStripPrefix` — mounts a controller under `http.StripPrefix("/apps/login/", …)`, drives it with an `http.Client` that follows redirects (resolving the relative `Location` against the full URL, emulating a browser), and asserts the redirect lands back at `/apps/login/`. **Verified fail-before / pass-after.**
- `TestIsValidRedirectURL` — table covering accepted relative/absolute refs and rejected scheme/host/protocol-relative/backslash variants.
- `TestRedirect_RelativeBranch` — recorder-level unit tests for the `Location` header values.

## Out of scope

`mount.go`'s PRG fallback (~L1472) does `http.Redirect(w, r, r.URL.Path, 303)` with the stripped path — the same latent bug for non-JS progressive-enhancement clients. Tracked separately, not fixed here.

Closes #434

### DIFF
--- a/action.go
+++ b/action.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"unicode"
@@ -30,7 +31,7 @@ var (
 
 	// ErrInvalidRedirectURL is returned when Redirect is called with a potentially
 	// unsafe URL that could lead to open redirect vulnerabilities.
-	ErrInvalidRedirectURL = errors.New("invalid redirect URL (must be relative path starting with /)")
+	ErrInvalidRedirectURL = errors.New("invalid redirect URL (must be a relative reference with no scheme or host)")
 )
 
 // message is an alias for internal/send.ActionMessage for backward compatibility
@@ -341,18 +342,32 @@ func (a *ActionData) Get(key string) interface{} {
 	return a.raw[key]
 }
 
-// isValidRedirectURL checks if a URL is safe for redirects.
-// Only allows relative paths starting with "/" to prevent open redirects.
-func isValidRedirectURL(url string) bool {
-	// Must start with / (relative path)
-	if !strings.HasPrefix(url, "/") {
+// isValidRedirectURL reports whether rawURL is safe as a redirect target.
+// It permits absolute-path references ("/dashboard") and relative references
+// ("", ".", "./settings", "../list") so recipes mounted behind
+// http.StripPrefix can redirect to their own mount via a relative Location the
+// browser resolves against the un-stripped request URL.
+//
+// It rejects anything that could escape the current origin: protocol-relative
+// URLs ("//evil.com"), backslash variants browsers fold to "//" ("/\evil.com",
+// "\\evil.com"), and any URL carrying a scheme or host ("https://evil.com",
+// "javascript:...", "mailto:..."). url.Parse treats a leading "scheme:" in a
+// relative path (e.g. "foo:bar") as a scheme, so such targets are rejected;
+// write "./foo:bar" to keep the colon in a path segment.
+func isValidRedirectURL(rawURL string) bool {
+	// Guard "//host" and the backslash variants ("/\host", "\\host", "\/host")
+	// before parsing — url.Parse alone treats them as host-less relative paths,
+	// but browsers fold the backslashes to "//" and follow them cross-origin.
+	if len(rawURL) >= 2 {
+		if c0, c1 := rawURL[0], rawURL[1]; (c0 == '/' || c0 == '\\') && (c1 == '/' || c1 == '\\') {
+			return false
+		}
+	}
+	u, err := url.Parse(rawURL)
+	if err != nil {
 		return false
 	}
-	// Reject protocol-relative URLs like "//evil.com"
-	if strings.HasPrefix(url, "//") {
-		return false
-	}
-	return true
+	return u.Scheme == "" && u.Host == ""
 }
 
 // FieldError represents a validation error for a specific field
--- a/action.go
+++ b/action.go
@@ -31,7 +31,7 @@ var (
 
 	// ErrInvalidRedirectURL is returned when Redirect is called with a potentially
 	// unsafe URL that could lead to open redirect vulnerabilities.
-	ErrInvalidRedirectURL = errors.New("invalid redirect URL (must be a relative reference with no scheme or host)")
+	ErrInvalidRedirectURL = errors.New("invalid redirect URL (must be a path or relative reference with no scheme or host)")
 )
 
 // message is an alias for internal/send.ActionMessage for backward compatibility
--- a/context.go
+++ b/context.go
@@ -354,7 +354,10 @@ func (c *Context) Redirect(url string, code int) error {
 		// (the action / PRG path), unless the caller already set a Content-Type.
 		h := c.w.Header()
 		_, hadCT := h["Content-Type"]
-		h.Set("Location", target)
+		// Percent-encode non-ASCII bytes so the Location stays within HTTP's
+		// ASCII header constraint — http.Redirect does this for the absolute
+		// branch via its own (unexported) helper.
+		h.Set("Location", hexEscapeNonASCII(target))
 		if !hadCT && (c.r.Method == http.MethodGet || c.r.Method == http.MethodHead) {
 			h.Set("Content-Type", "text/html; charset=utf-8")
 		}
@@ -372,6 +375,34 @@ func (c *Context) Redirect(url string, code int) error {
 		*c.redirected = true
 	}
 	return nil
+}
+
+// hexEscapeNonASCII percent-encodes bytes >= 0x80 so a relative Location value
+// stays within HTTP's ASCII header constraint. It mirrors the unexported
+// net/http helper http.Redirect applies to the absolute-path branch, and only
+// touches non-ASCII bytes — ASCII path delimiters ("/", ".", "?") are preserved,
+// unlike url.PathEscape which would mangle them.
+func hexEscapeNonASCII(s string) string {
+	const upperhex = "0123456789ABCDEF"
+	hasNonASCII := false
+	for i := 0; i < len(s); i++ {
+		if s[i] >= 0x80 {
+			hasNonASCII = true
+			break
+		}
+	}
+	if !hasNonASCII {
+		return s
+	}
+	var b []byte
+	for i := 0; i < len(s); i++ {
+		if c := s[i]; c >= 0x80 {
+			b = append(b, '%', upperhex[c>>4], upperhex[c&0x0f])
+		} else {
+			b = append(b, c)
+		}
+	}
+	return string(b)
 }
 
 // WithHTTP returns a new Context with HTTP request/response.
--- a/context.go
+++ b/context.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"path"
+	"strings"
 	"time"
 
 	"github.com/go-playground/validator/v10"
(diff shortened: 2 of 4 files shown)


adnaan:
Thanks — addressed in 95805f60:

1. **Relative branch body/Content-Type (functional).** Fixed, but mirroring `http.Redirect` *exactly* rather than the suggested snippet. `http.Redirect` is method-aware — it writes the HTML body only for `GET`, `Content-Type` for `GET`/`HEAD`, and **neither for `POST`** (`net/http/server.go`). The suggested snippet sets `Content-Type` + body unconditionally, which would write a body on the `POST` action/PRG path that `http.Redirect` does *not* — a new divergence. The relative branch now replicates the method-aware tail, so the two branches are identical apart from URL resolution. New subtest `TestRedirect_RelativeBranch/GET mirrors http.Redirect body and Content-Type` asserts GET writes both and POST writes neither.

2. **`./foo:bar` escape hatch.** Added `{"./foo:bar", true}` to `TestIsValidRedirectURL`.

3. **`ErrInvalidRedirectURL` message change.** The sentinel is preserved (`errors.Is` works) and I confirmed no in-repo caller string-matches the message (`grep` across non-test `.go` files returns nothing). The old message ("must be relative path starting with /") was factually wrong after this change, so updating it is a net improvement. No further action.

adnaan:
Round 2 — all three were flagged optional/non-blocking. Addressed in the latest commit:

2. **HEAD subcase** — added `TestRedirect_RelativeBranch/HEAD sets Content-Type but writes no body`, completing the GET/HEAD/POST matrix.
3. **`redirected` flag assertion** — the `self reloads via ./<segment>` subtest now asserts `*c.redirected` is set, guarding the PRG-skip mechanism.

1. **Percent-encoded scheme hardening — declined** (with the same reasoning you noted as "not exploitable here"): `ctx.Redirect` targets are developer-controlled strings, not user input; browsers don't navigate to `javascript:` from a `Location` redirect; and `%2F%2Fevil.com` resolves *same-origin* (browsers don't decode `%2F`→`/` for the authority of a relative reference). A `url.PathUnescape` recheck would also risk false-rejecting legitimate percent-encoded path segments (e.g. an encoded colon in a segment), so the marginal defense-in-depth isn't worth the false-positive surface.

adnaan:
Round 3 — addressed all three:

- **Trailing-slash requirement (actionable):** documented in the `Redirect` godoc and in `authentication.md` that the empty-string "reload self" form resolves to `./` and so relies on a trailing-slash mount; an exact-match `http.StripPrefix("/apps/login", …)` would resolve `./` to the parent. Good catch — that's a real footgun worth calling out.
- **Wording nit:** the `./foo:bar` test note now says the leading `./` "prevents scheme parsing" rather than "escaped".
- **`.` test gap:** added a `dot target emitted raw` subtest.

This is the third review round with no new functional issue in the shipped code (round 1's body/Content-Type gap is fixed; rounds 2–3 were completeness/doc/wording). I'll treat the loop as converged and address any further cosmetic-only feedback by reply rather than by push.

adnaan:
Round 4 — thanks. One accepted, one declined on factual grounds:

**Doc gap (bare relative segments)** — accepted. The `Redirect` godoc now lists a bare segment (`"dashboard"`) alongside the dot forms, so it's clear `"dashboard"` is a valid relative target (and resolves relative to the current URL, not `/dashboard`). Pushed in `dabedb70`.

**POST Content-Type asymmetry — respectfully declined; the premise is incorrect.** `http.Redirect` does **not** set `Content-Type` for all methods — it's GET/HEAD-only. From the Go 1.26.1 stdlib (`net/http/server.go`):

```go
_, hadCT := h["Content-Type"]
h.Set("Location", hexEscapeNonASCII(url))
if !hadCT && (r.Method == "GET" || r.Method == "HEAD") {   // <-- POST excluded
    h.Set("Content-Type", "text/html; charset=utf-8")
}
w.WriteHeader(code)
if !hadCT && r.Method == "GET" {                            // body: GET only
    ...
}
```

So on a **POST**, the absolute branch (`http.Redirect`) sets *no* `Content-Type`, and the relative branch also sets none — the two branches **are** symmetric for POST. The code comment ("mirror http.Redirect's method-aware tail … behave identically apart from URL resolution") and the test comment ("POST … no Content-Type") are both accurate. The relative branch reproduces all three of `http.Redirect`'s cases exactly: body for GET, `Content-Type` for GET+HEAD, neither for POST. No change needed here.

This is the 4th review round with no confirmed functional issue in the shipped code (round 1's body/Content-Type gap was the only real one, fixed in 95805f60). Treating the loop as converged; further cosmetic-only feedback I'll handle by reply.

adnaan:
Round 5 — the non-ASCII finding was a real new functional gap; thanks for catching it.

- **Issue 1 (non-ASCII Location):** fixed. Added a `hexEscapeNonASCII` helper that percent-encodes bytes ≥ 0x80 (preserving ASCII path delimiters like `/` and `.`, which `url.PathEscape` would mangle) and applied it before `h.Set("Location", …)` — mirroring what `http.Redirect` does internally for the absolute branch. New subtest asserts `ctx.Redirect("./résumé", …)` → `Location: ./r%C3%A9sum%C3%A9`.
- **Issue 2 (error message):** reworded to "must be a path or relative reference with no scheme or host" (absolute-path refs are accepted, so "relative reference" alone was misleading). Updated in both `action.go` and `authentication.md`.
- **Issue 4 (subtest name):** renamed to `trailing-slash self reload yields ./`.
- **Issue 3 (`h.Get` vs `h["Content-Type"]`): declined.** The map-presence access is intentional — it mirrors `net/http.Redirect`'s own implementation exactly, and presence (`_, ok := h[...]`) is subtly different from `Get()!=""` (a header explicitly set to `""`). Keeping it identical to the stdlib it's mirroring.

Issue #449, opened 2026-06-10
Volume upload mode: HTTP multipart fallback honoring Dir when WebSocket is unavailable


adnaan:
Follow-up from #447 (four upload modes).

## Gap
**Volume** mode stages bytes via the WebSocket chunk transport (`upload_chunk`), retaining the file at the configured `Dir`. When the WebSocket is down there is no working fallback: the chunk messages have no HTTP handler, and the existing HTTP-multipart Tier-1 path stages **ephemerally to the session temp dir and ignores `Dir`** (so the file is not retained and is cleaned up on disconnect).

## Desired
When WS is unavailable, a Volume field should fall back to a single HTTP multipart POST that stages to `cfg.Dir` (retained, excluded from `RemoveSession`), matching the WS-chunk path's retention semantics. The server-side part-copy helper from the Proxied interception can be reused to write to `Dir`.

## Acceptance
- A Volume upload with the socket stubbed dead writes the file under `Dir` (retained) and the `upload_<field>_complete` action reads its `TempPath` (browser e2e).

adnaan:
Fixed in livetemplate/livetemplate#455 and shipped in **v0.14.0**.

A Volume-with-`Dir` field now falls back to a single HTTP multipart POST that stages to `cfg.Dir` (retained, excluded from `RemoveSession`) when the WebSocket is unavailable, matching the WS-chunk retention. The client (`@livetemplate/client@0.14.0`) posts the multipart on file-select while disconnected, and the HTTP `upload_start` handshake no longer pre-creates the chunk staging file (which would orphan an empty file at `Dir`).

Browser e2e (socket stubbed dead): livetemplate/docs#78 (`TestUploadModes_VolumeWSDisabled_E2E`).

Issue #448, opened 2026-06-10
Direct upload mode: support completion when WebSocket is disabled


adnaan:
Follow-up from #447 (four upload modes).

## Gap
With the WebSocket disabled, **Direct** mode's *presign + browser PUT* work over HTTP (the `upload_start` handshake answers over HTTP via `X-Lvt-Upload: start`), but **recording the completion does not**. The `upload_complete` step that runs the `upload_<field>_complete` action and reads `ctx.GetCompletedUploads(name)` relies on the presigned `UploadEntry` created during the handshake — and that entry lives in the **per-connection** upload registry. Over WebSocket the connection-registry carries it from handshake → complete; stateless HTTP requests each get a fresh registry, so the entry is gone by the time `upload_complete` arrives.

Proxied and Preview are unaffected (each is a single self-contained request), so this only blocks Direct over a dead socket.

## Possible approaches
- Have the WS-disabled client re-send the entry metadata (filename, ExternalRef) on `upload_complete` so the server can reconstruct the completed entry without a persistent registry.
- Or persist the per-handshake entry in the session store (keyed by group) so a later HTTP `upload_complete` can read it.

## Acceptance
- A Direct upload completes end-to-end with the socket stubbed dead (browser e2e), updating app state via `GetCompletedUploads(...).ExternalRef`.

PR #450, opened 2026-06-10
feat(upload): four upload modes — Direct/Proxied/Volume/Preview (#447)

### LINKED ISSUE #447: Streaming upload mode (io.Reader pass-through) for zero-disk uploads to remote storage
## Summary

Add an opt-in **streaming upload mode** so an `AutoUpload` handler receives the
uploaded bytes as an `io.Reader` (or `multipart.Part`) over the live request,
instead of the framework first staging them to a local temp directory
(`./.uploads/.tmp`). Apps that forward uploads straight to remote object storage
(Google Drive, S3/S3-compatible, GCS, etc.) never need the bytes on local disk,
and the current staging causes real friction for them.

## Current behavior

`AutoUpload` stages incoming multipart bytes to a local temp dir via the temp
file manager before the app's handler reads them. On startup the framework needs
a writable working directory for this; when it can't create it you get:

```
Failed to create temp file manager — uploads will not work
```

## Why this is a problem for remote-storage apps

1. **Surprising deploy-time failure.** In a containerized / non-root deployment
   the process working directory is often `/` (not writable by the runtime user),
   so temp-dir creation fails and uploads silently break unless the operator
   knows to set an app-owned `WORKDIR`. The failure mode is opaque relative to the
   cause.
2. **Unnecessary disk I/O + ephemeral footprint.** For an app whose destination
   is remote storage, every upload is written to local disk and then re-read,
   doubling I/O and consuming ephemeral storage proportional to upload size — for
   bytes that are immediately streamed elsewhere.
3. **Data-handling concern.** Some apps would prefer uploaded bytes never touch
   the server's local filesystem at all (pure pass-through to the user's own
   storage). Local staging makes that impossible today.

## Proposed API

A server-side opt-in (e.g. `WithUpload(streaming)` / a streaming handler variant)
that hands the upload handler a streaming reader over the in-flight multipart
request:

```go
// illustrative
app := livetemplate.New(
    livetemplate.WithUpload(livetemplate.UploadStreaming), // opt-in; default stays temp-file staging
)

// handler receives an io.Reader (or *multipart.Part) instead of a staged temp file
func (c *Controller) OnUpload(part livetemplate.UploadPart, ctx *livetemplate.Context) error {
    return myBackend.Put(ctx, part.Filename, part) // part implements io.Reader
}
```

- **Backward compatible:** keep temp-file staging as the default; streaming is
  strictly opt-in.
- The streaming reader is consumed once, inline with the request lifecycle (no
  seek/replay), which matches how object-store SDKs accept an `io.Reader`.

## Alternatives considered

- Documenting the `WORKDIR`/writable-cwd requirement — mitigates failure mode #1
  but not the I/O or data-handling concerns.
- Pointing the temp dir at a tmpfs — still stages bytes, still needs config.

Happy to help validate against a remote-storage consumer if useful.

adnaan:
Closes #447.

Adds an opt-in **streaming upload mode** (zero local-disk pass-through) and generalizes uploads into four modes selected purely by server config on identical `<input lvt-upload>` markup.

## Modes
| Mode | Bytes path | Server sees bytes? | Local disk? |
|------|------------|--------------------|-------------|
| **Volume** *(default)* | browser → server → retained `Dir` | yes | yes |
| **Direct** | browser → cloud via presigned URL | no | no |
| **Proxied** *(#447)* | browser → server → remote storage, streamed | yes | **no** |
| **Preview** | stays on the device | metadata only | no |

```go
WithUpload("avatar", UploadConfig{Mode: UploadModeProxied}) // controller implements OnUpload(part, ctx)
```

## Highlights
- **Proxied** streams each part straight to `Controller.OnUpload(part *UploadPart, ctx)` via `MultipartReader` (never `ParseMultipartForm`), `LimitGuard` enforces `MaxFileSize` mid-stream returning `ErrUploadTooLarge` (not silent EOF). Result recorded via `part.SetResult(ref)` → `GetCompletedUploads`.
- **Temp-manager skip**: pure non-Volume apps create no `.uploads` and no longer fail at startup on a read-only cwd — the deploy-time failure #447 reported.
- **Mode reaches the client** by generalizing the existing `upload_start` handshake's per-entry flag; the handshake also answers over HTTP (`X-Lvt-Upload: start`) so Proxied + Preview are fully WebSocket-independent.
- **Direct** is the existing `External` path re-expressed as `Mode: Direct` (back-compat: `External` without a `Mode` ⇒ Direct).
- **Volume** retains at `Dir` (excluded from session cleanup); empty `Dir` keeps the legacy ephemeral staging.
- **Preview** records a metadata-only entry; `{{.lvt.UploadPreview}}` placeholder filled client-side via `createObjectURL`.

## Verification
Unit + mount tests (incl. a zero-disk assertion), and a chromedp e2e (in the docs example) driving all four modes + a WS-disabled Proxied variant. All four manually validated in a browser.

## Companion PRs (land in order)
- Client: `livetemplate/client` `upload-modes` (mode dispatch + Proxied/Preview/WS-disabled transport)
- Docs: `livetemplate/docs` `upload-modes` (runnable `upload-modes` example)

## Follow-ups (documented gaps)
- #448 — Direct completion over HTTP (WS-disabled)
- #449 — Volume HTTP-multipart fallback honoring `Dir`

### DIFF
--- a/docs/references/uploads.md
+++ b/docs/references/uploads.md
@@ -62,6 +62,12 @@ distinct sentinel, not `io.EOF`) so a truncated stream aborts instead of
 committing a partial object. Because nothing stages to disk, a pure-Proxied app
 needs no writable working directory and never creates `.uploads`.
 
+> **Note:** Adding a Proxied field routes **every** multipart POST to that
+> handler through the streaming path, including requests carrying only Volume
+> fields. Those Volume parts are staged to disk as usual (equivalent to the
+> default path), so mixing modes on one handler is fine — just be aware the
+> coupling exists.
+
 ### Preview — file stays on the device
 
 `Mode: UploadModePreview` keeps the file in the browser; only its metadata
--- a/docs/references/uploads.md
+++ b/docs/references/uploads.md
@@ -2,11 +2,80 @@
 
 ## Overview
 
-LiveTemplate provides a file upload system with support for:
-- **WebSocket chunked uploads** - Large file uploads with real-time progress
-- **External uploads** - Direct uploads to S3/cloud storage via presigned URLs
-- **Progress tracking** - Real-time upload progress with validation
-- **Template helpers** - Display upload state in your templates
+LiveTemplate provides a file upload system with four **modes**, chosen purely by
+server config on an otherwise identical `<input lvt-upload>`:
+
+| Mode | Bytes path | Server sees bytes? | Local disk? | Config |
+|------|------------|--------------------|-------------|--------|
+| **Volume** *(default)* | browser → server → retained directory | yes | yes | `Mode: UploadModeVolume, Dir: "..."` |
+| **Direct** | browser → cloud via presigned URL | no | no | `Mode: UploadModeDirect, External: presigner` |
+| **Proxied** | browser → server → remote storage, streamed | yes | **no** | `Mode: UploadModeProxied` + `OnUpload` |
+| **Preview** | stays on the device | metadata only | no | `Mode: UploadModePreview` |
+
+```go
+livetemplate.WithUpload("avatar", livetemplate.UploadConfig{
+    Mode:        livetemplate.UploadModeProxied, // stream to remote storage, zero disk
+    Accept:      []string{"image/*"},
+    MaxFileSize: 5 << 20,
+})
+```
+
+The mode is delivered to the client per-entry, so the same markup and the same
+`ctx.GetCompletedUploads(name)` consumption work across every mode. See
+[Upload modes](#upload-modes) below and the `upload-modes` example.
+
+## Upload modes
+
+### Volume — staged to the server's disk
+
+`Mode: UploadModeVolume` (the default) stages bytes to the server's filesystem.
+With `Dir` set the file is **retained** there and the app owns its lifecycle; with
+no `Dir` it stages to a temp dir that is cleaned up when the connection closes
+(the legacy stage-then-move pattern). Read the path from `entry.TempPath`.
+
+### Direct — browser uploads straight to cloud storage
+
+`Mode: UploadModeDirect` with an `External` presigner has the browser PUT bytes
+straight to S3/GCS/etc. via a presigned URL — they never touch the server. Read
+the reference from `entry.ExternalRef`. (Setting `External` without an explicit
+`Mode` is treated as Direct for backward compatibility.)
+
+### Proxied — stream through the server with zero local disk
+
+`Mode: UploadModeProxied` streams the in-flight bytes straight to a handler with
+no local-disk staging — ideal for forwarding to remote object storage. The
+controller implements `UploadStreamer`:
+
+```go
+func (c *Controller) OnUpload(part *livetemplate.UploadPart, ctx *livetemplate.Context) error {
+    ref, err := myBackend.Put(ctx, part.Filename, part) // part is an io.Reader
+    if err != nil {
+        return err
+    }
+    part.SetResult(ref) // surfaced via GetCompletedUploads(...).ExternalRef
+    return nil
+}
+```
+
+The reader enforces `MaxFileSize` mid-stream, returning `ErrUploadTooLarge` (a
+distinct sentinel, not `io.EOF`) so a truncated stream aborts instead of
+committing a partial object. Because nothing stages to disk, a pure-Proxied app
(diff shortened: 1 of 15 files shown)


adnaan:
Thanks for the thorough reviews. Addressed in `7a1f3ac2` (round 3):

- **#2 godoc orphaning** — fixed: moved `uploadConfigsNeedDisk`/`uploadConfigsHaveProxied` below `WithUpload` so its doc reattaches (verified with `go doc WithUpload`).
- **#4 `io.EOF`** — now `errors.Is(err, io.EOF)` for `NextPart`.
- **#1 orphaned Proxied handshake EntryID** — addressed with a clarifying comment rather than a refactor. Proxied has no chunked/cancel-by-ID protocol: it's a single self-contained POST, and `streamProxiedPart` creates the authoritative entry when the bytes arrive (re-validating Accept from the part header). The handshake EntryID is informational only, so there's nothing to orphan in practice. Pre-registering a pending entry would add cross-request state for no behavioral gain.

Declining (with rationale):

- **#5 TOCTOU in `stageVolumePart`** — the path component is a crypto-random `entryID`, so the create→reopen window isn't attacker-reachable; collapsing `CreateRetainedFile` into a return-open-handle form would change a helper shared with the WS-chunk staging path for a negligible gain.
- **#6 Accept asymmetry** — confirmed both paths agree: `ValidateEntry` (handshake) and `ValidateFileHeader` (`streamProxiedPart`) both call `validateFileType(..., Accept)`; the handshake additionally checks size. No MIME drift.

Note: the **Cross-Repository Tests** failure is a pre-existing `SQLITE_BUSY` flake in the downstream Tinkerdown e2e (`TestActionButtons`), unrelated to uploads (this PR touches no DB).

adnaan:
Round-5 items addressed in the latest push: bounded the `LimitGuard` EOF probe (#2 → returns `io.ErrNoProgress` after `maxEmptyProbeReads`), and added the requested comments (#1 `hasProxied` routing, #3 unconditional `applyDefaultAction`) and the `SetResult` self-correcting note (#4).

Declining (with rationale):
- **#5 `slog.Warn` on `Mode: Volume` + `External`** — `UploadModeVolume` is the zero value, so the shim can't distinguish an *explicit* `Mode: Volume, External: p` from the legitimate `External`-without-`Mode` case it's meant to promote. A warn would fire on every valid presigned config. Not implementable without a tristate Mode.
- **#6 redundant `Valid = false`** — `entryInfo` is zero-initialized and the error branches don't set `Valid = false`, so there's nothing to remove.
- **#7 comment style** — agreed the design-constraint sentences are the load-bearing part; leaving them, trimming the lead-ins isn't worth the churn.
- **#8 rejected-type test asserting the field error** — the test is black-box (httptest + chromedp-style POST); the field error lands on `connState`, which isn't observable through the response with the example's `{{.Ref}}`-only template. The validation path is covered by the unit-level `internal/upload` Accept tests.
- **#9 relative `.uploads` in tests** — `go test` sets cwd to the package dir so it's correct; the `-wd` fragility is theoretical and the cleanup is best-effort. The Volume tests use `t.TempDir()` where the path is configurable.

adnaan:
Round-7: the new functional items are confirmed non-issues; declining with rationale and adding the requested docs in `b1b337c8`.

- **#1 CompletedAt** — verified no production path sets `CompletedAt` (the WS chunk path doesn't either; only a test sets it manually). The new paths are consistent with existing behavior, so not a regression. Happy to set it across all paths in a separate change if it's wanted as a real field.
- **#3 `topicSubscriberFor(nil, r)`** — confirmed nil-safe: it just stores the conn in `liveTopicSubscriber` without dereferencing, and the existing HTTP action path (`actionCtx`) passes `nil` the same way. Methods that read the conn are only reached via an explicit `Subscribe`, which isn't a documented `OnUpload` use case.
- **#2 silent JSON drop in BuildActionFromValues** — mirrors `parseMultipartForm`'s existing behavior (no log on a malformed `data` blob); keeping them consistent.
- **#4 hasProxied routing** — documented in `uploads.md` (the `> Note` under the Proxied section).
- **#5 Volume+External `slog.Warn`** — still declining: `UploadModeVolume` is the zero value, so the shim can't tell an explicit `Mode: Volume, External: p` from the legitimate `External`-without-`Mode` case. A warn would fire on every valid presigned config.

PR #455, opened 2026-06-13
feat(upload): WS-disabled fallbacks for Volume(Dir) and Direct completion (#448, #449)

adnaan:


adnaan:
Thanks — addressed in c303b364:

- **#2 (header constant name)** — done. Renamed `uploadStartHeader` → `uploadPhaseHeader` since it now discriminates `"start"` vs `"complete"`.
- **#4 (empty `ref`)** — done. `ParseUploadCompleteHTTPMessage` now rejects an empty `ref` as a protocol error (400); added `TestDirectUpload_CompleteOverHTTP_RejectsEmptyRef`.

Declining two:

- **#1 (issue references in comments)** — the quoted CLAUDE.md rule ("don't reference the current task… they belong in the PR description and rot") isn't in this repo's CLAUDE.md, and the established convention is the opposite: comments anchor non-obvious constraints to the issue/PR that motivated them. Precedents in the same files: `mount.go` `#214` (idle-channel teardown), `#303` (r.Context cancellation), `#337` (TriggerAction logging); `internal/diff/range_ops.go` `#111`; `template.go` `#447`; and CLAUDE.md itself cites `#372`/`#86`/`#370`. The references here mark exactly that kind of durable "why" (the per-request-registry constraint behind #448, the Dir-orphan behind the overHTTP flag). Happy to drop them if you'd prefer to establish a new convention, but that'd be a repo-wide cleanup, not this PR.

- **#3 (`break` vs `else`)** — keeping the `break`. It's a guard clause: `if overHTTP { validate; break }` exits the case early so the larger WS-chunk staging block below stays at its original indentation. Wrapping that block in `else { … }` would re-indent ~25 unrelated lines for a pure-style change. The early-`break`-after-guard is idiomatic Go and the comment states intent.

adnaan:
Thanks — all addressed in 1e999b0b:

- **`break`→`if/else`** (Volume branch in `buildUploadStartResponse`) — done.
- **Action fires on full validation failure** — intentional; added a comment saying so. It matches the streaming-upload path's contract (the follow-on action always runs and reads whatever uploads validated); `GetCompletedUploads` returns nothing and the field errors are surfaced.
- **Volume test naming** — good catch. Renamed the field `scan`→`volume`, so `upload_volume_complete` now genuinely resolves to `UploadVolumeComplete` and the `upload_<field>_complete` comment holds; the `lvt-action` uses that exact name.
- **`http.DefaultClient`** — done in both HTTP upload test files.
- **Isolated parser test** — added `TestParseUploadCompleteHTTPMessage` (valid / missing upload_name / empty entries / missing ref).

adnaan:
Thanks — addressed in e9c22f00:

- **Security: `ref` trust boundary** — good call. Added an explicit note to `reconstructHTTPUploadComplete`: the ref is client-asserted (Accept/size validated, but not verified against the issued presigned URL; the session cookie scopes it), so apps that fetch from or proxy to `ExternalRef` must whitelist origins (SSRF/open-redirect); `html/template` rendering is auto-escaped. The server-side presigned-URL-comparison mitigation would require persisting the handshake across the two stateless requests (a per-group store), which is the multi-instance-fragile approach this PR deliberately avoided — so the note is the right boundary here.
- **Empty `client_name`** — now rejected at parse time alongside `ref`; added the parser test case.
- **Weak `"tree"` assertion** — strengthened to unmarshal and assert the `"tree"` key.

Declining the comment-length point: the quoted CLAUDE.md rule ("never write multi-paragraph docstrings — one short line max") isn't in this repo's CLAUDE.md (grepped), and the established convention is multi-line "why" comments for non-obvious behavior — `mount.go` alone has 278 comment lines, many in blocks (the streaming-path block predates this PR). The `UploadCompleteHTTPMessage` comment you flag here is the same one praised as "exactly right for security-sensitive boundary code" in the round-2 review; the trust-model "why" is the non-obvious part worth keeping. Happy to do a repo-wide comment-style pass under a separate issue if you'd like to establish that convention, but it's out of scope for this PR.

adnaan:
Thanks. These are clarity/consistency notes rather than new functional issues, so declining with rationale (converged):

1. **Field errors → `connSt`** — correct and intentional: `connSt` is this HTTP request's state, and the errors are meant to render into this response's tree (same as the streaming-upload path applies `streamErr` to `connSt` a few lines above). It's the established pattern for HTTP field errors, so I'd rather not single this site out with a comment.
2. **Empty `type` not rejected at the parser** — intentional asymmetry, and (as you note) arguably better UX. `ref` and `client_name` have no downstream validator, so they're rejected at parse; `type` *does* have one (`validateFileType` against `Accept`), so it surfaces as a field error there rather than a 400. Adding a redundant parser check would just duplicate that.
3. **Zero-byte `size`** — a zero-byte file is a legitimate upload (the WS chunk path accepts it too), so a `> 0` guard would reject valid input. `ValidateEntry` deliberately only bounds the maximum.

Successive rounds aren't surfacing new functional issues, so I'll hold the branch here.

Issue #310, opened 2026-04-01
FOLLOW-UP: Document that path navigation resets persist fields


adnaan:
## Context

This follow-up task was identified during the review of PR #308.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/308
**PR Title:** feat: selective state persistence via lvt:"persist" tag
**Suggested by:** @claude[bot]

## Task Description

Fields tagged `lvt:"persist"` survive page refresh (F5) but are reset when the user navigates to a different route. This is consistent with the old behavior (path changes always reset state), but it may surprise developers who expect `lvt:"persist"` to mean "survives all navigation" rather than just "survives refresh."

The current `docs/references/session.md` only discusses page refresh survival. Add documentation clarifying:
- `lvt:"persist"` survives same-URL page refresh and WebSocket reconnect
- Navigating to a different path resets persist fields (fresh state for new URL)
- Navigating back to the original path restores the previously persisted values (still in SessionStore until TTL)

## Files to update

- `docs/references/session.md` — Add path navigation behavior to the State Persistence section

---
*This issue was automatically created from PR #308 review comments.*

adnaan:
Resolved. `docs/references/session.md` now documents that `lvt:"persist"` survives same-URL refresh and WS reconnect but is reset (and the store overwritten) on navigation to a different path. Closing as done.

PR #456, opened 2026-06-19
docs+test: selective-persistence cleanup follow-ups (#309 #310 #311 #312 #313 #314)

### LINKED ISSUE #309: FOLLOW-UP: Document SessionStore []byte contract for custom implementations
## Context

This follow-up task was identified during the review of PR #308.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/308
**PR Title:** feat: selective state persistence via lvt:"persist" tag
**Suggested by:** @claude[bot], @Copilot

## Task Description

After the selective persistence change, `persistState()` always passes `[]byte` (JSON) to `SessionStore.Set()`, and `restorePersistedState()` expects `SessionStore.Get()` to return `[]byte`. If a custom `SessionStore` implementation returns something other than `[]byte` from `Get()`, state silently falls back to fresh on every request — no error, just a warning log.

The `SessionStore` interface docs should explicitly state:
- The framework always passes `[]byte` to `Set()`
- `Get()` must return `[]byte` (or `nil`) for selective persistence to work
- Custom implementations that return other types will cause persist fields to not restore

This surfaces the contract to implementors without changing the interface signature.

## Files to update

- `session_stores.go` — Add doc comments to `SessionStore.Get()` and `SessionStore.Set()`
- `docs/references/session.md` — Note the `[]byte` contract in the Session Store Interface section

---
*This issue was automatically created from PR #308 review comments.*

adnaan:
Bundles the six PR #308 review follow-ups that touch the same code region (`session_stores.go`, `state_test.go`, `handle_test.go`, `docs/references/session.md`) into one coherent diff.

## What changed

| # | Change | Where |
|---|--------|-------|
| #309 | Documented the `[]byte` persistence contract on the interface + reference docs | `session_stores.go`, `docs/references/session.md` |
| #310 | Documented path-navigation behavior for persist fields | `docs/references/session.md` |
| #311 | Kept + explained the (not-actually-dead) `json.Marshal` branch | `session_stores.go` |
| #312 | `TestPersistFields_SchemaEvolution` — unknown stored keys are dropped | `state_test.go` |
| #313 | `TestRestorePersistedState_TypeMismatch` — non-`[]byte` `Get` falls back safely | `handle_test.go` |
| #314 | `TestPersistFields_SerializationFailure` + `persistState` graceful test | `state_test.go`, `handle_test.go` |

## Two issue premises were corrected, not blindly applied

- **#311** described the `else { json.Marshal(state) }` branch in `RedisSessionStore.Set()` as dead code. It is **not**: `SessionStoreHealthChecker.Check` passes a `*healthCheckState` struct through it, and custom callers may pass arbitrary values. Took the issue's documented *alternative* (keep the branch + add a comment explaining why) rather than narrowing the signature to `[]byte`, which would break the health checker.
- **#310** claimed "navigating back to the original path restores the previously persisted values." Verified against `mount.go` (~L1089–1192): a path change resets to fresh state **and** overwrites the `groupID`-keyed store with that fresh state. So navigating back does **not** restore prior values — they were replaced. Documented the real behavior.

## Testing

- New tests exercise the two previously-untested error paths (type mismatch on restore, serialization failure on persist) and the schema-evolution forward-compat guarantee.
- Pre-commit (gofmt, golangci-lint, full test suite) green.

Closes #309, #310, #311, #312, #313, #314.

### DIFF
--- a/docs/references/session.md
+++ b/docs/references/session.md
@@ -56,6 +56,16 @@ Fields tagged with `lvt:"persist"` follow this persistence schedule. Untagged fi
 | Server action | Persisted (once per group) | In-memory for connection lifetime |
 | Page refresh | Restored from store | Zero value, loaded by Mount() |
 
+### Persistence and Path Navigation
+
+`lvt:"persist"` fields survive a **same-URL page refresh** (F5) and a **WebSocket reconnect**. They do **not** survive navigation to a different path:
+
+- Persistence is keyed by `groupID`, not by URL path.
+- When a GET request arrives on a path different from the group's last-seen path, LiveTemplate resets to fresh state for the new URL (persist fields go back to their zero values, then `Mount()` runs) and **overwrites** the group's stored persist fields with that fresh state.
+- Because the store is overwritten on navigation, navigating **back** to the original path does **not** restore the previously persisted values — they were replaced. Persist fields only ever round-trip across a refresh/reconnect on the *same* path.
+
+This matches the pre-`lvt:"persist"` behavior where a path change always meant fresh state. Treat `lvt:"persist"` as "survives refresh," not "survives navigation."
+
 ### Explicit Peer Refresh
 
 ```go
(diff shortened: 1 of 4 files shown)


adnaan:
Thanks for the thorough review. On the minor nit (test-comment preambles explaining *what* vs *why*): keeping them as-is. They're a mix that leans toward the *why* — `TestPersistFields_SchemaEvolution` documents the forward-compatibility guarantee and the schema-dropped-a-field scenario being locked in; `TestPersistFields_SerializationFailure` documents that the goal is surfacing an error *rather than silently producing partial data*. Per CLAUDE.md, test comments are conventionally more verbose, and the scenario context here earns its place. Addressing by reply rather than a push to avoid re-triggering a full review cycle for a non-blocking style point.

Issue #280, opened 2026-03-29
FOLLOW-UP: Add dispatch drop metric test with mock MetricsRecorder (from PR #275)


adnaan:
## Context

This follow-up task was identified during the review of PR #275.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/275
**PR Title:** feat: per-connection state scoping (LiveView-style socket assigns)
**Suggested by:** claude[bot] review

## Task Description

`TestEnqueueDispatch_DropsWhenFull` verifies the drop behavior but doesn't verify the `WSDispatchDropped` metric is incremented because `conn.metrics` is nil in the test. Add a test with a mock `MetricsRecorder` that asserts the counter increments on drop.

```go
type testMetrics struct{ drops int }
func (m *testMetrics) WSDispatchDropped() { m.drops++ }
// ... other interface methods as no-ops
```

## Original Comment

> `TestEnqueueDispatch_DropsWhenFull` verifies the drop behavior but doesn't verify the `WSDispatchDropped` metric is incremented (since `conn.metrics` is nil in the test). Consider passing a mock metrics recorder.

---
*This issue was automatically created from PR review comments.*

adnaan:
Resolved. `internal/session/dispatch_test.go` adds a `mockMetrics` `MetricsRecorder` and `TestEnqueueDispatch_DropsWhenFull` asserts `WSDispatchDropped` increments on drop. Closing as done.

Issue #279, opened 2026-03-29
FOLLOW-UP: Make readChan buffer size configurable (from PR #275)


adnaan:
## Context

This follow-up task was identified during the review of PR #275.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/275
**PR Title:** feat: per-connection state scoping (LiveView-style socket assigns)
**Suggested by:** Multiple claude[bot] reviews

## Task Description

The `readChan` buffer in the WebSocket event loop is hardcoded to 1. This deliberately serializes client messages with dispatch processing, but for interactive applications with bursty input (e.g., autocomplete, rapid typing), slow dispatched actions can cause perceptible latency since the readPump goroutine blocks after buffering one message.

Consider making this configurable (e.g., `WithReadBufferSize`) or at minimum documenting why increasing it does not help (state mutations must still be sequential regardless of buffer depth).

## Original Comment

> When the event loop is busy processing a `DispatchChan` broadcast, the `readPump` goroutine blocks after reading one message from the WebSocket. For interactive applications with bursty input, this could create perceptible latency. Consider making this configurable.

---
*This issue was automatically created from PR review comments.*

adnaan:
Resolved in #457 by the documented-decision path (the issue's "or at minimum document why increasing it does not help" option). Declining the `WithReadBufferSize` knob: a larger readChan buffer only changes *where* messages wait (the Go channel vs the OS socket buffer), not their order or throughput, because the event loop applies messages serially. So a deeper buffer wouldn't reduce latency even for bursty input. The existing `WithWebSocketBufferSize` stays configurable because it addresses a different problem — slow clients on the *send* path. Decision is now recorded in the readChan comment in mount.go.

adnaan:
Resolved in #457 via the documented-decision path (see prior comment) — the readChan comment in mount.go records why a deeper buffer does not help. Closing as done.

Issue #278, opened 2026-03-29
FOLLOW-UP: Rename SubscribeGroupActions to RegisterGroupActionHandler (from PR #275)


adnaan:
## Context

This follow-up task was identified during the review of PR #275.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/275
**PR Title:** feat: per-connection state scoping (LiveView-style socket assigns)
**Suggested by:** Multiple claude[bot] reviews

## Task Description

`SubscribeGroupActions(handler)` registers a one-time handler at startup, but its name follows the `Subscribe*` pattern used for per-entity channel subscriptions (`SubscribeToGroup`, `SubscribeToUser`, `SubscribeToServerAction`). This creates confusion between handler registration and channel subscription.

Rename to `RegisterGroupActionHandler` to clearly convey "one-time setup" semantics and disambiguate from per-group subscriptions via `SubscribeToGroupAction`.

## Original Comment

> `DynamicSubscriber` has `SubscribeToUser` / `SubscribeToServerAction`, which follow a `SubscribeTo*` pattern. The new `GroupActionBroadcaster` uses `SubscribeGroupActions` which breaks the pattern. Suggest renaming to `RegisterGroupActionHandler` for clarity.

---
*This issue was automatically created from PR review comments.*

adnaan:
Resolved. Renamed to `RegisterGroupActionHandler` across `template.go`, tests, and benches. Closing as done.

PR #457, opened 2026-06-19
refactor(session): per-connection state follow-ups (#277 #278 #279 #280)

### LINKED ISSUE #277: FOLLOW-UP: Rename Connection.Stores field to State (from PR #275)
## Context

This follow-up task was identified during the review of PR #275.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/275
**PR Title:** feat: per-connection state scoping (LiveView-style socket assigns)
**Suggested by:** Multiple claude[bot] reviews

## Task Description

The `Connection.Stores` field in `internal/session/registry.go` no longer reflects its purpose. In per-connection mode, `connState.state` is authoritative and `Stores` is only a snapshot updated in SharedState mode. The field name is a holdover from the pre-v0.7 Stores model.

Rename `Stores` to `State` across the codebase. A TODO comment already exists in the code.

## Original Comment

> The field name `Stores` no longer reflects its purpose in per-connection mode where `connState.state` is authoritative. This won't cause a bug, but it will confuse anyone reading the `Connection` struct in isolation. Worth a follow-up rename.

---

adnaan:
Bundles the four PR #275 per-connection-state review follow-ups.

## What changed

| # | Change | Where |
|---|--------|-------|
| #277 | Rename `Connection.Stores` → `State` (+ fix stale type-doc) | `internal/session/registry.go`, `mount.go` |
| #278 | Rename `SubscribeGroupActions` → `RegisterGroupActionHandler` (+ `SubscribeServerActions` → `RegisterServerActionHandler` for symmetry) | `pubsub/types.go`, `pubsub/redis.go`, `template.go`, tests |
| #279 | Document why the readChan buffer is intentionally fixed (knob declined) | `mount.go` |
| #280 | Add mock `MetricsRecorder` + test asserting `WSDispatchDropped()` fires on drop | `internal/session/dispatch_test.go` |

## Notes for review

- **#277**: the field held the per-connection typed `State` (used for peer fan-out), not the pre-v0.7 `Stores` map — the old type-doc (`livetemplate.Stores (map[string]Store)`) was stale and is corrected.
- **#278 — scope note**: the issue named only `SubscribeGroupActions`. I also renamed its sibling `SubscribeServerActions` → `RegisterServerActionHandler`. Both are one-time handler-registration methods called back-to-back in `template.go`; renaming only one would leave the pair inconsistently named (`RegisterGroupActionHandler` beside `SubscribeServerActions`), re-introducing the exact `Subscribe*`-vs-registration confusion #278 set out to fix. The per-entity, reference-counted `SubscribeToGroup`/`SubscribeToServerAction`/`SubscribeToGroupAction` methods are intentionally left untouched. This renames public `pubsub.Broadcaster`/`GroupActionBroadcaster` methods; no `Unreleased` CHANGELOG block exists to append to.
- **#279 — knob declined, documented instead**: the issue offered "make configurable *or at minimum document why increasing it does not help.*" A larger `readChan` buffer only changes *where* messages wait (Go channel vs OS socket buffer), not order or throughput, since the event loop applies them serially — so `WithReadBufferSize` would be an attractive nuisance. The existing `WithWebSocketBufferSize` is for the *send* path (slow clients), a different problem. Decision recorded in the code comment.

## Testing

- `/simplify` run pre-commit: clean (renames consistent across all layers, no leftovers).
- Pre-commit (gofmt, golangci-lint, full suite) green.

Closes #277, #278, #279, #280.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to LiveTemplate will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- **Breaking (pubsub):** renamed broadcaster handler-registration methods to use the `Register*Handler` convention, disambiguating one-time handler registration from the per-entity, reference-counted `SubscribeTo*` subscription methods. Implementers and callers of these `pubsub` interfaces must update:
+  - `Broadcaster.SubscribeServerActions` → `Broadcaster.RegisterServerActionHandler`
+  - `GroupActionBroadcaster.SubscribeGroupActions` → `GroupActionBroadcaster.RegisterGroupActionHandler`
+  - `TopicActionBroadcaster.SubscribeToTopicActions` → `TopicActionBroadcaster.RegisterTopicActionHandler`
+
 ## [v0.14.0] - 2026-06-13
 
 ### Changes
--- a/handle_test.go
+++ b/handle_test.go
@@ -2932,12 +2932,12 @@ func (f *fakeBroadcaster) Subscribe(_ pubsub.MessageHandler) error {
 	return f.subscribeErr
 }
 
-func (f *fakeBroadcaster) SubscribeServerActions(_ pubsub.ServerActionHandler) error {
+func (f *fakeBroadcaster) RegisterServerActionHandler(_ pubsub.ServerActionHandler) error {
 	f.serverActionCalls++
 	return nil
 }
 
-func (f *fakeBroadcaster) SubscribeGroupActions(_ pubsub.GroupActionHandler) error {
+func (f *fakeBroadcaster) RegisterGroupActionHandler(_ pubsub.GroupActionHandler) error {
 	f.groupActionCalls++
 	return nil
 }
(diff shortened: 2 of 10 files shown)


adnaan:
Addressed the review in 14747854:

- **`SubscribeToTopicActions` asymmetry** — agreed, renamed it to `RegisterTopicActionHandler`. It takes a handler (no entity id), so it's a one-time registration like the others, not a per-entity `SubscribeTo*` subscription. Completes the set: `Register{Server,Group,Topic}ActionHandler`. The per-entity reference-counted methods (`SubscribeToGroup`/`SubscribeToUser`/`SubscribeToServerAction`/`SubscribeToGroupAction`) are intentionally left as-is.
- **CHANGELOG** — added an `[Unreleased]` block (the repo drops it at release time, e.g. 615c112e) recording all three breaking pubsub renames with migration paths.
- **Drop-metric test** — added the `len(conn.DispatchChan) == 1` buffer-full precondition assertion before the dropped enqueue.

All three pubsub registration renames now share one convention; full suite + pre-commit green.

adnaan:
Thanks — confirming the verification point and converging on the cosmetics:

- **Package declaration**: `dispatch_test.go` is `package session` (white-box), so `conn.metrics` is reachable — the suite compiled and ran the new test, which empirically confirms it. ✅
- **Stale `Subscribe*` vocab**: grep for `SubscribeServerActions`/`SubscribeGroupActions`/`SubscribeToTopicActions` returns 0 across all `.go` files.
- **CHANGELOG header**: my new block uses `### Changed`, which is the correct [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) heading; the existing `v0.14.0` block's `### Changes` is the pre-existing drift. Leaving mine standard and not touching released history here.
- **`mockMetrics` doc comment**: keeping the one-line purpose doc — a brief, non-load-bearing note on a test double. Addressing this cosmetic point by reply rather than pushing, since this round surfaced no new functional issue (per the convergence guidance in CLAUDE.md).

No functional changes outstanding.

Issue #247, opened 2026-03-25
FOLLOW-UP: measureGCPressure should surface ExecuteUpdates errors


adnaan:
## Context

From PR #235 review. `measureGCPressure` in `systemcard_bench_test.go` silently ignores `ExecuteUpdates` errors in worker goroutines. If all workers error immediately, the GC pressure numbers are meaningless.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/235
**Suggested by:** @claude[bot]

## Task

Add error capture (similar to `measureUpdateThroughput`'s `firstErr` pattern) and panic/report if errors occur.

adnaan:
Resolved. `measureGCPressure` in `systemcard_bench_test.go` now captures the first worker error via an `atomic.Value` (`firstErr`), mirroring `measureUpdateThroughput`, and fails the run if any `ExecuteUpdates` errored. Closing as done.

Issue #245, opened 2026-03-25
FOLLOW-UP: WSIsUpgrade should check Sec-WebSocket-Key and Version headers


adnaan:
## Context

From PR #235 review. `WSIsUpgrade` checks Connection/Upgrade headers and GET method, but RFC 6455 also requires `Sec-WebSocket-Key` and `Sec-WebSocket-Version: 13`. Low impact since `Upgrade()` validates fully, but worth aligning with spec.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/235
**Suggested by:** @claude[bot], @copilot

adnaan:
Resolved. `WSIsUpgrade` (ws.go) now requires `Sec-WebSocket-Key` and `Sec-WebSocket-Version: 13` in addition to Connection/Upgrade/GET; `headerContainsValue` handles version lists. Covered by `ws_test.go` cases (missing/empty Key, missing/wrong Version). Closing as done.

Issue #244, opened 2026-03-25
FOLLOW-UP: Per-upgrader WriteBufferPool instead of package-level global


adnaan:
## Context

From PR #235 review. `defaultWriteBufferPool` is a package-level `*sync.Pool` shared across all `GorillaUpgrader` instances. If a user creates upgraders with different buffer sizes, the shared pool may return wrong-sized buffers.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/235
**Suggested by:** @claude[bot]

## Task

Move the write buffer pool to be per-upgrader instance, or enforce uniform buffer sizes when using a shared pool.

adnaan:
Resolved. `NewGorillaUpgrader` in `ws_gorilla.go` now assigns a per-upgrader `WriteBufferPool: &sync.Pool{}` (with a comment explaining upgraders with different `WriteBufferSize` never share buffers), `WithGorillaWriteBufferPool` allows an explicit override, and `Copy()` intentionally shares the pool only because it keeps the same `WriteBufferSize`. Covered by `TestGorillaUpgrader_PerInstanceWriteBufferPool` in `ws_test.go`. Closing as done.

PR #458, opened 2026-06-19
fix(ws): WebSocket upgrader follow-ups (#243 #244 #245 #247)

### LINKED ISSUE #243: FOLLOW-UP: Add compile-time assertion for WSConn interface sync
## Context

From PR #235 review. The `WSConn` interface is defined in two places (`ws.go` and `internal/session/registry.go`) to avoid circular imports. These must stay in sync.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/235

## Task

Add a compile-time assertion to ensure the two `WSConn` definitions stay compatible, or move the interface to a shared internal package (e.g., `internal/wsconn/`).

```go
// Option A: compile-time check in a test
var _ session.WSConn = (WSConn)(nil)

// Option B: shared internal package
```

adnaan:
Bundles the four PR #235 WebSocket-upgrader review follow-ups.

## What changed

| # | Change | Where |
|---|--------|-------|
| #243 | Bidirectional compile-time assertions that root `WSConn` ⇄ `session.WSConn` stay in sync | `ws.go` |
| #244 | Per-upgrader write-buffer `sync.Pool` instead of a package-level global | `ws_gorilla.go` |
| #245 | `WSIsUpgrade` fully validates the RFC 6455 handshake (adds `Sec-WebSocket-Key` + `Version: 13`) | `ws.go` |
| #247 | `measureGCPressure` surfaces `ExecuteUpdates` worker errors | `systemcard_bench_test.go` |

New `ws_test.go` (WSIsUpgrade table tests + per-upgrader pool test) and an `[Unreleased]` CHANGELOG entry.

## Notes for review

- **#243**: both assertions live in the root package because `internal/session` can't import root (import cycle) — root is the only package that can name both interfaces. Bidirectional (not the issue's one-liner) so drift in *either* direction breaks the build. The assertions compile today (interfaces are identical); they only fail on future drift.
- **#244**: `Copy()` shallow-copies `inner`, so copies share the pool pointer — safe, because a copy keeps the identical `WriteBufferSize` (only `SetCompression`/`SetCheckOrigin` mutate copies, never the size). `WithGorillaWriteBufferPool(nil/custom)` still overrides the per-instance default.
- **#245 — behavior note**: this is an exported-function behavior change (hence the CHANGELOG entry). A request carrying `Connection: upgrade` + `Upgrade: websocket` but missing `Sec-WebSocket-Key`/`Version` now routes to the plain-HTTP path instead of reaching `handleWebSocket` (where gorilla's `Upgrade()` would have returned 400). Real WebSocket clients always send these headers, so only malformed handshakes are affected. The two added header lookups are short-circuited behind the existing `Connection`/`Upgrade` checks, so plain HTTP traffic pays nothing.
- **#247**: mirrors the existing `measureUpdateThroughput` `firstErr`/panic pattern rather than extracting a shared helper — the two bench helpers differ in return type and surrounding measurement, and the inline duplication keeps each self-contained (per CLAUDE.md's anti-over-abstraction guidance).

## Testing

- New `TestWSIsUpgrade` (7 header cases + non-GET) and `TestGorillaUpgrader_PerInstanceWriteBufferPool` pass; existing `TestWebSocketDisabled_UpgradeRejected` (sends full headers) still green.

Closes #243, #244, #245, #247.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `Broadcaster.SubscribeServerActions` → `Broadcaster.RegisterServerActionHandler`
   - `GroupActionBroadcaster.SubscribeGroupActions` → `GroupActionBroadcaster.RegisterGroupActionHandler`
   - `TopicActionBroadcaster.SubscribeToTopicActions` → `TopicActionBroadcaster.RegisterTopicActionHandler`
+- `WSIsUpgrade` now fully validates the RFC 6455 handshake: in addition to the GET method and `Connection: upgrade` / `Upgrade: websocket` headers, it requires a non-empty `Sec-WebSocket-Key` and `Sec-WebSocket-Version: 13`. Requests missing these are routed as plain HTTP (they would have failed at `Upgrade()` anyway); well-behaved WebSocket clients always send them.
+
+### Fixed
+
+- `NewGorillaUpgrader` now gives each upgrader its own write-buffer `sync.Pool` instead of sharing a package-level global, so upgraders configured with different `WriteBufferSize` values can no longer draw mismatched buffers from a shared pool.
 
 ## [v0.14.0] - 2026-06-13
 
--- a/systemcard_bench_test.go
+++ b/systemcard_bench_test.go
@@ -606,6 +606,7 @@ func measureGCPressure(scenario appScenario, sessionCount int) gcResult {
 	var wg sync.WaitGroup
 	deadline := time.Now().Add(duration)
 
+	var firstErr atomic.Value
 	for idx := 0; idx < sessionCount; idx++ {
 		wg.Add(1)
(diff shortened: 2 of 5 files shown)


adnaan:
Addressed all three non-blocking items in 1d115398:

- **#244 Copy() invariant** — added a doc note on `Copy()` that the shallow copy intentionally shares the parent's `WriteBufferPool` pointer, safe because copies keep the same `WriteBufferSize` (only `SetCompression`/`SetCheckOrigin` mutate copies). Pins the hidden invariant.
- **#244 comment trim** — cut the per-upgrader pool comment from 6 lines to 2 (kept the load-bearing why, dropped the redundancy with the CHANGELOG/test).
- **#245 multi-value version test** — added `{"version list containing 13", ... "8, 13" ...}` to `TestWSIsUpgrade` documenting the RFC-valid comma-separated form `headerContainsValue` accepts. Passes.

No functional changes — all three are documentation/coverage. Full suite + pre-commit green.

Issue #459, opened 2026-06-19
Bare reference to an argument-requiring method should error (eval-node level), not stringify as a func value


adnaan:
## Context

Split out from #203 (parse/eval parity follow-up). #203 asked to improve `callMethod` behavior for non-zero-arg methods. Investigation showed `callMethod` is the **wrong altitude** for a behavioral fix — see the reasoning below — so #203 was resolved by documentation (its explicit option B) and the real fix is captured here.

## The gap

A **bare** reference to an argument-requiring method renders a confusing func representation instead of erroring like `text/template`:

- `{{.Get}}` where `Get(key string) string` → LiveTemplate stringifies the uncalled bound method (`fmt.Sprint` of a func value); Go errors with `wrong number of args for Get: want 1 got 0`.

## Why `callMethod` can't be the fix site

`resolveFieldChain` resolves (and for zero-arg methods, *calls*) each method **before** the caller knows whether trailing arguments follow. The uncalled-method return for arg-requiring methods is **load-bearing** for the with-args path:

- `{{.ctx.Get "key"}}` works precisely because `resolveFieldChain` yields the bound `Get` method as a value, which `callMethodOrFieldWithArgs` then invokes with `"key"` (covered by `TestStdlibParity_MethodCallOnMapValue`).

So any attempt to make `callMethod` error on `NumIn() != 0` breaks `{{.ctx.Get "key"}}`. Likewise, "call it if variadic" breaks `{{.Join "a"}}` for `Join(...string)` — `resolveFieldChain` would call `Join()` (empty), return a string, and `callMethodOrFieldWithArgs` then fails on the non-func. Both confirmed during the #203 investigation.

## Proposed fix (eval-node altitude)

The bare-vs-with-args distinction is known only at the eval nodes (`evalFieldNode` / `evalChainNode` / `evalVariableNode`), which see `len(args)`:

- **with-args branch** (`len(args) > 0 || pipeArg != sentinel`): keep today's behavior — resolve to the uncalled method, then `callMethodOrFieldWithArgs`.
- **bare-terminal branch**: a method that is zero-arg-callable (`NumIn()==0`, or variadic with `NumIn()==1`) should be *called*; a method that genuinely requires args should return a descriptive `ParseError` (`wrong number of args for X: want N got 0`) carrying the method name.

This also fixes the bare-variadic case (`{{.Tags}}` for `func(...string)` should call with no args) as a side effect.

## Tests to add

[...]

Issue #203, opened 2026-03-16
FOLLOW-UP: Improve callMethod behavior for non-zero-arg methods


adnaan:
## Context

This follow-up task was identified during the review of PR #199.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/199
**PR Title:** refactor: rewrite parse package with custom AST evaluator
**Suggested by:** @claude[bot]

## Task Description

`callMethod` in `eval.go` returns the method itself (as a callable value) when the method has non-zero arguments. This is used by `resolveFieldChain` to support method-as-value patterns, but could produce confusing downstream errors if the method value ends up in `valueToString` (which would call `fmt.Sprint` on it).

Consider returning a `ParseError` with a descriptive message instead of silently returning the method as a value, or document the expected call pattern more clearly.

## Original Comment

> `resolveFieldChain` is supposed to call zero-arg methods automatically. For a method that takes arguments, returning the `reflect.Value` of the method is likely to cause confusing downstream errors. Consider returning `nil, &ParseError{...}` with a clearer message rather than silently returning the method as a value.

---
*This issue was automatically created by prmonitor from PR review comments.*

PR #460, opened 2026-06-19
fix(parse): 3-index slice support + document callMethod arg behavior (#202 #203)

### LINKED ISSUE #202: FOLLOW-UP: Add 3-index slice support to builtinSlice
## Context

This follow-up task was identified during the review of PR #199.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/199
**PR Title:** refactor: rewrite parse package with custom AST evaluator

## Task Description

Go's `html/template` `slice` builtin supports `slice x i j k` (three indices, for `x[i:j:k]` — slice with capacity). The current `builtinSlice` implementation in `eval.go` only handles 0, 1, or 2 indices. This is a minor parity gap.

## Original Comment

> Go's `html/template` `slice` builtin supports `slice x i j k` (three indices, for `x[i:j:k]`). The implementation only handles 0, 1, or 2 indices. This is a minor parity gap worth noting even if not blocking.

---

adnaan:
Bundles the two PR #199 parse/eval parity follow-ups.

## #202 — 3-index slice support (real fix)

`builtinSlice` now handles the 3-index form `slice x i j k` → `x[i:j:k]` (capacity bound), matching `text/template`. A 3-index slice of a **string** is rejected with `cannot 3-index slice a string`, because `reflect.Slice3` only works on slices/arrays (the existing 2-index `reflect.Slice` path handles strings fine).

Tests: unit tests for capacity / string-error / too-many-indices, plus a `stdlib` parity test — including `{{slice (slice .S 0 2 5) 0 4}}`, which only renders correctly if the capacity bound was actually applied.

## #203 — documented (option B); real fix tracked in #459

The issue offered "return a `ParseError` **or** document the expected call pattern more clearly." The error path was found to break a legitimate, Go-parity pattern, so this PR takes the documentation option and files the real fix separately.

**Why not error in `callMethod`:** returning an argument-requiring method *uncalled* is load-bearing. `resolveFieldChain` resolves a chain **before** the caller knows whether trailing args follow, so it must not call a method that might still receive them. Returning the bound method is exactly what makes `{{.ctx.Get "key"}}` work — `resolveFieldChain` yields the bound `Get`, then `callMethodOrFieldWithArgs` applies `"key"` (covered by `TestStdlibParity_MethodCallOnMapValue`). Making `callMethod` error breaks that; making it *call* variadic methods breaks `{{.Join "a"}}` for `func(...string)` the same way.

**The real gap** — a *bare* `{{.Get}}` (no args) to an arg-requiring method should error like `text/template` instead of stringifying a func value — must be fixed at the **eval-node level**, where arg presence is known (`callMethod` only receives a `reflect.Value`, no AST context). Filed as **#459** with the full reasoning and test plan. The `callMethod` doc comment records the decision and points at #459.

## Testing

Closes #202, #203.

### DIFF
--- a/internal/parse/eval.go
+++ b/internal/parse/eval.go
@@ -352,20 +352,12 @@ func resolveFieldChain(value interface{}, fields []string) (interface{}, error) 
 	return value, nil
 }
 
-// callMethod calls a zero-argument method and returns its result.
-//
-// A method that requires arguments is returned UNCALLED (as a callable reflect
-// value), not invoked here. This is deliberate and load-bearing: resolveFieldChain
-// runs before the caller knows whether trailing arguments follow, so it must not
-// call a method that might still receive args. Returning the bound method is what
-// makes a chain like {{.ctx.Get "key"}} work — resolveFieldChain yields the bound
-// Get method, then callMethodOrFieldWithArgs applies "key".
-//
-// Caveat (tracked in #459): a *bare* reference to an argument-requiring method
-// (e.g. {{.Get}} with no trailing args) yields the uncalled method value, which
-// then stringifies to a confusing func representation instead of erroring like
-// text/template. A proper fix must distinguish the bare-terminal case at the
-// eval-node level, where the presence of args is known — callMethod cannot tell.
+// callMethod calls a zero-argument method and returns its result. A method that
+// requires arguments is returned UNCALLED so the caller can still apply trailing
+// args (e.g. the {{.ctx.Get "key"}} chain): resolveFieldChain resolves the method
+// before arg presence is known, so it must not invoke one that may receive args.
+// A bare reference to an arg-requiring method therefore yields the uncalled value
+// rather than erroring like text/template; the eval-node-level fix is tracked in #459.
 func callMethod(method reflect.Value) (interface{}, error) {
 	mt := method.Type()
 	if mt.NumIn() != 0 {
--- a/internal/parse/eval.go
+++ b/internal/parse/eval.go
@@ -369,7 +369,7 @@ func resolveFieldChain(value interface{}, fields []string) (interface{}, error) 
 func callMethod(method reflect.Value) (interface{}, error) {
 	mt := method.Type()
 	if mt.NumIn() != 0 {
-		return method.Interface(), nil // Return the method itself (it's a callable)
+		return method.Interface(), nil
 	}
 	results := method.Call(nil)
 	if len(results) == 0 {
--- a/internal/parse/eval.go
+++ b/internal/parse/eval.go
@@ -353,6 +353,19 @@ func resolveFieldChain(value interface{}, fields []string) (interface{}, error) 
 }
 
 // callMethod calls a zero-argument method and returns its result.
+//
+// A method that requires arguments is returned UNCALLED (as a callable reflect
+// value), not invoked here. This is deliberate and load-bearing: resolveFieldChain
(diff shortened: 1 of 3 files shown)


adnaan:
Addressed in fd397804:

- **String-3-index parity (wiring)** — added `TestSliceThreeIndexString_ErrorsViaTemplate`, which drives `{{slice .S 1 3 4}}` (S a string) through the real `Parse` + `BuildTree` path and asserts the `cannot 3-index slice a string` error. The existing unit test covered `builtinSlice` directly; this confirms the guard is reachable end-to-end.
- **Redundant inline comment** — removed the trailing `// Return the method itself` on `callMethod`'s return; the doc block already covers it.

**Declined: the array (`[6]int`) test case.** Verified empirically that `builtinSlice([6]int{...}, 1, 3, 5)` **panics** — `reflect.Value.Slice3: slice of unaddressable array`. A by-value array passed as `interface{}` is unaddressable, and `reflect.Slice3` (like the `reflect.Slice` used in the existing 1-/2-index cases) panics on unaddressable arrays. So arrays-by-value don't work in *any* arity here — a happy-path array test would panic rather than document "arrays work." `case 3` deliberately mirrors cases 1/2, including this shared limitation. Happy to track addressable-array support separately if it's wanted, but it'd be a change to all three cases, out of scope for this 3-index parity gap.

No functional changes outstanding.

adnaan:
Thanks — addressed in e5589254:

**1. callMethod comment length** — tightened from 14 lines to 6 (dropped the step-by-step walkthrough, which lives in this PR description / #459). Kept the load-bearing invariant (arg-requiring methods returned uncalled so `{{.ctx.Get "key"}}` can apply trailing args) and the #459 anchor.

A note on the cited rule: I couldn't find "never write multi-paragraph docstrings / one short line max" in CLAUDE.md — the actual comment rule is line 99, *"No unnecessary comments — Code should be self-documenting,"* and the bot-workflow section explicitly endorses comments that record a deliberate deviation ("the comment is the deviation record"). This invariant is genuinely regression-prone (an earlier attempt to make `callMethod` error here broke `{{.ctx.Get "key"}}`), so I kept a few lines of *why* rather than collapsing to one line. Round 1 of this PR and the pre-merge pass both read the same comment as appropriate.

**2. General type guard (maps/ints/funcs)** — declining as out of scope: the 0/1/2-index cases all use `reflect.Slice`, which has the identical panic-on-non-sliceable gap. It's pre-existing and consistent, not a regression introduced here. Worth its own issue if a general guard is wanted across all arities.

**3. Array (`[6]int`) test** — re-confirming the decline with evidence: `builtinSlice([6]int{...}, 1, 3, 5)` **panics** (`reflect.Value.Slice3: slice of unaddressable array`). A by-value array passed as `interface{}` is unaddressable, and `Slice3` — like the `Slice` in cases 1/2 — panics on unaddressable arrays. So a happy-path array test would panic rather than pass; arrays-by-value don't work in any arity here.

adnaan:
Converged — thanks. The three points here are the same non-blocking nits from the prior round, all already addressed or declined with rationale above: the comment is trimmed to the load-bearing why + #459 (you note it's justified and non-blocking), the wider type guard is a pre-existing gap shared by the 0/1/2-index cases (out of scope for the 3-index parity fix), and the by-value array case panics in `Slice3` (so a happy-path array test can't pass). No further code changes — holding for final CI.

Issue #305, opened 2026-03-31
Make ephemeralSweepTTL configurable via WithEphemeralSweepTTL()


adnaan:
## Context

Follow-up from PR #301 review. The HTTP template cache TTL in ephemeral mode is hardcoded at 30 minutes.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/301

## Problem

`ephemeralSweepTTL = 30 * time.Minute` is a reasonable default, but apps with different usage patterns may need to tune it:
- High-frequency navigation apps may benefit from shorter TTL
- Internal dashboards with slow interaction may need longer TTL

## Suggested API

```go
handler := tmpl.Handle(ctrl, livetemplate.AsState(&State{}),
    livetemplate.WithEphemeralState(),
    livetemplate.WithEphemeralSweepTTL(10 * time.Minute),
)
```

Add `ephemeralSweepTTL time.Duration` to `handleConfig`, default to 30 minutes when zero.

---
*Follow-up from PR #301 bot review.*

adnaan:
Resolved. `WithEphemeralSweepTTL()` is implemented (wired via `handleConfig.ephemeralSweepTTL`, defaults to `defaultEphemeralSweepTTL` when unset/zero/negative) and covered by `TestWithEphemeralSweepTTL` in `handle_test.go`. Closing as done.

PR #464, opened 2026-06-20
feat(mount): WithEphemeralSweepTTL + persistent Mount-on-POST test (#304 #305)

### LINKED ISSUE #304: Add test: persistent-mode Mount-on-POST receives persisted state
## Context

Follow-up from PR #301 review. Mount() now runs on HTTP POST, but there's no test verifying the ordering: load from SessionStore → Mount() → action.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/301

## Task

Add a test that:
1. GET renders page (Mount sets initial state, persisted)
2. POST increments Count (persisted as Count=1)
3. Second POST — verify Mount() receives Count=1 from the store (not a fresh clone with Count=0)

This prevents regressions on the state-loading → Mount → action ordering in persistent mode.

---
*Follow-up from PR #301 bot review.*

adnaan:
Bundles the two PR #301 follow-ups.

## #305 — `WithEphemeralSweepTTL` (new option)

Makes the ephemeral HTTP-template-cache sweep TTL configurable:

```go
tmpl.Handle(ctrl, livetemplate.AsState(&State{}),
    livetemplate.WithEphemeralSweepTTL(10*time.Minute))
```

- Defaults to **30 minutes** when unset or non-positive.
- **No effect in persistent mode** (state with `lvt:"persist"` fields) — there, eviction follows the SessionStore, not a TTL.
- A short TTL also **tightens the sweep interval**: `sweepInterval()` caps the ticker to the TTL (floored at 1m) in ephemeral mode, so a 2m TTL is actually honored rather than rounded up to the 10m default. Persistent mode keeps the 10m interval (its eviction is `SessionStore.List`-driven, and a tighter interval would only add load).
- Renames the const `ephemeralSweepTTL` → `defaultEphemeralSweepTTL`; extracts `defaultSweepInterval` / `minSweepInterval`.

**Note on the issue's example:** #305 shows `WithEphemeralState()`, which **no longer exists** — ephemeral mode is now automatic when the state has no `lvt:"persist"` fields. This PR adds only `WithEphemeralSweepTTL`.

## #304 — persistent-mode Mount-on-POST ordering test

`TestPersistentMode_MountOnPostReceivesPersistedState` uses a `Mount`-recording controller to verify the load → Mount → action ordering: on the **second** POST, `Mount` receives the persisted `Count=1`, not a fresh clone with `Count=0`. The assertion captures the Mount-call index *before* the second POST and checks that specific call (robust to extra Mount calls), with an end-to-end `Count: 2` backstop.

## Tests

- `TestPersistentMode_MountOnPostReceivesPersistedState` — the ordering invariant.
- `TestWithEphemeralSweepTTL` — option resolution (default / custom / zero / negative → default).
- `TestSweepInterval` — interval capping/flooring (persistent keeps default; ephemeral long-TTL keeps default; short-TTL shrinks to TTL; tiny-TTL floored).
- `go test ./...` passes; `/simplify` clean across reuse/simplification/efficiency/altitude.

Closes #304, #305.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `WithEphemeralSweepTTL(ttl)` `HandleOption` to tune how long idle HTTP template cache entries survive in ephemeral mode (state with no `lvt:"persist"` fields) before the sweep loop evicts them. Defaults to 30 minutes; no effect in persistent mode (eviction follows the SessionStore there). A short TTL also tightens the sweep interval (floored at 1 minute) so the value is actually honored.
+
 ### Changed
 
 - **Breaking (pubsub):** renamed broadcaster handler-registration methods to use the `Register*Handler` convention, disambiguating one-time handler registration from the per-entity, reference-counted `SubscribeTo*` subscription methods. Implementers and callers of these `pubsub` interfaces must update:
--- a/handle_test.go
+++ b/handle_test.go
@@ -784,6 +784,136 @@ func TestWebSocketDisabled_StatePersistsAcrossRequests(t *testing.T) {
 	}
 }
 
+// mountObsState/mountObsController record the Count that Mount observes on each
+// request, to verify the persistent-mode HTTP ordering: load from SessionStore →
+// Mount → action.
+type mountObsState struct {
+	Count int `lvt:"persist"`
+}
+
+type mountObsController struct {
+	mountCounts []int // Count seen at the start of each Mount (calls are sequential in tests)
+}
+
+func (c *mountObsController) Mount(s mountObsState, _ *Context) (mountObsState, error) {
+	c.mountCounts = append(c.mountCounts, s.Count)
+	return s, nil
+}
+
+func (c *mountObsController) Increment(s mountObsState, _ *Context) (mountObsState, error) {
+	s.Count++
+	return s, nil
+}
+
+func TestPersistentMode_MountOnPostReceivesPersistedState(t *testing.T) {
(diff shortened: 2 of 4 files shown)


adnaan:
Thanks — addressed the two actionable nits in 925a8eb7:
- `minSweepInterval`: `1 * time.Minute` → `time.Minute`.
- `liveHandler.ephemeralSweepTTL` comment: now states the invariant ("always positive; defaults to defaultEphemeralSweepTTL") instead of naming the resolution site, per the don't-reference-callers guidance.

The third observation (persistent-mode `ttl` param unused in the `!ephemeral` branch) is intentional — persistent eviction is SessionStore-driven, so the TTL genuinely doesn't apply there. Leaving as-is.

adnaan:
Thanks — converging here, as both review rounds report no functional issues and "ready to merge." On the three refinements:

1. **Comment style** — the `ephemeralSweepTTL` field comment starting with the identifier name is the idiomatic Go doc convention (golint/`go doc` expect it), not redundancy. The `mountObs*` block comment documents the types' *purpose* (recording observed Count to verify the load→Mount→action ordering), which the names alone don't convey — keeping it.
2. **At-threshold `sweepInterval` case** — fair point, but the `ttl >= defaultSweepInterval` boundary is trivially correct and already bracketed by the `3m → 3m` and `30m → default` rows; declining the extra at-exactly-10m row to avoid another full review/CI cycle for marginal coverage.
3. **Persistent-mode interval invariant** — already machine-checked at the unit level by `TestSweepInterval`'s "persistent keeps default" case (`sweepInterval(false, 2m) == defaultSweepInterval`); an end-to-end `WithStore + WithEphemeralSweepTTL` variant would be redundant with that.

No further code changes. Holding for the `claude-review` check to settle green.

Issue #239, opened 2026-03-25
FOLLOW-UP: Support formnovalidate server-side in ValidateForm


adnaan:
## Context

**Source PR:** https://github.com/livetemplate/livetemplate/pull/233
**Suggested by:** @claude-bot
**Priority:** Medium

## Task Description

The guide shows `<button name="save-draft" formnovalidate>` as a way to skip validation. But `ValidateForm()` runs regardless of which button was clicked — the `formnovalidate` intent is silently lost.

**Fix:** The client should send a flag (e.g., `_formnovalidate: true`) when the submitter had the `formnovalidate` attribute. `ValidateForm()` should check this flag and skip validation.

---
*This issue was automatically created from PR #233 review comments.*

PR #465, opened 2026-06-20
refactor(formvalidation): unify FormRule optional-bound sentinels (#241)

### LINKED ISSUE #241: FOLLOW-UP: Consistent sentinel design in FormRule struct
## Context

**Suggested by:** @claude-bot
**Priority:** Low

## Task Description

`FormRule` uses mixed sentinel patterns:
- `MinLength`/`MaxLength` use `-1` to mean "not set"
- `Min`/`Max` use `HasMin`/`HasMax` booleans

Use one pattern consistently — either `*int`/`*float64` pointers, or `Has*` booleans for all fields.

---

adnaan:
Implements **#241**. The sibling **#239** is intentionally out of scope — see note below.

## #241 — consistent sentinel design in `FormRule`

`FormRule` mixed two "not set" conventions:
- `MinLength`/`MaxLength`: `-1` sentinel
- `Min`/`Max`: `HasMin`/`HasMax` booleans

Unified on the `Has*` pattern already present in the struct: dropped the `-1` sentinel and added `HasMinLength`/`HasMaxLength`, so all four numeric bounds read uniformly. (The zero value is a *valid* bound, so an int sentinel was unsound to begin with.)

**Why `Has*` and not pointers** (the issue's other option): the struct already used `HasMin`/`HasMax`, so extending that follows the existing pattern (CLAUDE.md) and keeps reads as plain value comparisons. Pointers would force nil-deref at every read site and either leave the struct half-pointer/half-bool or churn `Min`/`Max` too.

- `ExtractFormSchema` sets the paired flag where it parses each value; the construction collapses to `FormRule{Field: name}`.
- The eligibility gate and `Validate`'s length checks switch from `>= 0` to the flag.
- All read/write sites switched in lockstep; no `-1`/`>= 0` sentinel references remain.

**Breaking:** `FormRule` is exported, so code constructing it with `MinLength: -1` or reading the `-1` sentinel must switch to the `Has*` flags. CHANGELOG `### Changed` entry added. Callers using `ExtractFormSchema`/`ValidateForm` normally are unaffected. (Pre-1.0/alpha, so no deprecation shim.)

## #239 — scoped separately (not in this PR)

#239 (server-side `formnovalidate` in `ValidateForm`) is a **cross-repo** change: the `livetemplate/client` must send a `_formnovalidate` flag for the submitter, the server checks it, plus docs/example + a client release. It's being handled as its own effort rather than bundled with this server-only refactor. #239 stays open.

## Testing

- `go test ./...` passes; form-validation suite updated for the new flags (including a strengthened `ExtractFormSchema` assertion that checks `HasMinLength`/`HasMaxLength`).

Closes #241.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Breaking (`FormRule`):** unified the optional-bound sentinels. `MinLength`/`MaxLength` no longer use `-1` to mean "not set"; all four numeric bounds are now gated by paired `Has*` booleans — new `HasMinLength`/`HasMaxLength` join the existing `HasMin`/`HasMax`. Code that constructed a `FormRule` with `MinLength: -1` (or read it expecting the `-1` sentinel) must switch to the `Has*` flags. Callers using `ExtractFormSchema`/`ValidateForm` normally are unaffected.
 - **Breaking (pubsub):** renamed broadcaster handler-registration methods to use the `Register*Handler` convention, disambiguating one-time handler registration from the per-entity, reference-counted `SubscribeTo*` subscription methods. Implementers and callers of these `pubsub` interfaces must update:
   - `Broadcaster.SubscribeServerActions` → `Broadcaster.RegisterServerActionHandler`
   - `GroupActionBroadcaster.SubscribeGroupActions` → `GroupActionBroadcaster.RegisterGroupActionHandler`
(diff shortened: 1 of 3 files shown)


adnaan:
Thanks — converging, as the review is LGTM with no functional issues. On the observations: (1) the Has* grouping mirrors the existing HasMin/HasMax layout (deliberate, per follow-existing-patterns); (2) minlength="0" behaves identically before/after; (3) the "no length flags → no length enforcement" invariant is implicitly covered by the Required/Email cases (rules with no length fields, no length errors expected) — leaving it there rather than adding a dedicated negative test, to avoid a churn cycle for marginal coverage. No further code changes.

PR #469, opened 2026-06-26
fix(render): pass dynamic HTML through verbatim, drop CSS-blind minify (#467)

### LINKED ISSUE #467: bug(render): dynamic-content minify collapses CSS-driven white-space:pre-wrap (corrupts <span class="chroma"> and similar)
## Summary

`internal/render.MinifyHTML` (tdewolff/minify) is applied to **dynamic** values, and it collapses whitespace inside elements that are whitespace-significant **via CSS** rather than via HTML tag name — silently corrupting rendered output for things like syntax-highlighted code, diffs, and ASCII art.

tdewolff is HTML-*tag*-aware: it correctly preserves `<pre>`/`<textarea>` because it knows those tags by name. But it cannot see stylesheets, so it has no idea that `<span class="chroma">` (or any `white-space: pre-wrap`/`pre` applied via a class) is whitespace-significant, and it flattens the content.

## Where

- `internal/build/html_segmentation.go:91` — `dyn = render.MinifyHTML(strDyn)` (per-segment dynamic)
- `internal/build/html_segmentation.go:104` — `fallback.SetDynamic(0, render.MinifyHTML(htmlDoc))` (single-segment fallback)
- `internal/build/html_diff.go:50` — `tree.SetDynamic(0, render.MinifyHTML(dynamicPart))`

These are in `CreateHTMLStructureBasedTree` (documented as the fallback used "when template parsing fails or for initial tree generation without template metadata") and the diff path. The **primary** parse-based path (`internal/parse/flatten.go`, `buf.Write(n.Text)`) preserves statics verbatim and is unaffected — so this only bites content that flows through the minified dynamic/fallback paths.

## Verified repro

Run inside `internal/render`:

```go
func TestPreWrapMinify(t *testing.T) {
	for _, in := range []string{
		`<span class="chroma">  indented   code  </span>`, // CSS pre-wrap — significant
		`<pre>  indented   code  </pre>`,                   // HTML pre — tag-known
		`<code>a</code> <small>b</small>`,                 // inter-inline space
	} {
		t.Logf("IN : %q\nOUT: %q", in, MinifyHTML(in))
	}
}
```

Output:

```
IN : "<span class=\"chroma\">  indented   code  </span>"
OUT: "<span class=chroma>indented code</span>"        <-- whitespace COLLAPSED ✗

IN : "<pre>  indented   code  </pre>"
OUT: "<pre>  indented   code  </pre>"                  <-- preserved ✓ (tag-known)

IN : "<code>a</code> <small>b</small>"
OUT: "<code>a</code> <small>b</small>"                 <-- preserved ✓
```

## Impact

Any app rendering CSS-`pre-wrap` content as a dynamic value (chroma/highlighted code, a diff viewer, preformatted-via-class blocks) through the fallback/diff path gets its whitespace silently flattened. Found while working on `livetemplate/prereview` (a diff reviewer whose `.content`/`.chroma` are `pre-wrap` via class); prereview avoids it only because it uses the primary parse-based path.

## Suggested direction

Minification of dynamic HTML is fundamentally unsafe without CSS knowledge. Options, roughly in order of safety:

1. Don't minify dynamic HTML values at all (preserve verbatim, like statics).
2. Make minify opt-in (`WithDynamicMinify()`, off by default).
3. At minimum, document that the dynamic-minify path is CSS-blind and will corrupt class-driven `pre-wrap`/`pre` content.

adnaan:
Closes #467.

## Problem

`internal/render.MinifyHTML` (tdewolff/minify) was applied to **dynamic** tree values. tdewolff is HTML-*tag*-aware — it preserves `<pre>`/`<textarea>` because it knows those tags — but it cannot see stylesheets, so it silently collapsed whitespace inside elements made whitespace-significant by a CSS **class** (e.g. `white-space: pre-wrap` on `<span class="chroma">`). This corrupted syntax-highlighted code, diffs, and ASCII art rendered through the structure-based and diff tree paths.

Statics were already emitted verbatim; only dynamics were minified, so the behavior was inconsistent as well as lossy.

## Fix

Pass dynamic HTML through verbatim, exactly like statics. The three `render.MinifyHTML(...)` call sites in `internal/build` (`html_segmentation.go` ×2, `html_diff.go` ×1) were the only non-test callers, so this orphaned `MinifyHTML`/`normalizeWhitespace`/`getMinifier` — `internal/render/minify.go` is deleted and the `github.com/tdewolff/minify/v2` dependency dropped via `go mod tidy`.

No wire/client impact: statics were already un-minified, so dynamic payloads just carry the whitespace the client already handles.

## Tests

- New `internal/build` regression tests assert both surviving paths (`CreateHTMLStructureBasedTree` fallback + `AnalyzeChangeAndCreateTree` diff) preserve whitespace-significant dynamic content verbatim.
- Full suite green; no golden/assertion relied on minified dynamics.

## Docs

Updated `CHANGELOG.md` and the doc references to the now-removed `minify.go`/`MinifyHTML` (CODE_STRUCTURE, ARCHITECTURE, CONTRIBUTING, performance-characteristics, new-contributor-walkthrough).


adnaan:
Addressed both review findings in bc4ea93:

1. **Multi-segment path coverage (CONFIRMED)** — added `TestCreateHTMLStructureBasedTree_MultiSegmentPreservesWhitespace`, which feeds real block-level `<div>`s (spaced beyond `segmentSize`) with whitespace-significant content in the final dynamic segment, and guards with a `>2 statics` check so it keeps exercising the block-tag loop rather than silently falling back.
2. **Text-only normalization removal (PLAUSIBLE)** — documented in the comment blocks of both `html_segmentation.go` and `html_diff.go` that text-only dynamics are now passed verbatim too, since the former `normalizeWhitespace` side-effect was equally unsafe for whitespace-significant content.

Note: the earlier "Test LVT against Core Changes" failure was an unrelated transient — `TestCompleteWorkflow_BlogApp` panicked fetching the client library from the unpkg CDN (`context deadline exceeded`), not a regression from this change. This push re-triggers CI.

adnaan:
Addressed all three round-2 findings in c6574e2:

1. **Stale tdewolff indirect deps in `e2e/docker/app`** — ran `go mod tidy` in the submodule; both `tdewolff/minify` and `tdewolff/parse` are gone from its `go.mod`/`go.sum`.
2. **Stale `BenchmarkMinifyHTML` baseline** — removed the `BenchmarkMinifyHTML-8` line from `testdata/benchmarks/baseline.txt` so benchstat no longer reports it as removed.
3. **TrimSpace-tolerant assertion** — `TestAnalyzeChangeAndCreateTree_PartialChangeKeepsStatics` now asserts the exact verbatim dynamic `" World"` (with leading space from the boundary split) and drops the stale "normalized" wording, making it a precise guard.

adnaan:
Round 4 — thanks, no correctness issues. On the two observations:

1. **`e2e/docker/app` go directive 1.24 → 1.26.0** — I tried reverting it to keep the diff minimal, but `go build ./...` in the submodule then fails with `go: updates to go.mod needed` — a transitive dependency now requires the higher directive, which is why `go mod tidy` bumped it. So the bump is necessary (and consistent with the root module's `go 1.26.0`), not just cosmetic. Keeping it.
2. **Rotting `— see #467` back-references** — fixed in 32ff78a: the comments now state the invariant directly (a tag-aware minifier can't see CSS, so it collapses whitespace made significant by `white-space:pre-wrap`), and the stale cross-file location hint on the diff-path test is dropped.

adnaan:
Round 5 — addressed in e90be91 (all CLAUDE.md style):

1. Trimmed both verbatim-passthrough comments to the invariant only ("a tag-aware minifier cannot see CSS and will collapse whitespace that white-space:pre-wrap makes significant"); dropped the failure-mode narrative and the fallback WHAT-comment.
2. Removed the `See #467` references from test source.
3. Dropped the WHAT-restating test doc comments; kept only the `preWrapDynamic` invariant sentence (trimmed of its issue ref).

adnaan:
Round 6 — addressed in 8c09dee:

1. **CHANGELOG precision** — reworded to "Dynamic content (HTML **and** plain text) is no longer minified or whitespace-normalized," noting that the removed `MinifyHTML` also trimmed/collapsed text-only dynamics. Good catch.
2. **`DynamicLen()-1` as index** — switched to `len(tree.Dynamics)-1` (the real last slot).
3. **Cross-file `preWrapDynamic` comment** — respectfully declining: round 4 of this same review explicitly asked to *drop* the cross-file location hint ("IDE jump-to-definition handles the latter"), and adding it back now would re-introduce exactly what was removed. `preWrapDynamic` lives in `html_segmentation_test.go` next to the function that's the primary subject of these tests; jump-to-definition resolves it.

I'm treating the review as converged here — the last three rounds found no correctness issues (and this round's nit #3 reverses round 4's), so I'll handle any further phrasing-level feedback by reply rather than additional pushes. Ready for maintainer sign-off.

adnaan:
Round 7 — all three are style/consistency (you confirm no correctness issues), and this round reverses earlier rounds, so I'm treating the review as converged and replying rather than pushing further (per CLAUDE.md "AI Code Review Workflow": stop when successive rounds surface no new functional issue).

1. **Collapse the two-line comment** — this is the *exact* two-line wording round 4/6 of this review asked me to adopt, now flagged as too long. I checked: the cited rule ("Never write multi-paragraph docstrings or multi-line comment blocks — one short line max") is **not** in this repo's CLAUDE.md; the actual rule is *"No unnecessary comments – code should be self-documenting"* (CLAUDE.md §Code Style). The current two lines are concise and state a non-obvious invariant (a tag-aware minifier can't see CSS), so I'm leaving them to avoid churning the same comment a third time.
2. **Comment placement** — it sits at the top of the build block that replaced the `MinifyHTML` call, documenting why dynamics are now passed through untouched; that's the right scope for it.
3. **`e2e/docker/go.mod` at `go 1.24`** — that's a separate module this PR never modifies. The `e2e/docker/app/go.mod` bump to 1.26.0 was a forced `go mod tidy` side-effect (a transitive dep requires it; reverting it breaks `go build` there), not a deliberate alignment, so bumping the unrelated browser-test module is out of scope here.

Rounds 4–7 have found no functional issues (round 7 reverses round 6), so this is the convergence point. Ready for maintainer sign-off / merge.

PR #470, opened 2026-06-26
fix(parse): strip HTML comments to match html/template (#468)

### LINKED ISSUE #468: bug(parse): HTML comments (<!-- -->) are emitted to output, diverging from html/template (which strips them)
## Summary

`html/template` strips HTML comments (`<!-- ... -->`) from output during its **escape pass** (see [this well-known behavior](https://stackoverflow.com/questions/34348072/go-html-comments-are-not-rendered)). livetemplate parses via `html/template` but renders by walking the **raw parsed tree** (`internal/parse/flatten.go`, `buf.Write(n.Text)`) and never triggers the escape pass — so `<!-- ... -->` comments survive verbatim into the static segments and **ship to the browser**.

Template comments `{{/* ... */}}` are correctly stripped (that happens at parse time, independent of escaping).

## Verified repro

```go
func TestCommentStripping(t *testing.T) {
	const src = `<div><!-- SECRET-HTML-COMMENT -->{{.X}}{{/* SECRET-TMPL-COMMENT */}}</div>`
	tmpl := Must(New("cmt"))
	tmpl.Parse(src)
	var buf bytes.Buffer
	tmpl.ExecuteUpdates(&buf, map[string]any{"X": "hi"})
	t.Log(buf.String())
}
```

Output payload:

```
{"0":"hi","s":["<div><!-- SECRET-HTML-COMMENT -->","</div>"]}
```

- `<!-- SECRET-HTML-COMMENT -->` → **present in `s[0]`** (emitted to the client) ✗
- `{{/* SECRET-TMPL-COMMENT */}}` → **absent** (stripped) ✓

For contrast, the same template through plain `html/template.Execute` drops the HTML comment.

## Impact

1. **Leak.** Developer/internal HTML comments reach the client and show up in view-source. (Found this while reviewing `livetemplate/prereview`, whose `prereview.tmpl` has two `<!-- -->` notes currently shipping to users.)
2. **Security / migration surprise.** `html/template` strips comments partly as hardening — IE conditional comments (`<!--[if ...]>`) are a known injection vector, and stripping avoids accidental data exposure. An app migrating from `html/template` to livetemplate silently *starts* emitting comments it used to drop, with no signal.

## Suggested direction

1. Strip `<!-- ... -->` comments during tree flattening to match `html/template` output semantics (preferred — least-surprise for anyone coming from `html/template`); **or**
2. Document the divergence explicitly and provide an opt-out/opt-in, with the security rationale called out.

Note the interaction with the verbatim-statics philosophy: stripping comments is a deliberate, targeted exception (`html/template` itself makes it), distinct from general whitespace, which must stay verbatim.

adnaan:
Closes #468.

## Problem

`html/template` strips HTML comments (`<!-- ... -->`) during its escape pass. LiveTemplate parses via `html/template` but builds its static segments by walking the **raw** parse tree (`internal/parse/flatten.go`), which never triggers that pass — so HTML comments survived verbatim into the statics and shipped to the client (visible in view-source) and over WebSocket tree updates. Template comments (`{{/* ... */}}`) were already stripped at parse time.

This is both a **leak** (internal comments reach the client) and a **migration surprise** (an app moving from `html/template` silently starts emitting comments it used to drop — `html/template` strips them partly as hardening against IE conditional-comment injection).

## Fix

Strip HTML comments at the source level in `parseInternal`, before bracket expansion (so directives inside soon-to-be-deleted comments aren't expanded). This single point feeds both the HTTP render path (`t.tmpl`) and the WS tree path (`t.templateStr`), so they stay consistent.

`render.StripHTMLComments` uses the `x/net/html` tokenizer rather than a regex, which makes it context-aware:
- comment-like text inside attribute values (`title="<!-- x -->"`) is **preserved**;
- comment markup inside `<script>`/`<style>`/`<textarea>` (RAWTEXT/RCDATA) is **left verbatim**;
- a comment wrapping an action (`<!-- {{.X}} -->`) is removed in full, matching `html/template`.

Reconstruction concatenates each token's raw bytes except comment tokens, so `{{...}}` actions and all other content (including `<!DOCTYPE html>`) round-trip losslessly. A `strings.Contains(s, "<!--")` fast path skips tokenization for comment-free templates.

## Residual divergences (documented, pre-existing — not regressions)

In `docs/references/template-support-matrix.md`: `<script>` (html/template rejects `<!--` outright) and `<textarea>`/RCDATA (html/template escapes `<!--` to `&lt;!--`) are kept verbatim here. Both were already the behavior before this change.

## Tests

- `internal/render`: unit table for `StripHTMLComments` covering attribute values, RAWTEXT/RCDATA, IE conditionals, action-wrapping comments.
- Root package: the issue's own repro promoted to an assertion (certified **red** on current behavior, **green** with the fix), plus a comment-like-attribute-value guard.
- Full suite green (golden-template test and build-layer comment-decoy tests unaffected).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- HTML comments (`<!-- ... -->`) are now stripped from template output, matching `html/template` (which removes them during its escape pass). LiveTemplate builds statics by walking the raw parse tree, which never triggers that pass, so developer/internal comments previously shipped verbatim to the client (visible in view-source) and over WebSocket tree updates. Stripping uses the HTML tokenizer, so comment-like text in attribute values is preserved and comments inside `<script>`/`<style>`/`<textarea>` are left verbatim. See `docs/references/template-support-matrix.md` for residual divergences. (#468)
 - `NewGorillaUpgrader` now gives each upgrader its own write-buffer `sync.Pool` instead of sharing a package-level global, so upgraders configured with different `WriteBufferSize` values can no longer draw mismatched buffers from a shared pool.
 
 ## [v0.14.0] - 2026-06-13
--- a/comment_strip_test.go
+++ b/comment_strip_test.go
@@ -1,0 +1,118 @@
+package livetemplate
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// TestExecuteUpdates_StripsHTMLComments is the regression test for #468: HTML
+// comments must not survive into the static segments shipped to the client,
+// matching html/template (which strips them during its escape pass). This is
+// the issue's own repro promoted to an assertion.
+func TestExecuteUpdates_StripsHTMLComments(t *testing.T) {
+	const src = `<div><!-- SECRET-HTML-COMMENT -->{{.X}}{{/* SECRET-TMPL-COMMENT */}}</div>`
+
+	tmpl := Must(New("cmt"))
+	if _, err := tmpl.Parse(src); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	var buf bytes.Buffer
+	if err := tmpl.ExecuteUpdates(&buf, map[string]any{"X": "hi"}); err != nil {
+		t.Fatalf("ExecuteUpdates() failed: %v", err)
+	}
+
+	out := buf.String()
+	if strings.Contains(out, "SECRET-HTML-COMMENT") {
+		t.Errorf("HTML comment leaked into wire output (should be stripped like html/template):\n%s", out)
+	}
+	if !strings.Contains(out, "hi") {
+		t.Errorf("dynamic value missing from output:\n%s", out)
+	}
+}
+
+// TestExecuteUpdates_PreservesCommentLikeAttributeValue guards against the
+// over-eager-regex failure mode: comment-like text inside an attribute value is
+// real content and must be preserved.
+func TestExecuteUpdates_PreservesCommentLikeAttributeValue(t *testing.T) {
+	const src = `<div title="<!-- not a comment -->">{{.X}}</div>`
+
+	tmpl := Must(New("attr"))
+	if _, err := tmpl.Parse(src); err != nil {
+		t.Fatalf("Parse() failed: %v", err)
+	}
+
+	var buf bytes.Buffer
+	if err := tmpl.ExecuteUpdates(&buf, map[string]any{"X": "hi"}); err != nil {
+		t.Fatalf("ExecuteUpdates() failed: %v", err)
+	}
+
+	if !strings.Contains(buf.String(), "not a comment") {
+		t.Errorf("comment-like attribute value was wrongly stripped:\n%s", buf.String())
+	}
+}
+
+// TestParse_FullHTMLMarkerOnlyInComment_StillWraps guards the regression where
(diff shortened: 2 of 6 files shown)


adnaan:
Addressed the review in 46c33a5:

1. **🔴 `isFullHTML` computed before stripping (CONFIRMED)** — fixed. `isFullHTML` is now computed inside `parseInternal` *after* `StripHTMLComments` (and flattening), so a `<!DOCTYPE`/`<html` marker that exists only inside a now-stripped comment no longer routes a bare fragment through the full-document wrapper path. Removed the now-redundant pre-strip computation + `isFullHTML` parameter from both `Parse` and `ParseFiles` call sites. Added `TestParse_FullHTMLMarkerOnlyInComment_StillWraps` (asserts the `data-lvt-id` wrapper is present).
2. **🟡 `actionMask` U+E000 collision** — eliminated, not just mitigated. The placeholder delimiter now grows (`maskUnit` repeated) until it is provably absent from the source before masking, so a placeholder can never collide with literal template text. No silent-corruption path remains.
3. **🔵 `// Drop the comment.` nit** — removed; the empty `case html.CommentToken:` plus the function docstring already convey it.
4. **🔵 `restoreActions` O(N×len)** — kept as sequential `strings.Replace`. This runs once at parse time (not per render) and is bounded by template size; I'm favoring the clearer two-line form over a single-pass scan here. Happy to switch if you feel strongly.

Added action-boundary regression tests (`{{"<!-- x -->"}}`, `{{/* <!-- */}}`, `{{if eq .Sep "<!--"}}`, and `{{"a}}b"}}` with a comment) to lock in the masking behavior.

adnaan:
Addressed all three round-2 findings:

1. **🔴 `{{define}}` inside an HTML comment resolved before stripping (CONFIRMED)** — fixed. Stripping now happens *before* every parse of user template text — `Parse`, `ParseFiles` (both the entry file and additional files), and the component-template loader — so the parse tree is comment-free before composition detection/flattening. A `{{define}}` inside a comment is removed rather than resolved; the referencing `{{template}}` is left undefined and the template fails (matching html/template). Removed the now-redundant post-flatten strip from `parseInternal`. Added `TestParse_DefineInsideComment_NotResolved`.
2. **🟡 `restoreActions` O(k·n)** — switched to a single `strings.NewReplacer(...).Replace(s)` pass (with an empty-actions fast return).
3. **🔵 `skipQuoted` trailing-backslash bounds check** — added an explicit `if i >= len(s) { return -1 }` after the escape skip.

Thanks for confirming the mask-collision and `Raw()` concerns are non-issues — that matches my analysis (the delimiter is grown until absent from the source, so `mask+N+mask` can't collide).

adnaan:
Round 3 — thanks, no correctness issues. On the three observations:

2. **Package placement** — agreed and done (827c0e7): moved `StripHTMLComments` to `internal/compat` alongside `NormalizeTemplateSpacing`, since both are pre-parse source transforms invoked at the same call sites. Tests moved with it.
3. **`maskActions` godoc** — trimmed to the additive sentence only.
1. **Four strip call sites / `ParseWithStrip` helper** — declining for now. There's no `ParseFS` today, so this is YAGNI, and I find the explicit `compat.StripHTMLComments(...)` visible at each parse site clearer than hiding the transform behind a wrapper (the two primary sites also need the stripped `text` afterward for `parseInternal`, so a `Parse`-only wrapper wouldn't cover them uniformly). If/when `ParseFS` lands, consolidating then — with all real call sites known — is the cleaner refactor. Happy to revisit if you'd prefer the helper now.

PR #471, opened 2026-06-27
fix(mount): PRG fallback emits relative self-reference under StripPrefix (#444)

### LINKED ISSUE #444: fix(mount): PRG fallback uses stripped r.URL.Path — same StripPrefix bug as #434 for non-JS clients
## Problem

The progressive-enhancement POST-Redirect-GET fallback in `mount.go` (~L1472) redirects non-JS clients using the **stripped** request path:

```go
redirectURL := r.URL.Path
if encoded := r.URL.Query().Encode(); encoded != "" {
    redirectURL += "?" + encoded
}
http.Redirect(w, r, redirectURL, http.StatusSeeOther) // 303
```

When the handler is mounted behind `http.StripPrefix("/apps/login/", …)`, `r.URL.Path` is the stripped value (`/` or `""`), so a non-JS client submitting a form is redirected to the wrong location — the same class of bug that #434 fixed for `ctx.Redirect`.

This was flagged out-of-scope during the #434 review (PR #443) and deferred here.

## Scope

- Only affects the **progressive-enhancement (non-JS) PRG fallback** — JS clients get JSON tree updates and are unaffected.
- Only manifests under a non-root `http.StripPrefix` mount.

## Proposed fix

Apply the same approach #443 used for `ctx.Redirect`: emit a **relative** `Location` (e.g. the empty/`./` self-reference, percent-encoded) so the browser resolves it against its own un-stripped URL, instead of building an absolute redirect from the stripped `r.URL.Path`. The relative-emit + `hexEscapeNonASCII` machinery from #443 (`context.go`) can be reused.

## Acceptance

- A non-JS POST to a recipe mounted under `http.StripPrefix` at a non-root prefix redirects back to the original mount, not the stripped path.
- Regression test mirroring `TestRedirect_RelativeSelf_UnderStripPrefix` but exercising the mount.go PRG fallback (no `ctx.Redirect` call in the action).

Ref: #434, PR #443.

adnaan:
## Problem

The progressive-enhancement **POST-Redirect-GET (PRG)** fallback for non-JS clients (`mount.go`) built its 303 redirect from `r.URL.Path` — which `http.StripPrefix` has already stripped:

```go
redirectURL := r.URL.Path // stripped to "/" or "" under http.StripPrefix
http.Redirect(w, r, redirectURL, http.StatusSeeOther)
```

Mounted behind `http.StripPrefix("/apps/login/", …)`, a non-JS form POST was redirected to the wrong location. In the canonical trailing-slash case the path strips to `""`, producing an **empty `Location` header** that most clients ignore entirely — so the user's submission silently failed to navigate.

This is the same bug class #434/#443 fixed for `ctx.Redirect`. It was flagged out-of-scope during the #443 review and deferred here.

## Fix

Mirrors #443: emit a **relative self-reference** (`"./<last-segment>"`) raw in the `Location` header so the **browser** resolves it against its own un-stripped URL, instead of the server building an absolute URL from the stripped path.

The relative-emit machinery is extracted from `ctx.Redirect` into two shared, unexported helpers in `context.go` — `relativeSelfReference(r)` and `writeRelativeRedirect(w, r, target, code)` — so both call sites use one source of truth. `ctx.Redirect`'s behaviour is a pure, behaviour-preserving extraction (guarded by the existing `TestRedirect_RelativeSelf_UnderStripPrefix` #434 test).

### Behaviour notes for reviewers

- **PRG `Location` is now relative for *all* mounts, not just stripped ones.** There's no clean way to detect StripPrefix at the handler, and for a root mount `./` resolves identically to `/` for a real client — keeping it universal matches what `ctx.Redirect` already ships and avoids a fragile conditional.
- Because of that, `TestProgressiveEnhancement_FlashCookieConsumed` (mounted at root) needed an update: it reused the redirect `Location` directly in `httptest.NewRequest`, which rejects relative URIs. It now resolves the `Location` browser-style via `req.URL.Parse` before the follow-up GET. This test edit is a consequence of the universal relative `Location`, not unrelated churn.
- Query handling is byte-identical (`r.URL.Query().Encode()`); the flash-cookie block is unchanged.

## Scope

Swept all `http.Redirect` sites in the library: `context.go` (the #443 fix) and `mount.go` (this bug) are the only two. `mount.go`'s `path.Clean(r.URL.Path)` cache-invalidation compares the stripped path against itself (self-consistent, not a redirect). **#444 is the last instance of this bug class.**

## Tests

- Adds `TestPRGFallback_RelativeSelf_UnderStripPrefix` — mirrors the #434 test but exercises the PRG fallback (action returns `s, nil`, no `ctx.Redirect`). **Red-before-green verified**: reverting the `mount.go` change makes it fail (empty `Location` under the trailing-slash mount).
- Existing #434 test still passes → the `ctx.Redirect` extraction is behaviour-preserving.
- Full suite green; pre-commit hook (fmt + golangci-lint + tests) passed.

Ref: #434, PR #443. Closes #444.

### DIFF
--- a/context.go
+++ b/context.go
@@ -369,16 +369,10 @@ func relativeSelfReference(r *http.Request) string {
 // writeRelativeRedirect emits target RAW in the Location header so the client
 // resolves it against its effective (un-stripped) request URI — http.Redirect
 // would instead resolve it server-side against the http.StripPrefix-stripped
-// path, the bug this avoids. It mirrors http.Redirect's method-aware tail so
-// behaviour matches the absolute-path branch apart from URL resolution: a short
-// HTML body plus Content-Type for GET, Content-Type only for HEAD, neither for
-// POST (the action / PRG path), unless the caller already set a Content-Type.
+// path, the bug this avoids. It mirrors http.Redirect's method-aware tail.
 func writeRelativeRedirect(w http.ResponseWriter, r *http.Request, target string, code int) {
 	h := w.Header()
 	_, hadCT := h["Content-Type"]
-	// Percent-encode non-ASCII bytes so the Location stays within HTTP's ASCII
-	// header constraint — http.Redirect does this for the absolute branch via
-	// its own (unexported) helper.
 	h.Set("Location", hexEscapeNonASCII(target))
 	if !hadCT && (r.Method == http.MethodGet || r.Method == http.MethodHead) {
 		h.Set("Content-Type", "text/html; charset=utf-8")
--- a/context.go
+++ b/context.go
@@ -357,10 +357,12 @@ func (c *Context) Redirect(url string, code int) error {
 }
 
 // relativeSelfReference returns "./<last-segment>", a relative reference a
-// browser resolves back to the request's own (un-stripped) URL. It is the
-// StripPrefix-safe "reload self" target shared by ctx.Redirect("") and the
-// progressive-enhancement POST-Redirect-GET fallback (mount.go), neither of
-// which can see the prefix http.StripPrefix removed from r.URL.Path.
+// browser resolves back to the request's own URL under a trailing-slash mount.
+// It is the StripPrefix-safe "reload self" target shared by ctx.Redirect("") and
+// the progressive-enhancement POST-Redirect-GET fallback (mount.go), neither of
+// which can see the prefix http.StripPrefix removed from r.URL.Path. Under an
+// exact-match mount without a trailing slash a browser resolves "./" to the
+// parent path; see Context.Redirect's doc for that caveat.
 func relativeSelfReference(r *http.Request) string {
 	_, last := path.Split(r.URL.EscapedPath())
 	return "./" + last
--- a/context.go
+++ b/context.go
@@ -345,36 +345,52 @@ func (c *Context) Redirect(url string, code int) error {
 		// breaks clients) and resolves back to the current URL.
 		target := url
 		if target == "" {
-			_, last := path.Split(c.r.URL.EscapedPath())
-			target = "./" + last
+			target = relativeSelfReference(c.r)
 		}
-		// Mirror http.Redirect's method-aware tail so the two branches behave
-		// identically apart from URL resolution: a short HTML body plus
-		// Content-Type for GET, Content-Type only for HEAD, neither for POST
-		// (the action / PRG path), unless the caller already set a Content-Type.
-		h := c.w.Header()
-		_, hadCT := h["Content-Type"]
-		// Percent-encode non-ASCII bytes so the Location stays within HTTP's
-		// ASCII header constraint — http.Redirect does this for the absolute
-		// branch via its own (unexported) helper.
-		h.Set("Location", hexEscapeNonASCII(target))
-		if !hadCT && (c.r.Method == http.MethodGet || c.r.Method == http.MethodHead) {
-			h.Set("Content-Type", "text/html; charset=utf-8")
-		}
-		c.w.WriteHeader(code)
-		if !hadCT && c.r.Method == http.MethodGet {
-			if _, err := fmt.Fprintf(c.w, "<a href=\"%s\">%s</a>.\n", html.EscapeString(target), http.StatusText(code)); err != nil {
-				slog.Warn("Failed to write redirect body",
-					slog.String("component", "context"),
-					slog.Any("error", err))
-			}
-		}
+		writeRelativeRedirect(c.w, c.r, target, code)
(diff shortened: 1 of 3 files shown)


adnaan:
Thanks — addressed the two doc/comment nits in the next push:

- **`relativeSelfReference` doc (exact-match mounts):** softened to scope the guarantee to trailing-slash mounts and cross-reference `Context.Redirect`'s existing caveat for the non-trailing-slash case. Good catch — the shared doc shouldn't over-promise.
- **Test comment first sentence:** trimmed the function-name restatement; kept the WHY context.

Declining the **"undocumented behavioral change vs `http.Redirect`"** finding — it rests on an incorrect premise. `http.Redirect` does **not** emit a `Content-Type` or body for POST; per the stdlib (`net/http/server.go`, Go 1.26):

```go
if !hadCT && (r.Method == "GET" || r.Method == "HEAD") {
    h.Set("Content-Type", "text/html; charset=utf-8")
}
w.WriteHeader(code)
// Shouldn't send the body for POST or HEAD; that leaves GET.
if !hadCT && r.Method == "GET" {
    body := ...
}
```

`writeRelativeRedirect` mirrors this exactly (Content-Type only for GET/HEAD, body only for GET). The PRG fallback path is always POST, so the 303 response is byte-identical to the old `http.Redirect` call — no behavioral change for any middleware to observe.

adnaan:
Round-4 follow-up:

**#2 (exact-match caveat in `mount.go`)** — applied. Added a one-line note pointing readers of `mount.go` to `relativeSelfReference`'s exact-match caveat so the trailing-slash assumption is visible locally.

**#1 (`Content-Type` "behavior change")** — declining; the premise is incorrect, same as the round-2 finding I addressed in [this comment](https://github.com/livetemplate/livetemplate/pull/471#issuecomment-4815426161). `http.Redirect` does **not** set `Content-Type` "regardless of method" — it gates on `GET || HEAD` (`net/http/server.go`, Go 1.26):

```go
if !hadCT && (r.Method == "GET" || r.Method == "HEAD") {
    h.Set("Content-Type", "text/html; charset=utf-8")
}
```

The PRG fallback path is always POST, so the old `http.Redirect` call emitted **no** `Content-Type` for it — exactly as `writeRelativeRedirect` does now. There is no `Content-Type` delta on this path, so nothing to disclose in the PR description.

**#3 (drop "It mirrors http.Redirect's method-aware tail")** — declining. This same `writeRelativeRedirect` comment was reviewed as "fully justified" in round 2 and "checked out clean" in round 3; the sentence is a WHY pointer (the GET/HEAD/POST branching deliberately matches stdlib so the two redirect branches differ only in URL resolution), not WHAT narration. Per CLAUDE.md's bot-convergence guidance, I'm treating re-litigation of already-accepted content as cosmetic and addressing it by reply rather than another push.

CI is green (Unit/Fuzz/LVT/Docs all passing). Believe this is converged — over to the maintainer for signoff.

adnaan:
Round-5 follow-up — addressing by reply; no push, as the PR's `#444` delta has no new functional issue this round.

**#1 (`Query().Encode()` sorts keys)** — declining as out of scope. As the review notes, this is **pre-existing** behavior (the original PRG block already used `r.URL.Query().Encode()`); #444 is a stripped-path fix and deliberately keeps query handling byte-identical. `Encode()`'s re-encode is also the safer incumbent for reflecting a client query into a `Location` header. Switching to verbatim `r.URL.RawQuery` is a reasonable separate improvement — happy to file a follow-up issue if you'd like verbatim order preservation, but I'd rather not fold a query-semantics change into this redirect fix.

**#2 (test doesn't prove order fidelity)** — fair: `preserves_query` guards the **branch** (verified — deleting the query-append block fails it) but the alphabetical input can't distinguish preserve from sort, since the code normalizes. That's the same pre-existing `Encode()` semantics as #1, not a #444 regression. Noting it for the follow-up rather than asserting normalized output here.

**#3 (drop the trailing `mount.go` comment sentence)** — declining. That sentence ("This assumes a trailing-slash mount; see relativeSelfReference…") was added in [45e9c51f](https://github.com/livetemplate/livetemplate/commit/45e9c51f) **at this bot's round-4 request** (finding #2: "add one sentence to the mount.go comment that mirrors relativeSelfReference's note"). Round 5 now asks to remove it — I'll leave it, since a `mount.go` reader benefits from the local pointer.

Two consecutive rounds with no new functional issue on the actual change — treating this as converged per CLAUDE.md's bot-convergence guidance. Over to the maintainer for signoff.

Issue #385, opened 2026-05-08
Tracking: PR #233 form-handling progressive enhancement follow-ups


adnaan:
Tracking issue to bundle the unaddressed follow-ups from [PR #233](https://github.com/livetemplate/livetemplate/pull/233) (progressive complexity model for form handling) and [PR #250](https://github.com/livetemplate/livetemplate/pull/250) (related docs work). These were filed five separate review comments but represent one coherent feature gap — without all of them landing, server-side form validation in progressive enhancement is silently incomplete.

## Sub-issues

| # | Priority | Area | Status |
| --- | --- | --- | --- |
| #236 | P2 | feature | Auto-wire `FormSchema` into `Context` from template statics — `WithFormSchema` exists but `mount.go` never calls it, making `ValidateForm()` a silent no-op for users |
| #237 | P1 | correctness | Replace empty-value button heuristic with explicit client-side submitter — see proposal in [PR #382](https://github.com/livetemplate/livetemplate/pull/382), implementation pending |
| #238 | P2 | security | Add CSRF protection note to progressive enhancement docs |
| #239 | P2 | feature | Support `formnovalidate` server-side in `ValidateForm` |
| #241 | P3 | code-quality | Consistent sentinel design in `FormRule` struct (e.g., `MinLength: -1` vs `HasMin bool`) |

## Why bundle these

- Five orphaned tickets from one review create the false impression that each is independently scoped — they're not. #236 is load-bearing for #239 (no schema = no formnovalidate handling).
- #237's design proposal landed in PR #382 but the implementation is still queued; coordinating the explicit-submitter wire change with the rest of the form cluster avoids a second migration.
- #238 (CSRF) is the most-deferrable but should ship with the rest so the docs aren't half-true.

## Recommended sequencing

1. **#236** first — auto-wire the schema so `ValidateForm` actually validates.
2. **#237** next — implement the explicit-submitter wire change per [PR #382](https://github.com/livetemplate/livetemplate/pull/382).
3. **#239** — `formnovalidate` honoring depends on #236.
4. **#241** + **#238** — cleanup pass.

Close this tracker when all five are closed.

adnaan:
All five sub-issues are now closed:

- **#236** — auto-wire `FormSchema` into `Context` from template statics ✅
- **#237** — explicit client-side submitter (replacing the empty-value heuristic) ✅
- **#238** — CSRF protection note in progressive-enhancement docs ✅
[...]

PR #472, opened 2026-06-27
feat(formvalidation): honor formnovalidate server-side in ValidateForm (#239)

### LINKED ISSUE #239: FOLLOW-UP: Support formnovalidate server-side in ValidateForm

adnaan:
Closes #239 — the **last open item** of the #385 form-handling progressive-enhancement tracker (#236/#237/#238/#241 already shipped).

## Problem

The docs advertise `<button name="save-draft" formnovalidate>` to skip validation, but `ctx.ValidateForm()` validated every field regardless of which button submitted — the `formnovalidate` intent was silently dropped.

## Approach — server-side schema inference (no client release)

Chosen over the issue's literal "client sends a flag" proposal because the server already knows the submitter on **every** tier, so this covers pure no-JS too (where server-side validation matters most) with zero client/wire changes:

- `ExtractFormSchema` scans `<button>`/`<input>` for `formnovalidate` and records each control's `name` into `FormSchema.NoValidateSubmitters`.
- The clicked button's name (`msg.Submitter`) is threaded into `Context` at the two form-action sites (WS `mount.go:889`, HTTP `:1531`). `ValidateForm` skips when the submitter — or, under button-name routing, the action it collapses to — is in the set.
- Works on WebSocket, HTTP-fetch, and no-JS native POST.

### Why a `Context.submitter` field (distinct from `action`)
Under `lvt-on:submit` routing the action is the *handler* and the submitter is the *clicked button* — they differ on the wire. Checking only `action` would miss `formnovalidate` there. Covered by dedicated HTTP and WebSocket tests where action ≠ submitter.

### Bundled fix: kebab-case action routing (maintainer-approved scope add)
While verifying the no-JS tier I found the documented pattern `<button name="save-draft">` → `SaveDraft()` **did not route**: `methodNameToActions` emitted only camelCase/snake_case/PascalCase, so kebab `save-draft` → `ErrMethodNotFound` and the action never ran. That would have left #239's no-JS coverage broken for the exact examples the docs show. `methodNameToActions` now also emits a kebab variant, so the canonical button-name pattern routes on no-JS. (Additive — kebab forms are a new namespace, no collision with existing forms.)

## Caveats (documented + tested)
- **Named submitters only** — a dynamic (`{{...}}`) name is blanked before extraction.
- **No-JS needs an empty-value button** — the no-JS submitter is identified by its empty-value form field (`detectSubmitButtonName`), so a `formnovalidate` button carrying a `value` isn't recognized there. Locked by `TestFormNoValidate_NoJS_ValuedButtonBoundary`.
- **Not a security boundary** — client-controlled convenience for draft flows; enforce server-authoritative rules unconditionally where it matters.

## Tests
- Unit: schema extraction, `ValidateForm` skip with submitter≠action, kebab `methodNameToActions`.
- Integration: no-JS button-name (asserts `meta.success` so a routing miss can't pass silently), explicit-submitter **HTTP** and **WebSocket** (submitter≠action), no-JS value-button boundary.
- Red-before-green verified (removing the skip block fails the skip tests; removing the kebab variant fails no-JS routing).

## Docs
In-repo docs updated: flipped the stale "not respected server-side" limitation **and** the adjacent stale #236 auto-wire row to the real residual limitations; documented behavior + caveats in the progressive-complexity guide and reference; CHANGELOG entries for both formnovalidate and kebab routing.

**Follow-up (separate, release-pinned):** the `../docs` content site + a runnable example pin the *published* livetemplate version, so those land after this releases — standard ecosystem lockstep.


adnaan:
Round-2 follow-up:

**#1 (inline casing comments)** — applied; dropped all four. The kebab WHY already lives in the function doc-comment.

**#2 (`submitter` bypasses the builder)** — applied as an **unexported** `withSubmitter` builder, used at both dispatch sites. This removes the lone direct-mutation and the "survives copies" comment, matching the builder pattern. I kept it unexported rather than a public `WithSubmitter`: the submitter is framework-populated from the client's `SubmitEvent` at dispatch time, not a value applications set when constructing a Context — the read side (`ctx.Submitter()`, added in round 1) is the user-facing surface. Happy to export it if you'd prefer symmetry with `WithUserID` et al.

**#3 (normalize the submitter in `resolveSubmitterFallback`)** — declining; it would break the `ctx.Submitter()` semantics added in round 1. `resolveSubmitterFallback` runs for **every** action, not just form submits. Setting `msg.Submitter = msg.Action` whenever the submitter is empty would make a non-form action report a submitter — e.g. `lvt-on:click="increment"` would yield `ctx.Submitter() == "increment"`, contradicting the getter's contract ("the control that submitted the form, or empty if none"). The fallback is deliberately scoped to `ValidateForm`, where "action is the button" only holds under button-name routing; `resolveSubmitterFallback` can't distinguish a no-JS form-button submit from an ordinary action, so it can't safely populate the submitter globally.

adnaan:
**Round 3 fixes — pushed in `460ec4a1`.** All three findings + the minor applied.

**Finding 1 — submit-control type guard.** `ExtractFormSchema` now only registers a `formnovalidate` name as a no-validate submitter when the control is a genuine submit control: a `<button>` whose `type` is not `button`/`reset` (default is submit), or an `<input type=submit|image>`. A `type=button`/`type=reset` carrying `formnovalidate` no longer registers a spurious no-validate name.

**Finding 2 — submitter canonicalization for no-JS.** The `detectSubmitButtonName` path in `parseURLEncodedForm` now sets `msg.Submitter` (not just `msg.Action`), so an empty-value submit button (`<button name="save-draft" formnovalidate>` with no `value`) is recognized as the form's submitter on the native-POST tier. With the submitter now populated on every form-submit path, `ValidateForm` drops its action-name fallback and does a single `NoValidateSubmitters[c.submitter]` lookup. This achieves round-2's goal (no routing knowledge in `ValidateForm`) in the correct place — `detectSubmitButtonName`, not by normalizing in `resolveSubmitterFallback` (which would have broken `ctx.Submitter()` for non-form actions like `lvt-on:click="increment"`).

**Finding 3 — skip non-data input types in rule inference.** The `inputAttrRegex` loop now `continue`s on `type` in `{submit, image, button, reset}`, so a `required` on a non-data control can't raise a spurious field error for a value never present in the POST body.

**Minor.** Tests use the unexported `withSubmitter()` builder rather than poking the field directly; added `TestExtractFormSchema_SubmitInputNoSpuriousRule` and `TestParseActionFromHTTP_ButtonNameSetsSubmitter`; extended `TestExtractFormSchema_FormNoValidate` (type=button/reset/non-submit-input exclusion) and `TestValidateForm_FormNoValidate` ("no submitter does not skip even if action matches").

The one deliberate behavior change is the fallback removal: `lvt-action="save-draft"` (hidden field) + no submitter + no-JS now *validates* where it previously *skipped*. That corner is not the documented pattern, the submitter-based path covers the real no-JS case, and re-adding the fallback would reintroduce the routing-knowledge-in-`ValidateForm` smell flagged in round 2 — so it stays removed. Full suite + `internal/send` green; pre-commit clean.

adnaan:
**Round 4 fixes — pushed in `daa8b2bf`.** Both confirmed bugs + the plausible flag + the nit addressed.

**🔴 #1 — `ValidateForm` nil-deref.** Added `if c.data == nil { return nil }`, matching every other `c.data` accessor. As you noted, the formnovalidate-only nil-guard change made this reachable; the guard also closes the pre-existing rules-bearing variant (nil-data dispatched context with a normal schema would have panicked before this PR too). `return nil` (not `Validate(nil)`) is deliberate — nil data is a non-form context, so raising required-field errors there would be spurious. Regression test: `TestValidateForm_NilData_NoPanic` constructs a `Context` with `c.data == nil` directly (note `NewContext` wraps `nil` in a *non-nil* `*ActionData`, so the panic only reproduces via direct construction, as on the Mount/dispatch path).

**🔴 #2 — `parseMultipartForm` missing `msg.Submitter`.** Fixed. I also fixed the **same bug class in `BuildActionFromValues`** (the streaming-upload path that reads via `MultipartReader`, so `r.MultipartForm` is never populated) — its `detectSubmitButtonName` branch had the identical omission, which a file-grep for `parseMultipartForm` wouldn't surface. Both now set `msg.Submitter = name`, matching `parseURLEncodedForm`. Tests: extended `TestParseActionFromHTTP_Multipart_ButtonNameAction` with a submitter assertion; added `TestBuildActionFromValues_ButtonNameSetsSubmitter`.

**🟡 #3 — `submitControlRegex` case-insensitivity.** Added `(?is)` to both `submitControlRegex` **and** `inputAttrRegex` so neither HTML-parsing regex in `formvalidation.go` diverges from the `buttonElemRegex`/`inputElemRegex` peers in `topic_wired_actions.go`. Locked by an uppercase-tag case (`<BUTTON … FORMNOVALIDATE>`) in `TestExtractFormSchema_FormNoValidate`.

**Nit — `ValidateForm` comment.** Trimmed the four-line comment to a single line stating the why ("one lookup covers all tiers — the submitter is unified upstream").

**One test assertion intentionally flipped — flagging explicitly.** In `TestParseActionFromHTTP_Multipart_ExplicitSubmitter`'s "empty lvt-submitter is no-op" subcase, the expectation changes from `Submitter == ""` to `Submitter == "save"`. This is a **consistency correction, not a regression**: with `lvt-submitter` empty, `resolveSubmitterFallback` no-ops and the heuristic detects the empty-value `save` button — which *is* the clicked submitter. The url-encoded path already records it (see `TestParseActionFromHTTP_ButtonNameSetsSubmitter`); the old multipart-only `""` assertion encoded exactly the divergence Finding #2 named. The subcase's real contract ("empty `lvt-submitter` doesn't short-circuit the heuristic") is still proven by `Action == "save"`.

Full root + `internal/send` suite green; pre-commit clean.

adnaan:
**Round 5 nits — pushed in `dc4f5622`.** All three comment-style observations applied (comment-only, no behavior change):

- **`internal/send/message.go` triplication** — kept the full "detected button is also the submitter" explanation once at the first occurrence (`BuildActionFromValues`); the two parallel copies are now a one-line back-reference.
- **`submitControlRegex` comment** — dropped the trailing WHAT sentence about the capture group.
- **`context.go` `ValidateForm` comment** — trimmed to the non-obvious why (submitter unified upstream), dropping the self-evident "one lookup".

Thanks for the "Confirmed correct" pass — agreed on all five (empty-string bypass guard, zero-rules redundant-but-defensive bypass, server-initiated dispatch paths having `formSchema == nil`, no real kebab collision, and the inert-`s`/active-`i` reading of `(?is)`).

Round 4 was the last round with a functional finding (both confirmed bugs fixed in `daa8b2bf`); round 5 surfaced no functional issues. Per CLAUDE.md's review-loop convergence guidance, I'll address any further **cosmetic-only** feedback by reply rather than by push to avoid churn. The branch is otherwise ready for maintainer review.

adnaan:
**Round 6 fix — pushed in `f332bece`.** Good catch; this was a real consumer-side gap.

**🔴 Bug fixed.** The proxied streaming-upload per-part context built `streamCtx.WithData(send.BuildActionFromValues(values).Data)` — passing only `.Data` and discarding `.Submitter`, so `ctx.Submitter()` returned `""` inside `OnUpload` even for a named-button submit. Now builds the message once and threads both:

```go
partMsg := send.BuildActionFromValues(values)
partCtx := streamCtx.WithData(partMsg.Data).withSubmitter(partMsg.Submitter)
```

Regression test `TestProxiedUpload_OnUploadReadsSubmitter` is red-before-green verified (fails without the fix, passes with it).

**Consumer-class sweep — so this is the last instance, not round 7's finding.** I enumerated every handler-facing context that carries form data and confirmed the submitter is threaded at each:

| Site | Threaded |
|---|---|
| WS dispatch (`mount.go:890`) | ✅ |
| HTTP final-action dispatch (`:1534`) — also covers the streaming final action (`msg` from `BuildActionFromValues` at `:1426` flows here; the only `return` between them is the `__navigate__` WS-only guard) | ✅ |
| `OnUpload` proxied per-part (`:1415`) | ✅ (this fix) |

`OnUpload` is invoked from exactly one site (`:2558`, with `partCtx`); `.WithData(` has exactly one site (`:1415`). `TriggerAction` / upload-complete / `stageVolumePart` are submitter-less by design (server-initiated or no context — cleared in round 5).

**Also addressed (comment-only):**
- Corrected the `TestValidateForm_NilData_NoPanic` comment — the nil `c.data` path is reachable only from direct struct construction, not from dispatch (which always goes through `NewContext`, wrapping `nil` in a non-nil `*ActionData`). The guard exists for exactly that direct-construction case.
- Trimmed the `withSubmitter` docstring to one line per CLAUDE.md.

Full root + `internal/send` suite green; pre-commit clean.

adnaan:
**Round 7 — pushed in `c3abb647`.** No functional findings this round; applied the one substantive gap, declining the two regex suggestions with rationale.

**Applied:**
- **#3 (CHANGELOG)** — added a **Fixed** note: `ExtractFormSchema` no longer emits a `FormRule` for submit controls (`<input type="submit|image|button|reset">`); a `required` on a named submit input previously raised a spurious "field is required" error when that control wasn't the submitter. Flagged as a public-API behavior change independent of the `formnovalidate` skip.
- **#2 (partial)** — documented on `inputAttrRegex` *why* it's case-insensitive: the public `ExtractFormSchema` also accepts hand-written, possibly mixed-case statics (template-rendered statics are lowercase).

**Declined (with rationale):**
- **#1 — collapse `submitControlRegex` into `buttonElemRegex`/`inputElemRegex`.** Reusing the two twins requires a closure (to avoid duplicating the submit-detection body) plus two `FindAllStringSubmatch` passes, versus one regex + one loop today — net not simpler, and it loses the single-pass tag capture I use to distinguish a `<button>`'s default-submit from an `<input>`'s explicit type. Round 4 already aligned this regex's flags (`(?is)`) with the twins, so the divergence risk you raise is mitigated without a structural merge. Happy to revisit if you'd prefer the two-pass form.
- **#2 — drop the inert `(?s)`.** You're correct that `(?s)` is a no-op against `[^>]*`. I'm keeping `(?is)` deliberately for consistency with the sibling `buttonElemRegex`/`inputElemRegex` in `topic_wired_actions.go`, which both carry `(?is)` — matching their exact form is what round 4 asked for, and divergence (even cosmetic) is the maintenance trap that finding named. Dropping `(?s)` from this one regex would re-introduce that asymmetry.

**Minor (nil-data comment)** — already addressed in `f332bece` (round 6). The current comment reads: *"A nil `c.data` is only reachable from direct struct construction, not from dispatch; the guard exists for exactly this case."* — which is the revision you're suggesting. The quoted "dispatched/Mount path" text is from the pre-round-6 version.

Per CLAUDE.md review-loop convergence: round 6 was the last round with a functional finding; round 7 surfaced none. I'll address any further cosmetic-only feedback by reply rather than push.

adnaan:
**Round 8 — pushed in `9f04fb9d`.** No functional findings; fixed the comment contradiction you flagged as the pre-merge blocker, declining the remaining comment-style nits.

**Applied:**
- **Confirmed finding (`context.go:590`)** — corrected the contradictory comment. It now reads: *"Guard direct Context construction; dispatch always produces non-nil data via NewContext → newActionData, so this only fires for zero-value Contexts."* Matches the test comment; the old "(Mount connect, dispatched actions)" was wrong.
- **`withSubmitter`** — trimmed to a single line.

**Declined (with rationale):**
- **`internal/send/message.go:124` canonical comment** — this is the *single* explanation deliberately kept when the triplicated copies were collapsed (the other two are now one-line back-references to it). It documents the non-obvious invariant — the same detected name must populate both `msg.Action` and `msg.Submitter` because they feed two distinct public methods (`Action()` / `Submitter()`). `msg.Submitter = name` shows the *what*; this is the *why*, and it's the only place it's stated.
- **Test-function comments** — each states the non-obvious *why* of its case (e.g. *why* nil `c.data` is only reachable via direct construction, *why* a submit input must not yield a rule). Removing them drops the rationale a future reader needs; the test name alone doesn't carry it.
- **Efficiency note (two `FindAllStringSubmatch` passes)** — as you note, it runs once per template parse over a few KB, so the cost is negligible; a single-pass categorizer is a reasonable future cleanup but out of scope for this PR.

**CI note:** the `Test LVT against Core Changes` failure on the prior commit was `TestProgressiveEnhancement_Disabled`, a pre-existing flake unrelated to #239 — a race between template auto-discovery and a *sibling* upload test's `.uploads/.tmp` teardown, exposed only under `-race`. Proof: the **same commit** `c3abb647` **passed** that test in the `Unit Tests` job and failed it only in the `-race` LVT job (and it passes locally). This push re-runs it.

Round 6 was the last round with a functional finding; rounds 7 and 8 surfaced none. Per CLAUDE.md convergence I'll handle any further cosmetic-only feedback by reply rather than push.

adnaan:
**Round 9 — converged; handling by reply, not push.** This round found no correctness or coverage issues (your words), and the three items are cosmetic comment-style preferences:

- **`message.go:124` multi-line comment** — this is the same canonical comment discussed in round 8. It's deliberately the *one* full statement of the non-obvious invariant (the detected name must populate both `msg.Action` and `msg.Submitter` because they feed two distinct public methods); the two sibling sites are the compact `// … (see BuildActionFromValues)` back-references *to it*. Keeping one full explanation + two pointers is the intended shape.
- **`context.go:128` `withSubmitter`** and **`context.go:592` bridging comment** — reasonable to drop; both are one short line and the `withSubmitter`/`Submitter()` cross-reference aids discoverability of the submitter-unification invariant, so I'm leaving them as-is rather than spend another review+CI round on a comment deletion.

Per CLAUDE.md's review-loop convergence guidance: rounds 7, 8, and 9 have each identified **no new functional issue** — only successive comment-phrasing alternatives. That's the stop signal. I'll address any further cosmetic-only feedback by reply rather than by push, to avoid an unbounded review/CI cycle.

All CI green on `9f04fb9d` (Unit, Fuzz, benchmark, Cross-Repo LVT/Tinkerdown/Docs, claude-review); PR is mergeable (CLEAN). Ready for maintainer review.

PR #478, opened 2026-07-05
feat(origin): support RFC 7239 Forwarded header for scheme detection (#198)

### LINKED ISSUE #198: FOLLOW-UP: Support RFC 7239 Forwarded header (from PR #190)
## Context

Follow-up task from [PR #190](https://github.com/livetemplate/livetemplate/pull/190).

**Priority:** Low
**Source reviewer:** claude-review

## Description

The standard `Forwarded: proto=https` header (RFC 7239) isn't currently handled. While `X-Forwarded-Proto` is what all major proxies use in practice, supporting the standard header would be more complete.

Add support for parsing the `Forwarded` header as a fallback or alongside `X-Forwarded-Proto`.

adnaan:
## Summary

Closes #198.

Same-origin WebSocket origin checks previously derived the request scheme **only** from `X-Forwarded-Proto`. This adds the standard **RFC 7239 `Forwarded`** header (`proto` parameter) as a fallback, so deployments behind proxies that emit the standard header rather than the de-facto `X-Forwarded-Proto` get correct scheme detection.

## Behavior

- **Precedence unchanged / zero regression:** `X-Forwarded-Proto` still wins (every major proxy sets it). `Forwarded` is consulted only when `X-Forwarded-Proto` is absent or invalid.
- **Same security boundary:** both headers sit behind the existing `WithTrustForwardedHeaders` gate — a direct (non-proxied) client cannot forge either to upgrade the detected scheme.
- **Parsing scope:** takes the first (client-most) comma-separated element's `proto` param, case-insensitive param name, accepts quoted values (`proto="https"`). Scoped to the `proto` token only — not a full RFC 7239 tokenizer — which is safe because `proto` is always a bare `http`/`https` token.

## Changes

- `template.go` — new `forwardedScheme` / `forwardedHeaderProto` / `normalizeScheme` helpers; `createSecureOriginChecker` now routes through `forwardedScheme`. The old inline `X-Forwarded-Proto` block is folded into the shared `normalizeScheme` primitive.
- `config.go` + `template.go` godoc — updated `TrustForwardedHeaders` / `WithTrustForwardedHeaders` / `WithAllowedOrigins` prose to mention the RFC 7239 fallback (no stale "X-Forwarded-Proto only" claims).
- `origin_checker_test.go` — table cases for: proto http/https, mismatch rejection, quoted value, case-insensitive param, multi-element first-hop, no-proto fallback to `r.TLS`, `X-Forwarded-Proto` wins over conflicting `Forwarded`, invalid `X-Forwarded-Proto` falling through to `Forwarded`, and the untrusted-mode security case (forged `Forwarded` ignored).

## Testing

- Full package suite + `go vet` green.
- `/simplify` pass: no changes (diff already minimal).


adnaan:
Thanks for the review. Addressed the actionable nit:

- **CHANGELOG** — added an entry under `[Unreleased] › Added` referencing #198 (commit f965c7a6), matching the repo convention of documenting public behavior changes with an issue reference.

The two non-blocking points are acknowledged and intentionally left as-is:

- **`forwardedHeaderProto` is scoped to `proto` only** — the naive semicolon split is safe precisely because `proto` is always a bare `http`/`https` token. The doc comment flags this so a future reuse for `for=` (IPv6 literals, quoted values) would know to reach for a real tokenizer. Not expanding the parser beyond what's needed.
- **`r.Header.Get` returns only the first field line** — pre-existing behavior carried over verbatim from the original `X-Forwarded-Proto` code, not introduced here; out of scope for this fix.

PR #479, opened 2026-07-05
fix(parse): error on bare arg-requiring method reference like text/template (#203)

### LINKED ISSUE #203: FOLLOW-UP: Improve callMethod behavior for non-zero-arg methods

### LINKED ISSUE #459: Bare reference to an argument-requiring method should error (eval-node level), not stringify as a func value

adnaan:
## Summary

Closes #203. Also closes #459 (the eval-node-altitude fix that #203 was split into).

A **bare** template reference to an argument-requiring method — e.g. `{{.Get}}` where `Get(key string) string` — previously resolved the *uncalled* bound method as a value, which then flowed into `valueToString` and rendered as a stringified Go func value (a confusing `0x…`-style output). `text/template` instead errors: `wrong number of args for Get: want 1 got 0`.

## Why the fix is at the eval-node altitude, not `callMethod`

`resolveFieldChain` resolves (and, for zero-arg methods, *calls*) each method **before** the caller knows whether trailing args follow. The uncalled-method return is **load-bearing** for the with-args path — `{{.ctx.Get "key"}}` works precisely because `resolveFieldChain` yields the bound `Get` method, which `callMethodOrFieldWithArgs` then invokes with `"key"`. So making `callMethod` error on `NumIn() != 0` would break `{{.ctx.Get "key"}}`, and eagerly calling variadic methods there would break `{{.Tags "a"}}`. The bare-vs-with-args distinction is known only at the eval nodes, which see `len(args)`.

#203 shipped its documentation option in PR #460 and split this behavioral fix into #459; this PR implements #459.

## Change

- `callMethod` now wraps an argument-requiring method in a small `uncalledMethod{fn, name}` rather than returning the raw reflect method value. This lets the eval nodes distinguish "method awaiting trailing args" from "genuinely bare reference," and keeps such a value from leaking to `valueToString`.
- The three eval-node bare branches (`evalFieldNode` / `evalChainNode` / `evalVariableNode`) and `evalArg` finalize the wrapper via `finalizeBareValue`:
  - a **variadic** method (`{{.Tags}}` for `Tags(...string)`) is **called with no args**, matching `text/template`;
  - a method that **genuinely requires args** returns a descriptive `ParseError` naming the method.
- The **with-args path** (`{{.ctx.Get "key"}}`, `{{.Tags "a" "b"}}`) unwraps the method and invokes it — unchanged.
- `resolveFieldChain` finalizes a pending `uncalledMethod` at the top of the loop, so a further field on an arg-requiring method (`{{.Get.Foo}}`) errors instead of silently returning nil.

## Testing

`TestStdlibParity_BareMethodArgs` locks behavior against `html/template` for both directions, derived from an empirical probe of stdlib:

| Template | Behavior (both engines) |
|---|---|
| `{{.Name}}` zero-arg method | called |
| `{{.Tags}}` bare variadic method | called with empty variadic |
| `{{.Tags "a" "b"}}` | called with args |
| `{{.Get "key"}}` / `{{.ctx.Get "key"}}` | called with args (unchanged) |
| `{{.Get}}` bare arg-requiring method | **errors**, naming `Get` |
| `{{printf "%s" .Get}}`, `{{if .Get}}…{{end}}`, `{{.ctx.Get}}` | **errors** |

Full repo suite + `go vet` green. `/simplify` pass applied (one redundant comment trimmed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `livetemplate_publishes_sent_total` now actually increments. From v0.11.0 through v0.15.0 the counter (and its predecessor `livetemplate_broadcasts_sent_total`) was defined but never wired to a production call site, so it always reported 0 — a footgun for operators with dashboards or alerts on peer-fan-out rate. It is now incremented once per receiving connection whenever a peer-fan-out dispatch is enqueued, covering `ctx.Publish`, `Session.TriggerAction`, and cross-instance group/topic re-fan-out. The count is per instance (each instance counts only its own local deliveries, so a clustered publish is not double-counted) and is recorded at enqueue time, so a downstream slow-client close can still drop the resulting WebSocket send. (#432)
+- A bare template reference to an argument-requiring method (e.g. `{{.Get}}` where `Get(key string) string`) now errors like `text/template` (`wrong number of args for Get: want 1 got 0`) instead of silently stringifying the uncalled method as a Go func value. A bare reference to a *variadic* method (`{{.Tags}}` for `Tags(...string)`) is now called with no arguments, also matching `text/template`. The with-args method path (`{{.ctx.Get "key"}}`) is unchanged. The fix lives at the eval-node altitude in `internal/parse/eval.go`, where the presence of trailing args is known — `callMethod`/`resolveFieldChain` run before that and so cannot distinguish "awaiting args" from "bare reference". (#203, #459)
 
 ### Added
 
--- a/internal/parse/eval.go
+++ b/internal/parse/eval.go
@@ -407,19 +407,30 @@ type uncalledMethod struct {
 }
 
 // callBare evaluates the method as a bare reference (no trailing args). A
-// variadic method is called with an empty variadic ({{.Tags}} calls Tags());
-// a method that genuinely requires arguments returns a ParseError matching
-// text/template's "wrong number of args for X: want N got 0".
+// variadic method with only its variadic parameter is called with an empty
+// variadic ({{.Tags}} calls Tags()); a method that genuinely requires arguments
+// returns a ParseError matching text/template's "wrong number of args" text —
+// "want N" for a fixed arity, "want at least N" for a variadic one.
 func (m uncalledMethod) callBare() (interface{}, error) {
 	mt := m.fn.Type()
-	if mt.IsVariadic() && mt.NumIn() == 1 {
-		return invokeNoArgs(m.fn)
+	if mt.IsVariadic() {
+		fixed := mt.NumIn() - 1 // drop the trailing variadic slot
+		if fixed == 0 {
+			return invokeNoArgs(m.fn)
+		}
+		return nil, m.argCountError(fmt.Sprintf("want at least %d", fixed))
 	}
-	return nil, &ParseError{
+	return nil, m.argCountError(fmt.Sprintf("want %d", mt.NumIn()))
+}
+
+// argCountError builds the bare-reference "wrong number of args" ParseError,
+// mirroring text/template's message ("...: <want> got 0") for method X.
+func (m uncalledMethod) argCountError(want string) error {
+	return &ParseError{
 		Phase:    "eval",
 		NodeType: "method",
 		Expr:     m.name,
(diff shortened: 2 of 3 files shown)


adnaan:
Thanks — good catch on the message-parity gap. Addressed in 313c07d2:

- `callBare` now handles the **mixed fixed+variadic** shape: for a variadic method it computes `numFixed = NumIn()-1` and emits `want at least N` (extracted into `argCountError`), so a bare `{{.Log}}` for `Log(level string, args ...interface{})` now reports `wrong number of args for Log: want at least 1 got 0`, matching `text/template` exactly (verified via a stdlib probe). Pure-fixed methods keep `want N`; pure-variadic (`{{.Tags}}`) still calls with no args.
- `TestStdlibParity_BareMethodArgs` gains a mixed fixed+variadic case and now asserts the **full `wrong number of args...` message core** against `text/template` (stripping stdlib's positional-prefix wrapping), rather than only checking the method name — so any future wording drift is caught.

The with-args "not enough args" message (`callFunc`) still uses LiveTemplate's own phrasing, but that path is outside the bare-reference scope of #203/#459 and pre-dates this change.

PR #480, opened 2026-07-05
refactor(json): migrate remaining encoding/json callsites to jsonutil.API (#231)

### LINKED ISSUE #231: FOLLOW-UP: Migrate remaining encoding/json callsites to json-iterator (from PR #229)
## Context

PR #229 replaced `encoding/json` with `json-iterator/go` in the hot paths (`internal/send/`, `internal/build/types.go`). Several non-hot-path files still use `encoding/json` directly.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/229

## Remaining callsites

These files still import `encoding/json` and could be migrated to `jsonutil.API` for consistency:

- `action.go` — `ActionData.Bind()` uses `json.Marshal`/`json.Unmarshal`
- `state.go` — `jsonState.MarshalBinary`/`UnmarshalBinary`, `SerializeState`, `DeserializeState`
- `mount.go` — HTTP response encoding (`json.NewEncoder`), WebSocket message parsing (`json.Unmarshal`)
- `health.go` — Health check response encoding
- `session_stores.go` — Session metadata serialization
- `pubsub/redis.go` — Redis message marshaling
- `internal/upload/protocol.go` — Upload protocol messages

## Priority

**Low** — These are not in the per-render hot path. The performance impact of migrating them would be marginal. The main benefit is consistency (single JSON library across the codebase).

## Notes

- `mount.go` has the most callsites (~8 locations) and would benefit most from migration
- `health.go` uses `json.NewEncoder` which maps to `jsonutil.API.NewEncoder`
- All migrations are mechanical — swap `json.` to `jsonutil.API.`

---

adnaan:
## Summary

Closes #231 (FOLLOW-UP from PR #229). PR #229 moved the hot paths (`internal/send`, `internal/build/types.go`) to json-iterator via `jsonutil`. This migrates the remaining **non-hot-path** callsites so the whole codebase goes through a single JSON library:

- `action.go` — `ActionData.Bind()`
- `state.go` — `MarshalBinary`/`UnmarshalBinary`, `ExtractPersistFields`/`InjectPersistFields`
- `mount.go` — HTTP response encoding + WS message parsing (~15 callsites)
- `health.go` — health-check response encoding
- `session_stores.go` — session metadata (de)serialization
- `pubsub/redis.go` — Redis message marshaling
- `internal/upload/protocol.go` — upload protocol messages

All swaps are mechanical: `json.Marshal/Unmarshal/NewEncoder` → `jsonutil.API.*`.

## Wire compatibility

`jsonutil.API` is `ConfigCompatibleWithStandardLibrary`, so **output is byte-for-byte unchanged** — same HTML-escaping (`EscapeHTML=true`; deliberately *not* the `NoEscape` variant) and the same `Encoder.Encode` trailing newline. The full suite (including response/health/tree tests that assert exact bytes) passes unchanged.

## `json.RawMessage`

`state.go` keeps its `encoding/json` import **solely for the `json.RawMessage` type** (a type reference in `InjectPersistFields`, not a call). json-iterator's compat config decodes the stdlib `RawMessage` type natively, so only the function calls were swapped.

## ⚠️ One behavioral nuance — cyclic user state

`state.go`'s `MarshalBinary`/`ExtractPersistFields` serialize **arbitrary user application state**. json-iterator does **not** detect reference cycles the way `encoding/json` does, so a cyclic user-state struct that previously returned a marshal *error* now overflows the stack.

- Such state is **already non-persistable** under either library — the only change is graceful-error → crash for an already-broken edge case.
- **Not** guarded with `recover()`: a Go stack overflow is a *fatal runtime error*, not a panic, so `recover()` cannot catch it. (The existing `recover()` in `internal/send/response.go` from #229 is best-effort for *other* panics, not true cycle-safety — copying it here would be cargo-cult.)
- Documented in the CHANGELOG so operators see it consciously.

## Testing

- `go build ./...`, `go vet ./...` clean.
- Pre-commit gate green: `go fmt`, golangci-lint (`errcheck,govet,ineffassign,staticcheck,unparam,unused`), full test suite (incl. `pubsub` Redis integration via testcontainers).


adnaan:
Thanks — good catch on `topic_runtime.go:35`. That's a genuine miss (the issue's file list didn't enumerate it), and it's exactly the production wire-format class this PR targets. **Folded in** (`41fad52e`): the `topic_forbidden` WS envelope now uses `jsonutil.API.Marshal`.

On the completeness question more broadly — I ran a full-repo sweep and the remaining `encoding/json` importers are **deliberate exclusions**, now called out in the CHANGELOG so they're not mistaken for another follow-up:

- **`internal/keys/hash.go`** — uses `json.Marshal` as a *canonical byte-exact reference* for hash-key generation. Its documented `Wire-stability invariant` (guarded by `TestFormatHashPart_ByteEquivalentToJSON`) forbids any encoder whose output bytes could differ from `encoding/json`'s, since a diff would silently flip `RangeStreamState.Hashes` and break stream-mode transition fingerprints across upgrades. json-iterator is *mostly* byte-compatible but not contractually identical, so migrating here would be a latent correctness bug, not a consistency win.
- **`internal/fuzz/*`** (`ts_oracle.go`, `verifier.go`, `apply.go`) — the fuzz harness uses stdlib `encoding/json` as the **oracle it validates json-iterator against**. Swapping it to the library under test would defeat the harness's purpose.
- **`e2e/docker/app/main.go`** — an e2e test-fixture app, not the library.

Confirming the cyclic-state read too — agreed on all points; `recover()` genuinely can't catch a stack-overflow fatal error, so the CHANGELOG note (rather than copying the `internal/send/response.go` recover) is the honest surface for that edge case.

adnaan:
Addressed the two doc nits in `089a45dd` (CHANGELOG-only, no code change):
- Fixed the imprecise "Three callsites" — now lists the exclusions by category without a wrong count.
- Added `e2e/docker/app/main.go` to the exclusion list, noting it's a demo/e2e fixture app rather than library code, so the "production paths" framing is unambiguous.

Re: the `pubsub` Redis integration suite — confirmed green: it runs in the pre-commit gate (testcontainers spins up a real Redis) and again as the `pubsub` package in CI Unit Tests, both passing on every commit here.

Treating this as converged (no functional issues across the last two rounds) — merging.

Issue #462, opened 2026-06-20
Design: let State opt out of (or make lazy) the eager zero-arg method precompute


adnaan:
## Context

Split out from #255 (a PR #254 follow-up). #255 asked for a `lvt:"-"` struct tag to exclude a zero-arg method from eager precompute. That exact mechanism is **infeasible** — Go struct tags apply to fields, not methods — so this issue captures the real design decision. Deferred deliberately (P4, no users yet) rather than committing to a public API shape unilaterally.

## The underlying behavior

`BuildDataMap` (`internal/context/data.go`) eagerly evaluates **every** exported zero-arg method of a State struct on **every** render and stores the result in the data map — even methods the template never references. The code comment already warns: *"Avoid methods with side effects or expensive computations in State types."* An opt-out list patches the symptom; the model itself is the deeper question.

### 1. Should precompute stay eager, or become lazy?
`html/template` calls methods lazily (only when the template references `{{.Method}}`). LiveTemplate precomputes eagerly because it converts the struct to a `map` up front. A lazy approach (resolve methods on demand during eval) would eliminate the expensive/side-effecting-method footgun entirely, with no opt-out API needed — but it's a larger change to the data→map pipeline. Worth deciding first, because if precompute goes lazy, an opt-out list is moot.

### 2. If it stays eager, how to opt a method out? (no method tags in Go)
- **Interface opt-out** — State implements `LvtExcludeFromPrecompute() []string` returning method names to skip. ~15 lines + one type-assert per render. Simple, but stringly-typed (drifts if a method is renamed) and adds permanent public API.
- **Registration-based** — pass an explicit skip-list into `BuildDataMap` / a `Handle` option. Keeps it off the State type but threads config through several layers.
- **Field-marker** — a sentinel struct field whose tag lists excluded methods (e.g. `_ struct{} \`lvt:"exclude:ExpensiveReport"\``). Ugly, also stringly-typed.

## Recommendation

Decide layer 1 first. If eager stays, the interface opt-out is the lightest of the layer-2 options. No code until the mechanism is chosen.

## Related

- #255 (original, infeasible-as-titled)
- #257 (the reflect.New optimization — shipped separately; unrelated to this design)

adnaan:
[...]

Issue #255, opened 2026-03-26
Support lvt:"-" struct tag to opt methods out of eager evaluation


adnaan:
## Context

This follow-up task was identified during the review of PR #254.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/254
**PR Title:** fix: preserve struct methods in template data map
**Suggested by:** @claude-bot

## Task Description

PR #254 added eager precomputation of zero-arg struct methods into the template data map. All qualifying methods are called on every render, even if the template doesn't reference them.

Add support for a `lvt:"-"` struct tag on methods (or a naming convention) to opt specific methods out of eager evaluation. This would allow developers to exclude expensive or side-effect-producing methods from automatic precomputation.

Example:
\`\`\`go
type State struct {
    Items []Item
}

// This would be precomputed on every render (default)
func (s State) ActiveCount() int { ... }

// This would be excluded from precomputation
// (needs design decision on how to express this — struct tags don't apply to methods)
func (s State) ExpensiveReport() Report { ... }
\`\`\`

Design options:
- Convention-based: methods starting with underscore or a specific prefix
- Registration-based: explicit opt-out list passed to BuildDataMap
- Comment-based annotation (not idiomatic Go)

---
*This issue was automatically created from PR review comments.*

adnaan:
The mechanism this asks for (`lvt:"-"` on a method) isn't expressible in Go — struct tags apply to fields, not methods. Rather than commit to a stringly-typed opt-out API on an alpha with no users, the design decision is captured in **#462**, which also raises the prior question of whether the precompute should be eager at all vs lazy like `html/template` (a lazy model would remove the expensive-method footgun without any opt-out API). Deferring #255 pending that decision. The sibling optimization #257 (avoid `reflect.New` for value-receiver methods) is shipping separately and is unrelated.

adnaan:
Resolved via #482 (merged, squash `48c2b9a1`) through the design captured in #462. The literal ask here — a `lvt:"-"` tag on a method — is infeasible in Go (tags attach to fields, not methods), and the chosen solution (**eager-over-referenced precompute**) removes the need for any opt-out mechanism entirely: a State method the templates never reference is simply never precomputed. See #462 for the full rationale.

PR #482, opened 2026-07-09
perf(context): scope eager method precompute to referenced idents (#462, #255)

adnaan:
## What

Implements **Option A** for #462 (the design issue split out of #255): keep eager evaluation of zero-arg State methods, but only for methods a template actually references. Closes the eager-precompute footgun with **zero new public API** and **zero user annotation**.

Resolves the design question in #462; supersedes #255 (whose literal ask — a `lvt:"-"` tag on a method — is infeasible in Go, since tags attach to fields, not methods).

## Why

`BuildDataMap` converts the top-level State struct to a `map`, and `html/template` cannot auto-call methods on a map — so today **every** exported zero-arg method is eagerly evaluated on **every** render, even ones no template references. An expensive or side-effecting helper method that no template renders still runs. The doc comment already warned "avoid methods with side effects or expensive computations in State types"; this removes the footgun instead of just warning about it.

## How

- **`internal/parse/refs.go`** — `CollectReferencedIdents`: a data-free AST walk returning the set of every identifier that could reference a field/method: each element of every field/chain/variable chain **plus every string literal** (a literal may be a dynamic map key via `{{index . "X"}}`).
- **`internal/context/data.go`** — `BuildDataMap` / `ExecuteTemplateWithContext` gain a `precomputeAllow` set. Non-nil → only listed methods are evaluated (the allow-check also gates the `reflect.New` allocation); **nil → precompute all** (preserves original behavior for direct callers).
- **`template.go`** — the set is computed **once at parse time** in `parseInternal` via `parse.CollectReferencedIdentsFromTemplate(t.tmpl)`, stored on `Template`, and shared across per-session clones by value like `formSchema`/`wiredActions` (immutable after parse, no extra lock). `Funcs()` needs no invalidation — it re-applies the funcMap but does not re-parse.

## Soundness

Deliberate **over-approximation**: a top-level method is skipped only if its name appears in *no* FieldNode and *no* string literal across `t.tmpl.Templates()`. But if it appears nowhere, no template can retrieve `dataMap["Method"]` at all — so it can never silently render nil. Rendered output is byte-identical for templates that reference the methods they use.

## Measured impact (`BenchmarkPrecompute`, arm64, `-count=10`)

| case | sec/op | B/op | allocs/op |
|---|---|---|---|
| `eager_all_methods` (before / nil set) | 8.109µ ± 2% | 1,608 | 15 |
| `referenced_only` (expensive methods unreferenced) | ~1.02µ | 440 | 5 |
| `all_referenced` (all precomputed) | 8.184µ ± 1% | 1,608 | 15 |

~8× faster and ~3.6× less memory when expensive methods go unreferenced. `all_referenced` vs `eager_all` is statistically indistinguishable — `benchstat`: `~ (p=0.159 n=10)`, identical B/op and allocs/op — so the allow-set lookup is **not** a regression. (The committed single-run `baseline.txt` shows a small noise delta on these two; the `-count=10` benchstat above is the robust comparison.)

## Documented residuals

- **Conditional references still compute:** `{{if .Show}}{{.Expensive}}{{end}}` still calls `Expensive()` every render — the name is in the template text.
- **Name collisions harmlessly over-compute:** a method name matching an unrelated field/string literal is still (safely) precomputed.

Both noted in `CHANGELOG.md` and `docs/references/controller-pattern.md`.

## Tests

- `internal/parse`: table tests for the collector (fields, chains, `if`/`range`/`with`, `{{index . "X"}}`, associated-template union, nil/empty).
- `internal/context`: allow-set scoping (only-referenced / empty-skips-all / nil-precomputes-all), plus the comparative benchmark.
- root package: footgun proof (unreferenced method never called via the real `Execute` path), **clone-path scoping survival** (production renders through clones), and **`{{index . "X"}}` end-to-end render**.

Full suite green locally.


adnaan:
Thanks for the thorough review. Addressing the one actionable note:

**Benchmark "not a regression" claim** — confirmed with a `-count=10` benchstat run comparing `eager_all_methods` vs `all_referenced`:

```
              │  eager_all   │        all_referenced        │
              │    sec/op    │   sec/op     vs base         │
PrecomputeX-8   8.109µ ± 2%   8.184µ ± 1%   ~ (p=0.159 n=10)
B/op            1.570Ki ± 0%  1.570Ki ± 0%  ~ (p=1.000)  (all samples equal)
allocs/op       15.00 ± 0%    15.00 ± 0%    ~ (p=1.000)  (all samples equal)
```

`p=0.159` → statistically indistinguishable; B/op and allocs/op are identical. The ~3.4% delta in the original table was single-run noise (the committed `baseline.txt` is an intentionally single-run snapshot). I've updated the PR description's table to the robust `-count=10` numbers and the p-value so the "not a regression" claim is backed by the data rather than a noisy single run.

The other two observations (string-literal over-approximation breadth; conditional-branch caveat) are the documented, accepted safe-over-compute tradeoffs — no change intended there, as you noted.

PR #69, opened 2025-12-07
feat!: Controller+State pattern - separate dependencies from data

adnaan:
## Summary

This PR implements a complete API redesign that separates Controller (dependencies, singleton) from State (pure data, cloned per session). Inspired by Phoenix LiveView's socket.assigns pattern.

### Key Changes:
- **New Context type**: Replaces ActionContext with unified Context for all lifecycle methods
- **Controller+State separation**: Dependencies on Controller, pure data in State
- **New method signatures**: `func (c *Controller) Action(state State, ctx *Context) (State, error)`
- **Lifecycle methods**: Mount(), OnConnect(), OnDisconnect()
- **AsState wrapper**: Generic wrapper for state serialization

### Breaking Changes:
- `Handle(store)` → `Handle(controller, AsState(state))`
- `ActionContext` → `Context`
- `Dispatch()` → `DispatchWithState()`
- Method signature: `func (s *Store) Action(ctx *ActionContext) error` → `func (c *Controller) Action(state State, ctx *Context) (State, error)`

### Benefits:
- No more accidental sharing of session-specific data
- Clear contract for what gets cloned vs shared
- Serializable state enables distributed session stores
- Type-safe state handling

## Test plan
- [x] All unit tests pass
- [x] All integration tests pass
- [x] E2E tests in lvt repository pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #484, opened 2026-07-10
feat(template): add WithParseFS for embed.FS templates + boilerplate-reduction proposal (#483)

### LINKED ISSUE #483: Scan docs/examples and suggest boilerplate reduction for users of livetemplate framework
Scan everything in /home/adnaan/code to deep dive into examples of livetemplate server and client usage and suggest improvements to the framework to reduce boilerplate for the user. the improvements could be both: adding or removing features including moving common patterns to the framework itself.

adnaan:
Closes #483.

Scans the livetemplate examples plus five independent consumer apps (`examples/`, `docs/examples/`, devbox-dash, tinkerdown, checklistkit, prereview) for boilerplate users write repeatedly, and ships the two unambiguous low-risk wins. The higher-impact, design-heavy items are written up as greenlight-ready proposals rather than implemented.

## What's in this PR

**Proposal doc** — `docs/design/boilerplate-reduction.md`. Ranked, evidence-backed findings (each cites ≥2 independent apps with file anchors), tiered:
- **Tier A — shipped here:** A1 `WithParseFS`, A2 request-context guidance.
- **Tier B — proposal + design sketch (await per-item greenlight):** B1 live-session fanout primitive, B2 first-class client-asset handler, B3 boot/route helpers.
- **Tier C — noted:** C1–C9, including a removal pass (C4: two redundant session-store options).

**A1 — `WithParseFS(fsys fs.FS, patterns ...string)` + `(*Template).ParseFS`.** Embedded-template apps currently hand-roll a ~22-line `os.MkdirTemp` spill because `WithParseFiles` only accepts OS filepaths — repeated ~identically across 10 docs recipes and two real apps that label it a "workaround" by name (`prereview`, `tinkerdown`). The framework already parses `embed.FS` internally for components, so the machinery existed. `ParseFiles` and `ParseFS` now share a `parseSources` core (no behavior change to `ParseFiles`). Precedence in `New()`: `WithParseFS` > `WithParseFiles` > auto-discovery.

**A2 — request-context guidance** in `docs/references/controller-pattern.md`. `*livetemplate.Context` has embedded `context.Context` since #69, but apps universally reach for `context.Background()` in actions (discarding request cancellation/deadline/tracing). Docs-only — no new API; an explicit `ctx.Context()` accessor was deliberately *not* added (redundant given the embed).

## Not in this PR (deliberate)

- **Tier B/C** stay proposal-only pending per-item greenlight.
- **Sibling-repo companions** land as coordinated follow-ups: the `examples/` `context.Background()`→`ctx` swap (doable now — that repo is in `go.work`), and converting `docs/examples` recipes to `WithParseFS` (**release-gated** — the `docs` repo pins a *published* livetemplate, so it can only land after a release ships `WithParseFS`).

## Test plan

- `parsefs_test.go`: `embed.FS` + `fstest.MapFS` parsing, error paths (no patterns / no match / bad glob), `New()` precedence, and a **golden-parity check** proving `ParseFS` renders byte-identically to `ParseFiles` (wrapper id normalized).
- Full suite green via pre-commit (fmt + golangci-lint + `go test ./...`, incl. Redis/pubsub integration).


adnaan:
Thanks for the review. Addressed the actionable items in `4b97e159`:

- **API-reference table** — added a `WithParseFS` row next to `WithParseFiles` (`docs/references/api-reference.md`).
- **CHANGELOG** — added an Unreleased `### Added` entry for `WithParseFS`/`ParseFS` (`#483`).
- **`ParseFS` doc comment** — softened the `html/template.ParseFS` comparison (stdlib has no first-match-is-main concept) and called out `fs.Glob`'s lexical ordering for a single wildcard pattern.
- **Test gap** — added `TestWithParseFS_WithComponentTemplates`, which exercises the `parseSources` clone-and-extend branch (`t.tmpl != nil`) via the `WithParseFS` path with a component set, executing `{{template "test:greeting:v1"}}` from the embedded main template.

The doc-comment nits (single-glob main-template selection) are folded into the same doc block rather than left as separate notes.

### On the `Test Docs Examples against Core Changes` failure

This is a **pre-existing flake in the docs `chat` e2e**, not a regression from this PR. Only `TestChatE2E` fails (`UI_Standards` + `Join_Flow`); every other docs example — including the more complex `avatar-upload` (uploads/ranges/trees) — passes. Reasoning:

- **Mechanism rules it out:** this PR only touches the `ParseFiles`/`ParseFS`/`parseSources`/`New()`-precedence path. Every passing example loads templates through that same refactored core, and the golden-parity test proves `ParseFS` renders byte-identically to `ParseFiles`. If `parseSources` were broken, the failures wouldn't be confined to one example's WS-timing assertions.
- **The failures are timing-shaped:** `Join_Flow` times out waiting ≤5s for `.messages` after a WS round-trip in Dockerized Chrome, and `UI_Standards` flags a transient `[style]` div (the client's connecting/loading indicator snapshotted mid-connect) — the chat template source has no inline styles.
- This same cross-repo check is green on `main`.

The fresh CI run from this push re-exercises it.

adnaan:
Round 2 items addressed in `c4738f7d` (both were flagged non-blocking):

- **Overlapping glob double-parse** — added a one-line callout to the `ParseFS` doc comment noting patterns resolve in order with no dedup, so a file matched by two overlapping patterns is parsed twice; pass non-overlapping patterns to avoid it.
- **Multi-match order sensitivity** — added `TestParseFS_SingleGlobLexicalMain`: one wildcard pattern matching two files, asserting the lexically-first match is the rendered main template and later matches compose in as `{{define}}` partials.

Not pursuing a dedup pass (parity with `ParseFiles`' existing duplicate-filename behavior, and the doc callout preempts the confusing error) — consistent with your "not worth a dedup pass for this PR" note.

adnaan:
Round 3 — thanks, both were non-blocking:

- **Doc note added** (`365f8903`): the `WithParseFS` doc comment now states that `WithTemplateBaseDir`/`WithIgnoreTemplateDirs` have no effect when `WithParseFS` wins, since it bypasses auto-discovery.
- **Declining the "auto-discovery actually skipped" test.** The precedence is structural, not incidental: `New()` is a straight `if config.TemplateFS != nil { ParseFS } else if len(config.TemplateFiles) == 0 { auto-discover } else { ParseFiles }`. `WithParseFS` sets `TemplateFS`, so the first branch wins and the auto-discovery `else if` is unreachable — there's no code path where auto-discovery could run alongside it. A real test would have to scan a cwd/base-dir with planted templates (environment-sensitive, flake-prone) to prove a negative that the control flow already guarantees. Not worth the fragility.

Both this round's items and the prior round's were non-functional polish (no bug found), so treating this as converged. The one CI blip earlier — `Test Docs Examples against Core Changes` failing on `TestChatE2E` — was a pre-existing flake (WS-timing + a transient loading-indicator inline-style in the chat example); it's passed green on all subsequent pushes.

Issue #330, opened 2026-04-11
htmx.org patterns example


adnaan:
Create a single multi page livetemplate example(in ./examples) to implement all of the patterns in https://htmx.org/examples/. Suggest additional patterns not possible in htmx or not covered(additional research in https://hexdocs.pm/phoenix/overview.html , https://pragmaticstudio.com/phoenix-liveview

Implementation notes:
- Always tier 1 before tier 2: https://github.com/livetemplate/livetemplate/blob/main/docs/references/progressive-complexity-reference.md
- See CLAUDE.md for implementation guidelines
- A main page with a list of patterns and a short note. Categorise patterns in groups. suggested but propose a better categorisation if  loading, forms, records, display, realtime, keyboard etc
- One page per example
- Analyse each pattern to see how it can be implemented as a livetemplate pattern. If a pattern cannot be implemented in the livetemplate way: ./livetemplate/blob/main/docs/design/FIRST_PRINCIPLES.md, ./livetemplate/docs/references/progressive-complexity-reference.md due to a missing feature, we skip that example and implement the missing feature in ./livetemplate and/or ./client before coming back to the pattern. No custom javascript ever !
- Since this is a large scale implementation, the delivery for this issue is a multi session proposal doc with implementation tracker which can be followed across multiple sessions. See an example for a large scale proposal: https://github.com/livetemplate/livetemplate/blob/main/docs/archive/proposals/attribute-reduction-proposal.md. This would mean that the proposal can be spread across multiple repos apart from examples repo to implement a missing feature if need be. We will first review and merge the doc in ./livetemplate/docs/proposals folder. An LLM should be able to then implement this proposal in multiple sessions.

PR #486, opened 2026-07-11
feat(client): pin the wire-compatible client version + template helpers (#483 B2)

adnaan:
## What

**B2 of the boilerplate-reduction work (#483).** Every shipped example (15 of them) plus real apps import a *test* package — `e2etest "github.com/livetemplate/lvt/testing"` — into their production `main.go` to serve the browser client bundle, and that helper fetches from `unpkg.com/@livetemplate/client@latest` at runtime. Two smells:

1. A **test package in production** serving.
2. An **unpinned client version** (`@latest`) with **no server↔client version handshake** — so a browser can load a client *newer* than the server it talks to. The client's own CHANGELOG documents this footgun: the `__navigate__` action is *"a no-op on server versions before livetemplate#344 — deploy the server update before or simultaneously with this client version."* These are 0.x releases, where minors are allowed to break the wire.

## Approach: pin the version, don't bundle

Embedding the ~140KB bundle into core was considered and rejected: for a **library**, a gitignored `//go:embed` target breaks every downstream `go get`, and *committing* the blob sheds core's deliberate zero-asset-leaf identity (its `go.mod` depends on neither `lvt` nor `client`). That pays a cost on every consumer to serve a minority offline need that prereview already self-solves by vendoring.

Instead core owns the one thing it uniquely knows — **which client version is wire-compatible with it**:

```go
const ClientVersion   = "0.16.5"   // moves in lockstep per release
const ClientScriptURL = "https://cdn.jsdelivr.net/npm/@livetemplate/client@" + ClientVersion + "/dist/livetemplate-client.browser.js"
const ClientStyleURL  = "https://cdn.jsdelivr.net/npm/@livetemplate/client@" + ClientVersion + "/livetemplate.css"
```

Pinned to `0.16.5` because that is what `@latest` resolves to today — **behavior-preserving**. This also *removes* the toil `@latest` appeared to save: the version now moves only when you deliberately `go get -u`, in lockstep with the compatible server.

To let templates consume the pinned URL with **zero per-app wiring** (no State field, no manual `Funcs` registration), `New` seeds two framework template funcs into every `Template`'s `FuncMap` before parse:

```html
<link rel="stylesheet" href="{{ lvtClientStyleURL }}">
<script src="{{ lvtClientScriptURL }}" defer></script>
```

Offline / air-gapped / CSP-strict deployments self-host (vendor `@livetemplate/client@<ClientVersion>`, serve same-origin, write their own tag). The `lvt/testing` helpers **stay** — real E2E tests keep using them; we only stop shipping them in production examples.

## Tests (offline — a network assertion would flake whenever the client publishes ahead of core)

- `ClientVersion` is bare semver.
- URLs are HTTPS, carry the pinned version, and **never** fall back to `@latest` (the regression guard that locks the decision).
- URLs point at the correct artifact paths (`/dist/...browser.js`; CSS at package root).
- End-to-end render **through a Clone** (production renders via per-session clones) proving the funcs parse and emit the URLs.

## Scope

**Core-only.** The visible payoff — converting the 15 examples off `e2etest` and the browser/wire-compat verification — is **release-gated**: the docs repo pins a *published* livetemplate, so examples can't reference `ClientVersion`/the funcs until this ships in a release. That conversion is the follow-up PR, exactly like A1 (`WithParseFS`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-Nothing yet.
+### Added
+
+- `ClientVersion`, `ClientScriptURL`, and `ClientStyleURL` exported constants pin the `@livetemplate/client` browser bundle this LiveTemplate release is wire-compatible with, plus two framework-seeded template functions — `lvtClientScriptURL` and `lvtClientStyleURL` — that render those URLs. Templates can now reference `{{lvtClientScriptURL}}` / `{{lvtClientStyleURL}}` with no per-app wiring (the funcs are seeded into every template's FuncMap in `New`, before any parse path runs, so they resolve in full-HTML documents, fragments, and component templates alike; a user `Funcs` call still overrides them by key). This replaces the previous pattern of hardcoding an unpinned `@latest` CDN URL — which was unsafe because there is no runtime server↔client version handshake, so a client-only release could ship a wire-protocol change to browsers still talking to an older server. Pinning moves the client version only on a deliberate `go get -u`, in lockstep with the compatible server. `ClientVersion` is pinned to the release `@latest` resolves to today, so behavior is preserved. Self-hosters (offline / air-gapped / CSP-strict) vendor `@livetemplate/client@<ClientVersion>` and serve it from their own origin instead. (#483)
 
 ## [v0.17.0] - 2026-07-10
 
--- a/README.md
+++ b/README.md
@@ -102,9 +102,11 @@ func main() {
     <button name="decrement">-</button>
 </form>
 
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@livetemplate/client@latest/livetemplate.css">
-<script defer src="https://cdn.jsdelivr.net/npm/@livetemplate/client@latest/dist/livetemplate-client.browser.js"></script>
+<link rel="stylesheet" href="{{lvtClientStyleURL}}">
+<script defer src="{{lvtClientScriptURL}}"></script>
 ```
+
+`lvtClientStyleURL` / `lvtClientScriptURL` are framework-provided template functions that render the CDN URL for the `@livetemplate/client` release this LiveTemplate version is wire-compatible with — so the browser client stays pinned in lockstep with the server instead of drifting on `@latest`. To self-host, replace them with your own tag pointing at a vendored `@livetemplate/client@<version>`.
 
 **3. Run it**
 
--- a/client_assets.go
+++ b/client_assets.go
@@ -1,0 +1,56 @@
+package livetemplate
+
+import "html/template"
+
+// ClientVersion is the version of the @livetemplate/client browser bundle that
+// this release of livetemplate is wire-compatible with.
+//
+// It moves in lockstep with each livetemplate release, so applications never
+// hand-maintain a client version: upgrading the livetemplate dependency
+// (go get -u github.com/livetemplate/livetemplate) automatically selects the
+// matching client. Pinning to this version — rather than the CDN's "@latest"
+// tag — prevents a client-only release from silently shipping a wire-protocol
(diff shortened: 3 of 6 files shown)


adnaan:
Thanks — addressed the two doc-completeness gaps in the latest push (802bfeab):

1. **README `@latest` anti-pattern — fixed.** `README.md` quick-start (`counter.tmpl`) now uses `{{lvtClientStyleURL}}` / `{{lvtClientScriptURL}}`. That snippet is an auto-discovered livetemplate template (`New("counter")`), so the funcs resolve there — it now demonstrates the pinned mechanism instead of contradicting it. Added a one-line note explaining the funcs and the self-host escape hatch.
2. **CHANGELOG — added.** New `### Added` entry under `[Unreleased]` covering the three constants + two template funcs, following the existing verbose style and referencing #483.
3. **SRI — tracked, deliberately deferred (declining for this PR).** SRI belongs with the example-conversion follow-up, not here: this PR ships no rendered `<script>`/`<link>` tag of its own (the README snippet is illustrative), and an `integrity=` hash pins to a *file digest* that must be regenerated in lockstep with every `ClientVersion` bump — that hardening + its regeneration step lives naturally alongside the examples that actually emit the tags. It is recorded in the "Follow-ups" note of the B2 section in `docs/design/boilerplate-reduction.md` so it does not silently drop.

adnaan:
Round 2 — thanks, this review found no blocking issues. On the three non-blocking notes:

1. **Collision/override caveat — added.** Pushed a one-line godoc caveat on `(*Template).Funcs` noting that funcs merge by name, so a user func overrides a framework-seeded one (e.g. `lvtClientScriptURL`/`lvtClientStyleURL`) — surfacing the behavior for readers skimming `Funcs` rather than `client_assets.go`. Kept it generic (no per-func coupling).
2. **SRI** — agreed, out of scope; version pinning closes the wire-drift risk and jsdelivr per-version URLs are immutable. Tracked in the design-doc Follow-ups.
3. **Auto-discovery in the test** — correct, that `New("name")`-only pattern is the dominant convention in `template_test.go`; no new risk. No action.

adnaan:
Round 3 — no blocking issues, thanks. Two notes:

1. **Override test — added.** New `TestClientURLTemplateFuncsAreOverridable` registers a `Funcs` override for `lvtClientScriptURL`, parses + clones + executes, and asserts the override wins *and* the framework default does not leak through. This directly locks the behavior the Round-2 godoc/CHANGELOG promise, closing the coverage gap you flagged.
2. **`docs/proposals/patterns.md:2075` still says "use @latest, do not pin"** — declining to touch it here. That file is `**Status:** Proposed` (issue #330) and documents the tradeoff for the **archived** top-level `examples/` repo, not authoritative current guidance. The `@latest`→`{{lvtClientScriptURL}}` reconciliation belongs with the release-gated examples-conversion follow-up, where the runnable examples actually flip; editing an archived proposal doc from this core-only PR would be scope creep. Noting it here so it is tracked rather than silently dropped.

adnaan:
Round 4 — no blocking issues, and no new functional findings. Both notes are non-blocking framing points; addressing by reply rather than another push (per the repo's bot-loop convergence guidance — successive rounds no longer surface a functional issue):

1. **`package livetemplate_test` (external/black-box) — intentional, confirmed.** As you note, it is a deliberate deviation: the point of B2 is a zero-wiring *consumer* experience, so the tests exercise `ClientVersion`/`ClientScriptURL`/`ClientStyleURL` and the seeded funcs strictly through the exported surface, exactly as an app would. Recording the intent here so it does not read as accidental drift.
2. **Semver regex rejects pre-release suffixes — accepted as-is.** The project pins GA releases; if a pre-release client ever needs pinning, `TestClientVersionIsSemver` loosens in that same change. No action now.

Considering this converged: R1 surfaced real gaps (README `@latest`, CHANGELOG), R2/R3 were a doc-caveat + a coverage test for that caveat, R4 is framing-only. Holding here pending maintainer signoff to merge.

PR #493, opened 2026-07-14
docs(proposal): recursive {{template}} support design (Tier C · C8)

adnaan:
## Tier C · C8 — recursive `{{template}}` support (design proposal)

Part of the boilerplate-reduction sweep (issue **#483**, now closed — C8 was the one item with a genuine capability gap and was spun out into its own design, tracked here).

**Status: approved design of record.** The maintainer signed off on the three headline decisions in-session (2026-07-15) — this PR records the approved design; it does not itself request a decision. **No implementation lands in this PR** (docs-only); Phase 1 is ready to implement next.

### The gap
livetemplate flattens `{{template "name" .}}` textually at parse time (`internal/parse/flatten.go`, `walkAndFlatten`) with no cycle/depth guard, so a self-referential template **stack-overflows at `Parse`**. prereview hits this directly: it wants a reactive folder tree, can't use a recursive livetemplate template, and drops the tree to a standalone `html/template` injected as opaque `template.HTML` — losing incremental diffing. (tinkerdown corroborates the recursive-tree-UI *pattern* but serves static HTML, so it's a latent case, not a second bug victim — see the proposal's Evidence section.)

### The proposal
`docs/proposals/recursive-templates-proposal.md`. Approved approach: **selectively** route *recursive* invocations through runtime invocation → nested `TreeNode` (keeping non-recursive `{{template}}` flattening byte-for-byte unchanged), reusing the existing `{{range}}` nested-tree machinery — no new wire ops.

The "no new client changes" claim is **verified against `github.com/livetemplate/client` v0.16.5** (the client has no fingerprint concept; `deepMergeTreeNodes` caches statics positionally and recurses depth-agnostically). All prereview/tinkerdown citations were **verified against the actual repos** on this machine.

### Decisions recorded (maintainer sign-off, 2026-07-15)
1. **Selective**, not uniform runtime invocation (minimal blast radius).
2. **Max-depth guard, erroring uniformly** on initial render and live update (over a pointer-visited set; depth increments before the check so data pointer-cycles are still bounded).
3. **`data-key` falls back to content-hash** like `{{range}}` (explicit key preferred for large trees).

The proposal's § Decisions covers these (items 1/2/4, keeping their original open-question numbers); § Remaining design notes keeps items 3/5/6 (implementation invariants / scope boundaries, not open decisions).

### Scope
- Proposal doc + `docs/design/boilerplate-reduction.md` C8 entry → 📝 DESIGN APPROVED.
- 7 phased implementation sketch, each with a hard test gate (correctness unit + chromedp e2e with console/server/WS/HTML capture + `go test -bench` benchmarks). Phases 6–7 are companion migrations (prereview escape-hatch removal; tinkerdown nav→reactive, honestly scoped as larger/optional).

### DIFF
--- a/docs/design/boilerplate-reduction.md
+++ b/docs/design/boilerplate-reduction.md
@@ -351,11 +351,17 @@ example would return to its documented ~10 lines; checklistkit's 5 builders coll
   `{{.lvt.AriaInvalid "field"}}` is exactly this pattern (`.lvt` is a struct value in the map).
   → **Resolution: document the view-helper pattern (guide + example), keep a regression anchor.
   No core API change.**
-- **C8. Recursive `{{template}}` support (cross-app: prereview + tinkerdown).** livetemplate
-  flattens `{{template}}` calls at parse time, which overflows on recursion, so a recursive
-  file tree drops to a standalone `html/template` injected as `template.HTML`
-  (prereview `gitdiff/filetree.go:18-25`, noting *"the same native-`<details>` approach
-  tinkerdown uses"*). → Propose recursive-partial support.
+- **C8. Recursive `{{template}}` support (cross-app: prereview + tinkerdown).** 📝 **DESIGN
+  PROPOSED — awaiting greenlight.** livetemplate flattens `{{template}}` calls at parse time with
+  no cycle guard, so a self-referential template stack-overflows during `Parse`; a recursive file
+  tree drops to a standalone `html/template` injected as opaque `template.HTML` (prereview
+  `internal/review/filetree.go`, noting *"the same native-`<details>` approach tinkerdown uses"*),
+  which removes the whole subtree from reactive diffing. The one Tier C item with a genuine
+  capability gap and ≥2 hand-rolled consumers → **full design written**:
+  [`docs/proposals/recursive-templates-proposal.md`](../proposals/recursive-templates-proposal.md).
+  Recommended approach: route recursive invocations through the runtime nested-`TreeNode` path
+  `{{range}}` already uses (bounded by data depth, not parse-time expansion). Needs maintainer
+  greenlight before implementation.
 - **C9. Static-asset passthrough alongside the `/` LiveHandler.** ✅ **RESOLVED (docs, no new
   API).** The original pitch was a framework static-passthrough wrapper. On audit the common
   case — assets under a **known prefix** — is already handled by plain `net/http`:
--- a/docs/design/boilerplate-reduction.md
+++ b/docs/design/boilerplate-reduction.md
@@ -352,7 +352,7 @@ example would return to its documented ~10 lines; checklistkit's 5 builders coll
   → **Resolution: document the view-helper pattern (guide + example), keep a regression anchor.
   No core API change.**
 - **C8. Recursive `{{template}}` support (cross-app: prereview + tinkerdown).** 📝 **DESIGN
-  PROPOSED — awaiting greenlight.** livetemplate flattens `{{template}}` calls at parse time with
+  APPROVED (2026-07-15) — Phase 1 next.** livetemplate flattens `{{template}}` calls at parse time with
   no cycle guard, so a self-referential template stack-overflows during `Parse`; a recursive file
   tree drops to a standalone `html/template` injected as opaque `template.HTML` (prereview
   `internal/review/filetree.go`, noting *"the same native-`<details>` approach tinkerdown uses"*),
@@ -360,8 +360,10 @@ example would return to its documented ~10 lines; checklistkit's 5 builders coll
   capability gap and ≥2 hand-rolled consumers → **full design written**:
   [`docs/proposals/recursive-templates-proposal.md`](../proposals/recursive-templates-proposal.md).
   Recommended approach: route recursive invocations through the runtime nested-`TreeNode` path
-  `{{range}}` already uses (bounded by data depth, not parse-time expansion). Needs maintainer
-  greenlight before implementation.
+  `{{range}}` already uses (bounded by data depth, not parse-time expansion). Approach A
+  (selective), a max-depth guard erroring uniformly, and content-hash `data-key` fallback are all
+  decided (see the proposal's § Decisions); Phase 1 (crash → `ParseError`) is independently
+  shippable and up next.
 - **C9. Static-asset passthrough alongside the `/` LiveHandler.** ✅ **RESOLVED (docs, no new
   API).** The original pitch was a framework static-passthrough wrapper. On audit the common
   case — assets under a **known prefix** — is already handled by plain `net/http`:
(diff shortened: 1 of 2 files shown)


adnaan:
Thanks — addressed in `8eac8342`:

- **Duplicate `5.` in Open questions** → fixed; dynamic-invocation is now `6.` (the "Open question 5" reference in the PR body correctly points at the depth-change item).
- **Max-depth behavior on a live update vs initial render** (your one design gap) → folded into Open question 2 and Phase 4: recommend erroring uniformly on both paths so behavior doesn't diverge, flagged for confirmation since a runaway *update* shouldn't necessarily blank a working page.

The other three points are design input for the maintainer discussion this PR solicits, and I've left them for that sign-off rather than pre-deciding them in the doc:
- **Phase 1 can land independently** — agreed, and the doc already frames it that way ("independently shippable", turns the stack-overflow into a `ParseError`); it's a real crash-hardening fix that doesn't need Approach A greenlit.
- **`data-key` fallback like `{{range}}`** — that's the doc's current recommendation (Open question 4); consistent with the CLAUDE.md range `data-key` guidance you cite.
- **Cross-repo citations** (`prereview`/`tinkerdown`) — can't be bot-verified from this repo; the "no client changes" claim was verified against `client` v0.16.5 as the PR body notes.

No further functional issues outstanding — holding for maintainer sign-off on the open questions.

adnaan:
Good catch — this was a real design bug, fixed in `232dbc71`.

You're right: the worked example's `{{if .IsDir}}…{{end}}` means a leaf (`IsDir=false`) and a directory (`IsDir=true`) build **structurally different** trees → distinct fingerprints, so "cache once per template name" would conflate them. Since `GetStructureFingerprint()` is per-instance and the fingerprint is content-independent, the honest optimization is actually *simpler* than what I'd written:

- A recursive template emits a **small, bounded set of branch-shapes** (leaf-shape, dir-shape), not one shape and not O(depth) shapes.
- Nodes of a given shape already share a fingerprint **value** by equality — no per-name cache needed.
- The **existing per-instance lazy cache** already avoids recomputation.

So Section 5 now says the design adds **no** fingerprint caching at all; it must only *not* introduce a per-name cache that collapses distinct branch-shapes. Updated Phase 4 (added a leaf⇄directory branch-flip test) and the Scope list (`fingerprint.go` reused as-is) to match.

adnaan:
Fixed in `25d339c6` — you're right on both.

- **"No recursion-unsupported marker" was wrong** — the matrix has the `Circular template references | ❌ | … infinite loop` row, inside my own cited range. Corrected to the accurate (and stronger) point: the marker *exists* but **mislabels the failure** — it's not a runtime infinite loop, it's a **stack overflow during `Parse`** — and there's no implementation guard. Phase 1 both fixes that doc row and turns the crash into a `ParseError`. Also swapped the drift-prone `:174-179` line range for a content anchor (per the project's grep-anchor convention).
- **Phase 3 parity** — clarified it means HTML *content* parity, not wire-format parity (the wire format deliberately differs: nested `TreeNode` vs inlined string).
- **OQ2 as the maintainer's first answer** — agreed; noted it drives whether Phase 4's guard is a hard-error path or needs a partial-render fallback.

No functional issues outstanding. Holding for maintainer sign-off on the open questions (selective-vs-uniform, depth-guard style + live-update behavior, `data-key` fallback).

adnaan:
Converged — this round surfaced no new functional issue; all prior fixes verified against `main`. Addressed the one remaining nit (matrix-correction-is-deferred-to-Phase-1) by adding a note to the PR description rather than pushing, so we don't spin another review round on a cosmetic point.

Two spin-off doc items you flagged that are **out of scope here** and I'll track separately, not bundle:
- CLAUDE.md's `TemplateInvokeConstruct` in the "Key Constructs" list has no corresponding Go type (grep → 0 hits) — a standalone CLAUDE.md fix.
- The `template-support-matrix.md` row wording lands in Phase 1, not this PR.

This PR is now ready for **maintainer sign-off on the open questions** (selective-vs-uniform; depth-guard style + live-update behavior; `data-key` fallback). No implementation lands until that decision.

adnaan:
Two rounds of follow-up from the maintainer, both landed:

**1. Cross-repo citations — now verified against the actual repos** (both prereview and tinkerdown are on this machine; the review bot couldn't reach them, but I can):
- **prereview** — `filetree.go` comment, `FileBrowserHTML` signature, and `filetree.tmpl`'s `treeNode` recursion + `data-key` + native `<details>` all confirmed. It's the **one direct consumer**: a livetemplate reactive tree forced into an opaque `template.HTML` slot.
- **tinkerdown** — corrected the citation (the recursive renderer is `writeNavNode` at `server.go:3660`, *not* `buildNavPageNode`, which only builds the data structure). More significantly: tinkerdown serves pages as **static HTML** (`servePage`: *"just serve the static HTML"*), so its nav never hit the flatten bug — stdlib `html/template` handles recursion fine. **Reframed tinkerdown from "second victim" to "corroborating pattern / latent case"**; C8's justification now rests on the objective crash + prereview alone, not on consumer count. Also fixed a `$`-rebinding overclaim (prereview's only `$` is in the non-recursive wrapper).

**2. Implementation phases expanded** per the maintainer's direction:
- **Per-phase test matrix** as a hard gate: correctness unit tests + chromedp e2e (console/server/WS/HTML capture per CLAUDE.md) + `go test -bench` performance benchmarks (render O(nodes), update wire-size vs the opaque-`template.HTML` baseline).
- **Companion migrations split out**: **Phase 6** removes prereview's escape hatch (true simplification); **Phase 7** migrates tinkerdown's nav to a reactive recursive template — honestly scoped as a larger, optional migration (it entails making the static nav reactive) that proves the feature on a second independently-authored app.

Decisions on the three open questions are recorded in § Decisions (selective; max-depth erroring uniformly; `data-key` content-hash fallback).

adnaan:
Both points addressed in `3ce860af`:

- **Data pointer-cycle vs deep-but-finite** (your substantive one) — added the invariant to Phase 4 that makes max-depth sufficient: depth is incremented **before** the guard check on every level, so a self-referential `.Children` pointer cycle (a loop shorter than max-depth) still trips the ceiling and errors rather than hanging. Documented this as the deliberate tradeoff behind Decision 2 (max-depth over pointer-visited: simpler, at the cost of also erroring on a legitimately-deep finite tree — hence configurable), and added a self-referential-pointer unit test to the acceptance list. Approach A's Con now names both failure shapes.
- **OQ numbering** — added a one-line note in § Decisions explaining the 1/2/4 vs 3/5/6 split (decided items keep their original open-question numbers).

No functional issues outstanding. Design is ready for Phase 1.

adnaan:
On the approval-status point — you're right that the doc and the old PR body talked past each other, and I've reconciled the **PR body** (not the doc) since the sign-off is genuine:

- The three decisions **were** signed off by the maintainer in-session (2026-07-15) — so the doc's `Status: Approved` is truthful, not a pre-declared approval. The PR body previously still carried the stale "Requested from review" framing from when it was opened; I've reworded it to record the decisions as approved rather than pose them as open, so the two agree. Option 2 from your review.
- **"5 open questions" → 6**: reconciled — the body no longer cites a count; the proposal has 6 numbered items (1/2/4 decided, 3/5/6 design notes).
- **#483 closed**: noted in the body — C8 was the one item with a real capability gap and was spun out into this standalone design; the sweep issue closed.

Citations all re-verified against `main` on your end and against the actual prereview/tinkerdown repos on mine. No functional issues outstanding — this is the approved design of record, ready for Phase 1.

PR #494, opened 2026-07-16
feat(parse): reject recursive {{template}} at parse time (Tier C · C8 Phase 1)

adnaan:
## Tier C · C8 — Phase 1: reject recursive `{{template}}` at parse time

First, independently-shippable phase of the [recursive-templates design](https://github.com/livetemplate/livetemplate/blob/main/docs/proposals/recursive-templates-proposal.md) (issue #483, item C8).

### The bug
livetemplate inlines every `{{template "name" .}}` invocation textually during flattening (`internal/parse/flatten.go`, `walkAndFlatten`). A self-referential template re-entered its own body with **no cycle check**, writing an unbounded template string and **stack-overflowing during `Parse`** — a fatal, unrecoverable crash triggered by template source alone (a DoS-hardening gap).

### The fix
An **active-path cycle guard**. `walkAndFlatten` now threads a stack of the template names currently being inlined on the path from the entry point to the current node. `checkFlattenCycle` rejects a `{{template}}` whose name is already on the stack with a `ParseError` naming the cycle (`treeNode -> treeNode`, `a -> b -> a`) instead of overflowing.

- **Active-path set** (push on descend, implicit pop on return) — *not* a global visited-set — so the same template invoked twice on non-nested paths (a **diamond**) still flattens; only genuine self-reference is rejected.
- Stack seeded with the entry point's name so a self-referential entry point is caught on first invocation.
- Non-recursive composition keeps the existing flatten path **byte-for-byte** (verified by test).

### Scope
- Turns the crash into a diagnosable error and lays the parse-time detection Phases 2–5 build on (runtime nested-`TreeNode` invocation). This PR is **reject-only**; it does not yet render recursive templates.
- Corrects the `template-support-matrix.md` "Recursive / circular template references" row (failure mode: parse-time `ParseError`, not a runtime infinite loop).

### Tests
- **`internal/parse/flatten_cycle_test.go`** — the discriminating **diamond** test (proves active-path, not global-visited), direct + mutual + self-referential-entry cycles → `*ParseError` not panic, non-recursive composition unaffected.
- **`recursive_template_test.go`** — the real public `New(...).Parse(...)` path returns an error, not a crash. (A stack overflow can't be `recover()`'d, so a regression kills the test binary — an unambiguous signal.)
- **`FuzzFlattenTemplate`** — seeded with self-referential sources; 370K execs, no overflow.

E2E is n/a (parse-time only, no runtime surface). `/simplify` run against the diff (1 fix applied: allocation-free cycle-path string).

### DIFF
--- a/docs/design/boilerplate-reduction.md
+++ b/docs/design/boilerplate-reduction.md
@@ -352,7 +352,7 @@ example would return to its documented ~10 lines; checklistkit's 5 builders coll
   → **Resolution: document the view-helper pattern (guide + example), keep a regression anchor.
   No core API change.**
 - **C8. Recursive `{{template}}` support (direct consumer: prereview; latent: tinkerdown).**
-  📝 **DESIGN APPROVED (2026-07-15) — Phase 1 next.** livetemplate flattens `{{template}}` calls at
+  🔨 **DESIGN APPROVED (2026-07-15); Phase 1 SHIPPED (2026-07-16).** livetemplate flattens `{{template}}` calls at
   parse time with no cycle guard, so a self-referential template stack-overflows during `Parse`; a
   recursive file tree drops to a standalone `html/template` injected as opaque `template.HTML`
   (prereview `internal/review/filetree.go`), which removes the whole subtree from reactive diffing.
@@ -364,8 +364,10 @@ example would return to its documented ~10 lines; checklistkit's 5 builders coll
   Recommended approach: route recursive invocations through the runtime nested-`TreeNode` path
   `{{range}}` already uses (bounded by data depth, not parse-time expansion). Approach A
   (selective), a max-depth guard erroring uniformly, and content-hash `data-key` fallback are all
-  decided (see the proposal's § Decisions); Phase 1 (crash → `ParseError`) is independently
-  shippable and up next.
+  decided (see the proposal's § Decisions). **Phase 1 (crash → `ParseError`)** landed: an
+  active-path cycle guard in `internal/parse/flatten.go` rejects self-referential templates at
+  parse time with a `ParseError` naming the cycle, instead of stack-overflowing; the support matrix
+  row is corrected. Phases 2–5 (runtime nested-`TreeNode` invocation) remain.
 - **C9. Static-asset passthrough alongside the `/` LiveHandler.** ✅ **RESOLVED (docs, no new
   API).** The original pitch was a framework static-passthrough wrapper. On audit the common
   case — assets under a **known prefix** — is already handled by plain `net/http`:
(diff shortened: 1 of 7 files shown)


adnaan:
Thanks — addressed both points:

- **Slice-aliasing invariant:** added a comment at the `stack = append(...)` site documenting that the backing-array reuse is safe only because the walk is strictly sequential, and that the invariant must hold if the walk is ever parallelized (42e2eaec).
- **Depth vs. cycle:** correct and intentional — Phase 1 rejects *cyclic* invocation graphs, not raw recursion depth. A deep-but-acyclic chain of uniquely-named `{{define}}/{{template}}` invocations is not guarded here; that's the max-depth guard decided for **Phase 4** (`docs/proposals/recursive-templates-proposal.md` § Decisions #2), which will apply on both the render and update paths. Left as a deliberate scope boundary rather than folded in, since the depth guard belongs with the runtime-invocation machinery it protects. The PR title says "recursive" in the common self-referential sense; the body and commit are precise about "cyclic".

adnaan:
Round 2 — thanks. Both non-blocking observations noted:

- **`Phase: "parse"`** — intentional and, as you note, more accurate: this genuinely happens during flattening. The `ParseError` struct doc comment already lists `"parse"` as a valid phase (`// "parse", "eval", "build"`), so this fills in a documented-but-previously-unused case rather than inventing a convention.
- **Sibling `fmt.Errorf` paths (`template %q not defined` / `has no parse tree`) not using `*ParseError`** — agreed these are inconsistent with the new structured error, but they're pre-existing and out of scope for a reject-only cycle guard. Deferring to **Phase 2**, which reworks this exact `TemplateNode` switch (mark-the-cycle-and-emit-a-runtime-boundary instead of error-and-return) — that's the natural place to unify all three onto `*ParseError` without an unrelated diff here.

Two rounds with no new functional issue — treating this as converged. Holding for maintainer signoff before merge.

adnaan:
Closing as **superseded by #495** — no work is lost.

Verified that #494 is fully contained in #495: this branch's tip (`42e2eaec`) is an ancestor of #495's `tierc-c8-runtime-invocation`, so #495 includes both of this PR's commits (`82357eee` Phase 1 guard + `42e2eaec` docs) plus Phases 2–5, the per-leaf diff, the full-document fix, and the fuzz/oracle coverage.

Phase 1's parse-time *rejection* of recursive `{{template}}` is also **superseded within #495** — commit `736aac24` turns that rejection into runtime rendering — so merging this PR on its own would ship "error cleanly on recursion" only to immediately replace it with "recursion works."

All of C8 will land as the single squashable unit in **#495**, which keeps the whole feature revertable in one command.

PR #495, opened 2026-07-16
feat(parse): render recursive {{template}} at runtime (Tier C · C8 Phases 2-5)

adnaan:
**Stacked on #494 (Phase 1).** Base is `tierc-c8-phase1-cycle-guard`; merge #494 first, then this retargets to `main` cleanly.

Turns the Phase 1 parse-time *rejection* of recursive `{{template}}` into full **runtime support** — the last genuine capability gap from the boilerplate-reduction sweep (issue #483, item C8). A recursive `{{template}}` (file trees, comment threads, org charts) now renders through the reactive tree path and gets **minimal, dynamics-only updates** instead of a full-string resend.

### How it works (all inside `internal/parse` — no cross-package threading)
- **Detection** — `detectRecursiveTemplates` runs a reachable-from-self DFS over the `{{template}}` invocation graph to find cycle members. Non-recursive composition still inlines exactly as before.
- **Registry via appended defines** — `walkAndFlatten` emits recursive calls **verbatim** (un-inlined); `FlattenTemplate` appends each flattened body as a `{{define}}` block. The un-wrapped flattened string carries them, so `parse.Parse` reconstructs a `name→AST` registry on `parse.Template` from the associated templates for free.
- **Runtime invocation** — `invokeTemplate` (new `invoke.go`), dispatched from `walkAST`'s `TemplateNode` case, evaluates the pipe, rebinds dot to the pipe value in a fresh scope (Go resets `.`/`$` on a template call), re-walks the registered body, and wraps the result in a single dynamic slot (mirroring `{{if}}`).
- **Backstop** — `checkFlattenCycle` (Phase 1) stays: an under-detected cycle re-enters and errors cleanly instead of overflowing.
- **Depth guard** — `build.Context` gains `InvocationDepth`/`MaxInvocationDepth`, incremented in a per-invocation copy before the check, so a pointer cycle in the data trips the ceiling rather than overflowing `walkAST`. Configurable via `WithMaxTemplateDepth(n)` / `LVT_MAX_TEMPLATE_DEPTH` (default 128).

### Tests
- **Render parity** — exact HTML for nested / single-level / **mutual** recursion.
- **Minimal update** — add-child emits a single granular range op carrying only the new node (key extraction descends through the invocation wrapper); a depth-grows change resends only the new level's statics.
- **Initial HTML** — `Execute` renders recursion natively.
- **Depth guard** — sole protection on the tree-only path; no-crash on the full path; `WithMaxTemplateDepth(3)` rejects a deep tree on update that the default renders.
- Phase-1 flatten tests flipped to assert emit-verbatim + a `checkFlattenCycle` backstop test.

### Known behavior note
The depth guard's error propagates on the **update** path but degrades to the HTML-structure fallback on **first render** (the framework's general AST-build-failure behavior); infinite data trips html/template `Execute`'s own limit first. Documented in the proposal (Open question 2 / Phase 4).

### Follow-up (not in this PR)
Phase 5 tail — benchmarks, chromedp e2e (lvt repo), support-matrix ❌→✅ flip, recipe — and companions #6/#7 (prereview, tinkerdown) land after a release.

### DIFF
--- a/.github/workflows/cross-repo-test.yml
+++ b/.github/workflows/cross-repo-test.yml
@@ -31,6 +31,23 @@ jobs:
         repository: livetemplate/lvt
         path: lvt
 
+    # Mirror of the "Use matching docs branch" step below: a core change that
+    # alters the wire format also changes lvt's exact-format golden files, which
+    # can only be regenerated on a matching lvt branch (lvt's own CI runs against
+    # the published core). If a lvt branch of the same name exists, use it.
+    - name: Use matching LVT branch if available
+      working-directory: lvt
+      env:
+        LVT_REF: ${{ github.head_ref || github.ref_name }}
+      run: |
+        if git ls-remote --exit-code --heads origin "$LVT_REF" >/dev/null 2>&1; then
+          echo "Using livetemplate/lvt branch: $LVT_REF"
+          git fetch origin "$LVT_REF" --depth=1
+          git checkout FETCH_HEAD
+        else
+          echo "No matching livetemplate/lvt branch for $LVT_REF; using default branch"
+        fi
+
     - name: Checkout client library
       uses: actions/checkout@v4
       with:
--- a/config.go
+++ b/config.go
@@ -38,6 +38,14 @@ type EnvConfig struct {
 	// Environment: LVT_DEV_MODE (true/false, 1/0)
 	DevMode bool
 
+	// MaxTemplateDepth caps how deep recursive {{template}} invocations may nest
+	// while building the reactive tree. It stops unbounded recursion (e.g.
+	// self-referential data) from overflowing the stack, surfacing a clear error
+	// instead. Zero uses the built-in default (128); set a higher value only if
+	// your data is legitimately deeper.
+	// Environment: LVT_MAX_TEMPLATE_DEPTH (positive integer)
+	MaxTemplateDepth int
+
 	// WebSocketDisabled disables WebSocket connections (HTTP-only mode).
 	// Environment: LVT_WEBSOCKET_DISABLED (true/false, 1/0)
 	WebSocketDisabled bool
@@ -163,6 +171,15 @@ func LoadEnvConfig() (*EnvConfig, error) {
 		config.DevMode = b
 	}
 
+	// Load MaxTemplateDepth
+	if val := os.Getenv("LVT_MAX_TEMPLATE_DEPTH"); val != "" {
+		n, err := strconv.Atoi(val)
+		if err != nil || n <= 0 {
+			return nil, fmt.Errorf("invalid LVT_MAX_TEMPLATE_DEPTH %q: must be a positive integer", val)
+		}
+		config.MaxTemplateDepth = n
+	}
+
 	// Load WebSocketDisabled
 	if val := os.Getenv("LVT_WEBSOCKET_DISABLED"); val != "" {
 		b, err := parseBool(val)
@@ -284,6 +301,10 @@ func (c *EnvConfig) ToOptions() []Option {
 
 	if c.DevMode {
 		opts = append(opts, WithDevMode(true))
+	}
+
+	if c.MaxTemplateDepth > 0 {
+		opts = append(opts, WithMaxTemplateDepth(c.MaxTemplateDepth))
 	}
 
 	if c.WebSocketDisabled {
--- a/docs/design/boilerplate-reduction.md
+++ b/docs/design/boilerplate-reduction.md
@@ -352,7 +352,7 @@ example would return to its documented ~10 lines; checklistkit's 5 builders coll
   → **Resolution: document the view-helper pattern (guide + example), keep a regression anchor.
   No core API change.**
 - **C8. Recursive `{{template}}` support (direct consumer: prereview; latent: tinkerdown).**
-  📝 **DESIGN APPROVED (2026-07-15) — Phase 1 next.** livetemplate flattens `{{template}}` calls at
+  🔨 **DESIGN APPROVED (2026-07-15); Phase 1 SHIPPED (2026-07-16).** livetemplate flattens `{{template}}` calls at
   parse time with no cycle guard, so a self-referential template stack-overflows during `Parse`; a
   recursive file tree drops to a standalone `html/template` injected as opaque `template.HTML`
   (prereview `internal/review/filetree.go`), which removes the whole subtree from reactive diffing.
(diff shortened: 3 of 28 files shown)


adnaan:
### Benchmark finding — the value story, measured (corrects the "minimal updates" premise)

`recursive_template_bench_test.go` (depth-5 branch-3 tree, `-benchtime=3x`):

| Benchmark | ns/op | payload | allocs/op |
|---|---|---|---|
| `RecursiveRender` (first render) | ~31 ms | — | ~90.7K |
| `RecursiveUpdate` (deep-leaf edit) | ~34 ms | **24,024 update_bytes** | ~99K |
| `OpaqueHTMLBaseline` (whole tree) | ~1.3 ms | **23,559 full_html_bytes** | ~5.2K |

**What this shows:** for a deep edit in a narrow tree, the reactive update (24 KB) is *essentially equal in bytes* to re-sending the entire opaque `template.HTML` string it replaces (23.5 KB), and costs **~25× the CPU**. The cause is the keying caveat already in the proposal: recursive range items key by a **deep content hash** (the item's real `<li data-key>` is hidden inside the invocation wrapper), so any edit re-sends its whole **enclosing top-level branch** via `["r", oldkey],["p", [branch]]` — not a per-leaf `["u"]`. The update-size win only materializes for **wide** trees (branch is a small fraction) or once the **data-key-through-the-wrapper follow-up** lands.

**So the win C8 ships is _capability_ + _reactive DOM-state preservation_, not payload size** — recursion renders at all (vs a parse-time stack overflow) and updates apply in place (focus/scroll preserved) rather than an `innerHTML` blob replace. The TL;DR, §6, and Phase 5 tracker in the proposal now state this honestly rather than the original "dramatically smaller" claim.

This does not block the *capability* landing, but it does mean the headline "minimal updates for recursion" is a follow-up, not a shipped property.

adnaan:
### Update — the payload win is now DELIVERED (per-leaf recursive updates)

Following the benchmark finding above (deep edits ≈ full re-send), the maintainer chose to build the real fix rather than ship the caveat. Done in `8d1c3438`:

| Metric | Before | After |
|---|---|---|
| Deep-leaf update (depth-5 branch-3) | ~24,024 B | **~200 B** (~100× smaller than the 23.5 KB opaque re-send) |
| Op shape | `["r",oldkey],["p",[branch+statics]]` | nested `["u", key, …]` chain to the single changed leaf, statics-free |

**Three coordinated changes** (see commit body): key recursive items by their data-key *through* the invocation wrapper (stable keys) · keep nested-range-containing ranges off the stream-mode transition (retains old item trees) · teach the differential range diff to emit `["u", key, <recursive diff>]` for kept-changed items.

**⚠️ Framework-wide, not just recursion:** change 3 means *every* nested data-keyed range now gets per-item `["u"]` where it previously took a full-tree fallback. Backed by the nested-range fuzz + TS-oracle suites and a new nested-range kept-item unit test. Client-side is unchanged (`deepMergeTreeNodes` already applies nested-ops-in-payload) — server-only.

**Also fixes a pre-existing C8 bug the browser e2e surfaced:** full HTML-document recursive templates were silently falling back to HTML-string diffing (the flatten-appended `{{define}}` blocks were dropped by body extraction), so the reactive tree never ran for full-doc apps. `ExtractTemplateBodyContent` now re-attaches them; pinned by `TestRecursiveTemplate_FullDocument_Reactive`.

**Verified end-to-end** by the lvt browser e2e (`lvt` `3a2a092`): a deep rename morphs the `<li>` in place (marker survives), the unaffected sibling's DOM survives, the WS frame is scoped to the changed leaf. Full Go suite green.

`/simplify` applied (shared `itemChange` helper, folded traversal, strengthened the fuzz statics invariant, +nested-range unit test). Two altitude follow-ups noted as out-of-scope: a structured flatten→parse hand-off for the trailing-define re-attach, and a semantic wrapper tag so `wrappedItemKey` doesn't structurally guess the invocation wrapper.

adnaan:
### Framework-wide change — validated against real apps + hardened in-repo coverage

The per-item `["u"]` change affects *every* nested data-keyed range, so it was validated two ways.

**In-repo coverage was insufficient and is now fixed (`5ab8249a`).** The nested-range fuzz (`TestFuzzNestedRanges_*`) only ran **Go-side invariants** (minimality, key-stability) — it verified the diff is minimal/valid but never that the **real client applies** the nested `["u"]` chains. Client-application coverage existed only for flat async/modal templates. Added three TS-oracle tests (round-trip the diff through the real TypeScript client, compare applied HTML):
- `TestFuzzNestedRanges_TSOracle` — stable `id=` keys.
- `TestFuzzNestedRanges_UpdateHeavy_TSOracle` — biased to kept-item content edits (the `["u"]` path).
- `TestFuzzNestedRangesContentHash_TSOracle` — `data-item` (unrecognized attr → **content-hash** keys, the shape checklistkit/devbox/prereview use). Content-hash keys are legitimately unstable, so it skips the KeyStability invariant; HTML-comparison + minimality still run. **Zero HTML divergence, 100 rapid cases each.**

**Real-app smoke tests (built against this branch via temp go.work):**

| App | Nested-range shape | Result |
|---|---|---|
| **prereview** | stable `data-key`, **3-deep** (VisibleLines→CommentsByEndLine→thread), heterogeneous fold/line rows | 10 browser e2e **PASS** |
| **checklistkit** | **content-hash** `data-item`, `{{template}}` invocation, heterogeneous status branches | 4 browser e2e **PASS** |
| **devbox-dash** | triple-nest IssueGroups→Issues→Labels (mixed id/content-hash) | 9 e2e **PASS** |
| **tinkerdown** | sidebar nav is **Go-side recursion** (not a livetemplate range) → unaffected | non-browser suite green |
| **docs** | not in go.work (pins published lib); recipes simpler than above | n/a (release-gated) |

Corners covered: both keying modes (stable + content-hash), fragment + full-doc, `{{template}}` invocation, heterogeneous items, 3-deep nesting — validated both through the real client (fuzz oracle) and 3 real apps.

adnaan:
Thanks — all four "worth double-checking" items are non-blocking and are deliberate/already-tracked; confirming rather than pushing another round:

1. **Keying generalizes beyond recursion (`allWrappedItemKeys`/`wrappedItemKey`).** Intentional (Decision 4): a range item with no own `data-key` but a single nested keyed child now keys by that child's key. This is exactly the fragility filed as **#497** — `createConditionalWrapper` builds both the `{{template}}`-invocation wrapper and `{{if}}`/`{{with}}` wrappers, so a structural predicate can't distinguish them; the right fix is a semantic wrapper tag, tracked there. Blast radius is bounded by the all-or-nothing `allWrappedItemKeys` guard (a range never mixes stable keys with content hashes), and keying a single-`{{if}}`-wrapped keyed child by its `data-key` is an improvement (stable keys) rather than a regression. Good call that the explicit non-recursive single-`{{if}}`-wrapped case isn't fuzzed — noting it on #497.

2. **A single structural flip falls back the whole range.** Intended tradeoff. `generateKeptItemUpdateOps` bails to a full-range replace on the first `structureChanged` item because a `["u"]` payload must be dynamics-only (it can't carry the statics a branch flip needs). Emitting per-item `["u"]` for the unaffected items *plus* a full re-send for the one structural item is a possible refinement, but the all-or-nothing fallback is simpler and always correct; structural flips are far rarer than content edits (the case the PR optimizes). Fine to revisit if profiling ever shows it matters.

3. **Depth-guard first-render/update asymmetry — already diagnosable.** The first-render fallback isn't silent: `generateInitialTreeWithoutRegistry` logs `slog.Warn("Template parsing failed, falling back to HTML structure-based tree", …, error=<err>)`, and for an over-depth tree that `err` is precisely `"recursive template %q exceeded the maximum invocation depth of %d …"`. So the recursion-depth case is distinguishable from other AST failures in logs today; a dedicated metric is a reasonable future add but not needed for diagnosability.

4. **`trailingDefineBlocks` picks up any `{{define}}` after `</body>`.** As you note, arguably correct — a trailing `{{define}}` is a template definition and keeping it is right, not a leak. The indirection (why defines live after `</html>` at all) is the structured flatten→parse hand-off tracked in **#496**, which would remove the heuristic entirely.

Also added a direct table-driven test for `trailingDefineBlocks` in this round per your earlier note (`TestExtractTemplateBodyContent_TrailingDefines`).

PR #499, opened 2026-07-18
fix(release,ci): survive an aborted release, and stop interpolating the tag into shell

adnaan:
Both problems surfaced while releasing v0.19.1.

## 1. An aborted release left a half-bumped tree

The v0.19.1 release aborted at `build_and_test` on a test failure that did not reproduce (the same suite passed three times afterwards). The abort itself was fine. The state it left behind was not:

```
 M CHANGELOG.md
 M VERSION
```

Because the order was `update_versions` → `generate_changelog` → `build_and_test`, both files were already rewritten when the tests ran. And `main()` refuses to start on a dirty tree — so the retry was blocked until someone worked out what to revert. That is a confusing trap at exactly the wrong moment.

**`build_and_test` now runs first.** No Go code reads the `VERSION` file, so the bump cannot change the test result, and a failure leaves the tree untouched.

**A trap covers the remaining window** — a failed changelog step, a rejected commit, a tag clash. It restores from `HEAD`, not the index:

> `commit_and_tag` stages both files before committing, so on a failed commit they are already staged. A plain `git checkout --` restores from the index — copying the modified versions back over themselves, achieving nothing.

The first version of this fix had precisely that bug. The abort test below is what caught it, which is the argument for having written the test rather than eyeballing the diff.

## 2. `${{ github.ref_name }}` interpolated into a `run:` block

```yaml
-F "client_payload[ref]=${{ github.ref_name }}"
```

A tag name is chosen by whoever pushes the tag, and this job holds a PAT with write access to `livetemplate/docs`. Values now arrive via `env:` and are quoted; **no interpolation remains in the `run:` block**. The ref is validated against `[A-Za-z0-9._/-]+` before the token is spent, mirroring what the docs sync gained in livetemplate/docs#115.

Exploiting it needs tag-push access, so this is hardening rather than an open door.

## Verification

Both paths exercised in a throwaway clone with `publish_github` stubbed out, so nothing escaped:

| Scenario | Result |
|---|---|
| Forced failure after the bump | `VERSION` restored, **0 dirty files**, retry unblocked |
| Successful run | release commit + tag intact, trap correctly stood down |

Also confirmed the reorder in the log output — `Running Go tests…` now precedes `Updating VERSION file`.

## Note on scope

An audit of every `run:` block across the repo's workflows found one other interpolation, `benchmark.yml` line 148. It is **not** a vulnerability — `message` is set to hardcoded literals a few lines above — so I left it alone rather than churn it under a security banner.

lvt has the identical `release.sh` ordering bug; fixed separately in livetemplate/lvt.

### DIFF
--- a/.github/workflows/dispatch-docs-sync.yml
+++ b/.github/workflows/dispatch-docs-sync.yml
@@ -19,13 +19,28 @@ jobs:
           # Same PAT used by all four source repos. Setup once via:
           #   gh secret set DOCS_DISPATCH_TOKEN --repo livetemplate/<this> --body <token>
           GH_TOKEN: ${{ secrets.DOCS_DISPATCH_TOKEN }}
+          # A tag name is chosen by whoever pushes the tag, so it does not go
+          # into the script body via ${{ }} — that would be shell injection into
+          # a job holding a cross-repo write token. It arrives as an environment
+          # variable and is quoted at every use. github.repository is set by the
+          # runner rather than by user input, but travels the same path so no
+          # interpolation remains in this run: block at all.
+          SOURCE_REPO: https://github.com/${{ github.repository }}
+          SOURCE_REF: ${{ github.ref_name }}
         run: |
+          set -euo pipefail
           if [ -z "$GH_TOKEN" ]; then
             echo "::warning::DOCS_DISPATCH_TOKEN not set; skipping docs sync dispatch"
             exit 0
           fi
+          # The docs sync rejects refs outside this shape too; failing here means
+          # a bad tag never spends the token.
+          if ! printf '%s' "$SOURCE_REF" | grep -Eq '^[A-Za-z0-9._/-]+$'; then
+            echo "::error::refusing to dispatch an unsafe ref: $SOURCE_REF"
+            exit 1
+          fi
           gh api -X POST /repos/livetemplate/docs/dispatches \
             -F event_type=sync-source \
-            -F "client_payload[source_repo]=https://github.com/${{ github.repository }}" \
-            -F "client_payload[ref]=${{ github.ref_name }}"
-          echo "::notice::dispatched docs sync for ${{ github.ref_name }}"
+            -F "client_payload[source_repo]=$SOURCE_REPO" \
+            -F "client_payload[ref]=$SOURCE_REF"
+          echo "::notice::dispatched docs sync for $SOURCE_REF"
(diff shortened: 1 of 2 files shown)


adnaan:
Mirrors the review outcome on livetemplate/lvt#339, which reviewed the same function first.

Two fixes came out of that round and are applied here too:

1. **Per-path restore** — `git checkout HEAD -- VERSION CHANGELOG.md` resolves the pathspec before touching the worktree and aborts entirely if any path is missing from HEAD, so a single call restores *neither* file. Verified directly: with CHANGELOG.md untracked, the combined form left VERSION at its bumped value.
2. **Remove a CHANGELOG.md the aborted run created** — otherwise the "retry unblocked" guarantee holds only for the common case, since `git status --porcelain` lists untracked and staged-new files alike and `main()` refuses to start on either. Safe by construction: main() already refuses a dirty tree and --porcelain reports untracked files, so a CHANGELOG.md absent from HEAD at trap time cannot have pre-existed.

Note the earlier red `Test LVT against Core Changes` on commit ee6ae90 was the known auto-discovery flake (`readdirent .../internal/upload/.uploads/` — auto-discovery racing a sibling upload test's teardown), not this change: the diff contains no Go code, `Unit Tests` passed on the same commit, and only the -race job failed.

adnaan:
Converged — this round raised no new functional issue, and independently re-derived the reasoning rather than restating it. Stopping pushes here per the project's review-loop guidance; answering both follow-ups by reply.

**1. No automated test for the trap.** Agreed, and your framing is right — three iterations to get this correct is the argument for it, not against. I'm not adding a bats harness inside a fix PR though: there's no existing shell-test scaffolding in the repo, so it's a new dependency and a new convention decided under a bugfix banner. Worth doing deliberately if this script grows more failure handling.

For the record, the manual abort test did earn its keep — it caught all three bugs, every one of which survived reading the diff:

1. restore-from-index instead of HEAD — the trap logged success and restored nothing
2. combined pathspec — one untracked file meant *neither* got restored
3. created `CHANGELOG.md` left staged — `rm` alone leaves an `A ` entry, so "retry unblocked" held only for the common case

**2. Post-commit / pre-push abort.** Good catch, and correctly scoped as pre-existing — filed as #500 rather than folded in here, since fixing it means adding a resume path to `main()`, which is a different change from "don't leave a half-bumped tree."

Worth noting the interaction with this PR: the trap deliberately stands down once `release_committed=true`, which is right (restoring would discard the release commit) but is also exactly what leaves that state unflagged. So #499 doesn't cause #500, but it does make it the *only* remaining way to end up in a confusing post-abort state. lvt shares the structure and presumably the gap.

PR #503, opened 2026-07-19
refactor(parse): hand recursion defines back from FlattenTemplate (closes #496)

### LINKED ISSUE #496: Recursion registry: structured FlattenTemplate→parse hand-off instead of trailing-{{define}} string-rescan
Follow-up from the C8 per-leaf work (PR #495, altitude review).

**Current state.** `FlattenTemplate` appends the recursion cycle members as `{{define}}` blocks *after* `</html>`. The serving tree path extracts the body via `ExtractTemplateBodyContent`, which (as of #495) string-rescans for `{{define` past the body close and re-attaches those blocks so the recursion registry survives. Before #495 they were dropped, silently degrading full-document recursive templates to HTML-string diffing.

**Why this is the wrong altitude.** `ExtractTemplateBodyContent` (a pure HTML body slicer) now encodes a private invariant of `FlattenTemplate`'s output format via a bare `strings.Index("{{define")`. It assumes *nothing but `{{define}}` blocks* live after `</body>` — if `FlattenTemplate` ever appends anything else after them (a trailing script, a diagnostic comment), `afterBody[idx:]` sweeps it into the template. Any change to the flatten tail layout silently degrades recursion with no error surfaced.

**Proposed fix.** A structured hand-off: `FlattenTemplate` returns `(document, defines)` (or the serving path threads the recursion defines explicitly to the parser), so body extraction stays a pure `<body>…</body>` slice with zero knowledge of recursion. Failing that, append the defines *just before* `</body>` (inside the body `FlattenTemplate` already owns) so extraction picks them up with no special-casing.

Pinned by `TestRecursiveTemplate_FullDocument_Reactive`; not a blocker (renders correct), a robustness/altitude improvement.

adnaan:
Closes #496.

`FlattenTemplate` emits `{{define}}` blocks for templates on an invocation cycle, which cannot be inlined. They were concatenated onto the flattened document, and `ExtractTemplateBodyContent` — a pure HTML body slicer — rescanned them back out:

```go
idx := strings.Index(afterBody, "{{define")
```

That made an HTML slicer encode `FlattenTemplate`'s private output layout. Appending anything after the defines — a trailing script, a diagnostic comment — would sweep it into the template and silently degrade recursion to HTML-string diffing, with no error anywhere.

## The change

**`FlattenTemplate` returns `(document, defines)`.** Callers assemble what they need: `document+defines` for `html/template`, body slice + defines for the reactive parse. The rescan is gone.

**Body extraction reports whether it sliced.** `ExtractTemplateBodyContentSliced` returns the body plus whether a `<body>` region was actually cut out — which is exactly when the tail gets dropped. The caller re-attaches only then.

That last part isn't defensive coding. `templateStr` still ends with the defines, so the paths returning it unchanged (a fragment has no `<body>`) would carry every block twice. **The first version of this change appended unconditionally and doubled every define on fragments** — caught by the byte-identity check below, not by any existing test.

## Byte identity as the acceptance bar

The issue notes this renders correctly today, so the refactor is only a win if it's *provably* behaviour-preserving. A temporary harness captured the two assembled strings — what `html/template` parses, and what the reactive parse consumes — across six shapes on `main` and on this branch:

| Shape | Result |
|---|---|
| full-document recursive | identical |
| fragment recursive | identical |
| mutual recursion | identical |
| non-recursive composition | identical |
| full-document, no composition | identical |
| fragment, no composition | identical |

Byte-for-byte, all six. The harness was deleted before commit.

## Why a field, not eager computation

An earlier version computed body content eagerly in `parseInternal` to avoid adding a field. That turned a lazily-computed, invalidatable cache into one that goes stale if `templateStr` is reassigned — which two existing tests do directly, and both broke. Keeping it lazy preserves the invalidation semantics exactly, and both tests pass **unchanged**.

`recursionDefines` is therefore a `Template` field. `Clone` copies fields by explicit enumeration rather than struct assignment, so a dropped field would silently degrade recursion — and since production renders through per-session clones, a master-only assertion would never notice. `TestRecursiveTemplate_CloneCarriesDefines` guards it, **verified to fail** when that `Clone` line is removed:

```
--- FAIL: TestRecursiveTemplate_CloneCarriesDefines
    Clone dropped recursionDefines:
```

## Coverage moved, not deleted

- `internal/build` now pins the pure-slicer contract and the `sliced` flag, including that content after `</html>` is dropped rather than rescued
- `TestRecursiveTemplate_BodyContentCarriesDefines` pins re-attachment a layer up, asserting the define appears **exactly once**
- The flatten cycle tests now assert registry blocks land in the `defines` return **and do not leak into the document** — a stronger claim than the previous substring check against the concatenation, which would still pass if the two were re-merged

Full suite green.

### DIFF
--- a/docs/proposals/recursive-templates-proposal.md
+++ b/docs/proposals/recursive-templates-proposal.md
@@ -436,10 +436,15 @@ Three categories, each a hard gate — no phase merges with a category left as "
         them for full HTML documents — leaving the recursion registry empty at serve time, so the whole
         template silently degraded to HTML-string diffing (none of the reactive tree / per-leaf path
         ran). `ExtractTemplateBodyContent` now re-attaches trailing `{{define}}` blocks; pinned by
-        `TestRecursiveTemplate_FullDocument_Reactive`.
-      - [ ] **Release-gated docs:** flip the `template-support-matrix.md` "Recursive / circular template
-        references" row from ❌ to ✅, update `current-limitations.md`, add a recipe. Matrix reflects
-        *released* behavior, so this lands with the release, not before.
+        `TestRecursiveTemplate_FullDocument_Reactive`. **Superseded by #496:** that re-attachment was a
+        `strings.Index("{{define")` rescan inside a pure HTML slicer, so `FlattenTemplate` now returns
+        `(document, defines)` and the caller re-attaches. The guarantee is unchanged; only where it
+        lives moved.
+      - [x] **Release-gated docs:** landed with v0.19.0/v0.19.1 — `template-support-matrix.md`'s
+        "Recursive / circular template references" row is ✅ with a **Recursion depth** section,
+        `CONFIGURATION.md` documents `LVT_MAX_TEMPLATE_DEPTH` (#498), and the `file-tree` recipe ships
+        in the docs site (livetemplate/docs#114). `current-limitations.md` needed no change — it never
+        claimed recursion was unsupported.
 
 ### Companion migrations (dependent repos — land per the lockstep convention once a release ships)
 
--- a/internal/build/wrapper.go
+++ b/internal/build/wrapper.go
@@ -193,40 +193,46 @@ func injectWrapperDivStringBased(htmlDoc string, wrapperID string, loadingDisabl
 // <body class="dark">). Uses html.Tokenizer (see locateBodyAndFirstScript)
 // to avoid being fooled by '<body' substrings in head content.
 //
-// It also re-attaches any trailing {{define}} blocks that FlattenTemplate appends
-// AFTER the document (past </body></html>) for recursive {{template}} support: the
-// body's {{template}} calls reference those defines, and the extracted content is
-// parsed on its own for reactive tree generation, so dropping them would leave the
-// recursion registry empty and silently degrade recursive templates to HTML-string
-// diffing.
+// This is purely a slice between the body tags and knows nothing about what
+// produced the template. Recursive {{template}} support needs FlattenTemplate's
+// {{define}} blocks, which sit after </html> and so fall outside this slice —
+// the caller appends them, having received them from FlattenTemplate directly.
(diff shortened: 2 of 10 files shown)


adnaan:
Both applied, and one of them was drift I introduced rather than a nit.

**1. `cachedBodyContent` field comment.** Correct — it still claimed to hold `ExtractTemplateBodyContent(t.templateStr)`, which this PR made false. Now describes what it actually holds: the `<body>` slice plus `recursionDefines` when a body region was sliced. Worth noting the review's own framing is why this matters — `Clone`'s comment two fields down was accurate, so the two sat side by side disagreeing.

**2. The proposal's superseded claim.** You called it a historical progress-log entry and suggested leaving it, which I mostly agree with — rewriting it would falsify what that phase actually did. But it names a mechanism that no longer exists, so someone grepping for how re-attachment works lands on a dead end. Split the difference: the entry keeps its original text and gains a **Superseded by #496** note. History preserved, misdirection removed.

**Also fixed, not raised in the review:** the very next checkbox — *"Release-gated docs: flip the matrix row from ❌ to ✅, update `current-limitations.md`, add a recipe"* — was still unticked, but that work landed today. The matrix row is ✅ with a Recursion depth section and `CONFIGURATION.md` documents `LVT_MAX_TEMPLATE_DEPTH` (#498), and the `file-tree` recipe shipped in livetemplate/docs#114. I verified `current-limitations.md` genuinely needed no change — it never claimed recursion was unsupported, so that clause was satisfied vacuously rather than skipped.

That one is worth flagging because it's the same failure mode as the comment: a stale record that reads as current. An unticked box for completed work is how a follow-up gets done twice, or gets treated as outstanding long after it isn't.

Thanks for tracing all three paths by hand given the sandbox — the fragment case is exactly where the first version of this change was wrong, so having it independently checked is worth more than usual here.

PR #504, opened 2026-07-19
refactor(parse): tag wrappers instead of inferring them from shape (closes #497)

### LINKED ISSUE #497: Tag the invocation wrapper semantically so wrappedItemKey doesn't structurally guess it
Follow-up from the C8 per-leaf work (PR #495, altitude review).

**Current state.** Recursive `{{template}}` range items are keyed by their real `data-key` read *through* the invocation wrapper. `wrappedItemKey` (`internal/parse/keys.go`) detects "the invocation wrapper" structurally: empty statics + exactly one nested `*TreeNode` child.

**Why this is fragile.** The invocation wrapper and the conditional wrapper are produced by the *same constructor* — `createConditionalWrapper` (conditional.go:40), called from both `conditional.go` ({{if}}/{{with}}) and `invoke.go` ({{template}} invocation). Both set `Statics = defaultFieldStatics` (`["",""]`) and `SetDynamic(0, child)`. So "empty statics + single nested child" matches a `{{if}}`/`{{with}}` branch item exactly as well as a `{{template}}` invocation — no structural predicate can separate them. A range whose items happen to be single-child conditional/with wrappers exposing a keyed child would silently inherit invocation-wrapper (data-key) keying.

**Mitigation in place.** `allWrappedItemKeys` is all-or-nothing (a range never mixes stable data-keys with content hashes), which limits blast radius, and reading the child's real `data-key` is often the right key anyway. The fuzz + TS-oracle + recursive e2e suites all pass, so no observed breakage.

**Proposed fix.** Tag the node semantically at creation: give `invokeTemplate` its own wrapper constructor or a marker field (e.g. a wrapper-kind enum on `TreeNode`), then have `wrappedItemKey` key off that marker rather than reverse-engineering it from child count.

Not a blocker; a correctness-robustness/altitude improvement.

adnaan:
Closes #497.

`wrappedItemKey` identified a parser-created wrapper **structurally** — empty statics plus exactly one nested `TreeNode` child. That shape describes a conditional wrapper, an invocation wrapper, and a plain field node holding a tree alike, so no predicate over the finished node could recover what produced it.

`createWrapper` (renamed from `createConditionalWrapper`, whose name `invoke.go` already apologised for) now records a `WrapperKind`, and `wrappedItemKey` asks for the tag. All wrappers are born at that one constructor, so tagging there is enough — field nodes and `createEmptyConditionalWrapper` stay untagged automatically.

## Diverging from the issue's literal proposal

The issue says *"tag the invocation wrapper."* Doing exactly that would be a **regression**, and I measured it rather than reasoned it. On `main`:

| Shape | keying today |
|---|---|
| recursive `{{template}}` range | data-key |
| **`{{if}}`-wrapped keyed range** | **data-key** |
| unkeyed ranges | content hash |

`{{if}}`-wrapped keyed ranges key by their child's `data-key` today. Requiring `WrapperInvocation` drops them to content hashing — worse keying, more DOM churn. The child's real `data-key` is the right identity for a conditional-wrapped item too; the issue itself concedes as much ("often the right key anyway"). So **any** wrapper kind qualifies, and the kinds are recorded separately so the distinction exists if something ever needs it.

## The step the issue didn't anticipate

This is the part that matters most. **`walkList` does not keep a wrapper as a node** — it merges the wrapper's statics and dynamics into a fresh tree:

```go
tree := NewTreeNode()
...
statics[len(statics)-1] += childStatics[0]   // wrapper dissolved
```

So what reaches `wrappedItemKey` has the wrapper's *shape* without being the wrapper. The kind is now carried across that merge when exactly one child contributed.

Without it, the tag is never observed and **every** through-wrapper keying regresses to content hashing — far wider than the coincidence the tag prevents. Not theoretical: the first version of this change lacked propagation and both recursive and `{{if}}`-wrapped ranges silently fell back to hashes.

## This is latent robustness, not a live-bug fix

Stated plainly because the issue says "no observed breakage" and I could not manufacture any.

A temporary harness captured the keying decision across **ten** shapes on `main` and on this branch — recursive `{{template}}`, `{{if}}`/`{{with}}`-wrapped keyed, plain keyed, unkeyed, and four constructed candidate false positives. Output is **identical**. Removing the marker check changes none of them either.

That's consistent with the code: `createWrapper` is the only thing that nests a `TreeNode` under empty statics, and field nodes hold **strings** (`createSingleDynamicTree`), so no template reachable today produces the untagged shape. The guard is against a future construct acquiring through-wrapper keying by coincidence — which is why the unit test constructs that shape by hand.

## Wire and fingerprint are untouched

`MarshalJSON` and `ToMap` name the fields they emit; the fingerprint hashes statics and dynamic shape. The marker reaches neither. Two nodes differing only in kind render identically and share a fingerprint, which is correct.

## Both guards negative-gated

Verified to fail when what they protect is removed:

```
# marker check removed
--- FAIL: TestWrappedItemKey_RequiresTheMarker

# walkList propagation removed
--- FAIL: TestWalkList_CarriesWrapperAcrossTheMerge
    expected the item keyed by its child's data-key, got "466b3955bebc"
```

Full suite green.

### DIFF
--- a/internal/build/types.go
+++ b/internal/build/types.go
@@ -29,12 +29,6 @@ func PositionKey(index int) string {
 	return strconv.Itoa(index)
 }
 
-// TreeNode represents a node in the template tree structure with type safety.
-// It replaces the old map[string]interface{} representation while maintaining
-// wire format compatibility through custom JSON marshaling.
-//
-// TreeNode should not be value-copied after first use of GetStructureFingerprint,
-// as it contains an atomic.Value for fingerprint caching. Always use *TreeNode.
 // WrapperKind identifies a node the parser created purely to give a child tree
 // its own dynamic slot. The kinds are structurally identical — telling them
 // apart, or telling them from an ordinary field node, is only possible because
@@ -61,6 +55,12 @@ const (
 // because we put it there" should use this rather than testing a specific kind.
 func (k WrapperKind) IsWrapper() bool { return k != WrapperNone }
 
+// TreeNode represents a node in the template tree structure with type safety.
+// It replaces the old map[string]interface{} representation while maintaining
+// wire format compatibility through custom JSON marshaling.
+//
+// TreeNode should not be value-copied after first use of GetStructureFingerprint,
+// as it contains an atomic.Value for fingerprint caching. Always use *TreeNode.
 type TreeNode struct {
 	// Statics are the static HTML parts of the template (key: "s")
 	Statics []string
@@ -736,6 +736,11 @@ func (tn *TreeNode) Clone() *TreeNode {
 		Fingerprint:  tn.Fingerprint,
 		AutoKey:      tn.AutoKey,
 		dynamicCount: tn.dynamicCount,
+		// Carried deliberately: Clone builds from named fields, so a new field
+		// is dropped unless listed. Losing the kind would make a cloned wrapper
+		// indistinguishable from an ordinary node again — the precise failure
+		// this tag exists to prevent (issue #497).
+		Wrapper: tn.Wrapper,
 	}
 
 	// Clone statics
--- a/internal/build/types.go
+++ b/internal/build/types.go
@@ -35,6 +35,32 @@ func PositionKey(index int) string {
 //
 // TreeNode should not be value-copied after first use of GetStructureFingerprint,
 // as it contains an atomic.Value for fingerprint caching. Always use *TreeNode.
+// WrapperKind identifies a node the parser created purely to give a child tree
+// its own dynamic slot. The kinds are structurally identical — telling them
+// apart, or telling them from an ordinary field node, is only possible because
+// the constructor records which one it made.
+type WrapperKind uint8
+
+const (
+	// WrapperNone is the zero value: not a parser-created wrapper. Every node
+	// built anywhere else keeps it, so the marker cannot be acquired by accident.
+	WrapperNone WrapperKind = iota
+
+	// WrapperConditional wraps an {{if}}/{{with}} branch to keep the branch
+	// structure stable for diffing.
+	WrapperConditional
+
+	// WrapperInvocation wraps a {{template}} invocation body. A recursive
+	// subtree's depth varies with the data, so it occupies one slot to keep
+	// those changes from restructuring the parent's statics.
+	WrapperInvocation
+)
+
+// IsWrapper reports whether the parser created this node as a wrapper, of any
+// kind. Consumers that only care "is there a child tree hidden one level down
+// because we put it there" should use this rather than testing a specific kind.
+func (k WrapperKind) IsWrapper() bool { return k != WrapperNone }
+
 type TreeNode struct {
 	// Statics are the static HTML parts of the template (key: "s")
 	Statics []string
@@ -56,6 +82,19 @@ type TreeNode struct {
 
 	// Metadata contains additional information like ID key mappings (key: "m")
 	Metadata *TreeMetadata
+
+	// Wrapper records that the parser created this node as a single-dynamic-slot
+	// wrapper around a child tree, and which construct asked for it. Wrappers are
+	// structurally indistinguishable from each other and from a plain field node
+	// — all carry `["", ""]` statics and one nested child — so consumers that
+	// need to know "the parser wrapped this" must be told, not left to infer it
+	// from shape (issue #497).
+	//
+	// Internal only: MarshalJSON and ToMap build the wire map by naming the
+	// fields they emit, and the structure fingerprint hashes statics and dynamic
+	// shape, so this reaches neither. Two nodes differing only in Wrapper render
+	// identically and share a fingerprint, which is correct.
+	Wrapper WrapperKind
 
 	// dynamicCount tracks the number of non-nil entries in Dynamics for fast HasDynamics.
 	dynamicCount int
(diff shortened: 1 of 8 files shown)


adnaan:
This was a real regression in my own first revision, and the diagnosis is exactly right. Fixed.

**Reproduced before touching anything**, and confirmed it was mine:

| shape | main | first revision |
|---|---|---|
| minified | data-key | data-key |
| indented | data-key | **content hash** |
| container-wrapped | data-key | **content hash** |

A `TextNode` yields one static even when it's only the whitespace around an indented construct, so `contributingChildren` counted formatting. The minified body has one child; the indented one has three. Only the first propagated the tag.

That is precisely the regression this PR's description claims to avoid — arriving through template *formatting* rather than through wrapper kind, which is why my reasoning didn't reach it.

**The fix is your suggested direction**: count children carrying a wrapper tag, not children contributing statics. Your observation about why the old code was immune is the part that convinced me — it counted `*TreeNode` entries in `Dynamics`, and text nodes contribute none, so neighbouring text never mattered. My propagation introduced a dependency on sibling *statics* that never existed.

**The blind spot was in the verification, not just the code.** Every shape in the "ten shapes measured identical" matrix was minified — none of them *could* expose it. That claim was true and useless. The matrix is now twelve shapes including indented, container-wrapped and text-sibling variants, identical between this branch and main across all of them.

`TestWalkList_WrapperSurvivesSurroundingText` pins it, verified to fail on the old logic for exactly the indented, container-wrapped and text-sibling cases:

```
--- FAIL: .../indented           keyed by "466b3955bebc", want "/a.go"
--- FAIL: .../container_wrapped  keyed by "acd33ba237ae", want "/a.go"
--- FAIL: .../text_sibling       keyed by "466b3955bebc", want "/a.go"
```

**Also applied: the `{{with}}` correction.** You're right that `handleWith` delegates to `walkAST` and never calls `createWrapper` — my matrix agrees, `{{with}}`-wrapped keyed ranges don't get through-wrapper keying. Both comments claiming otherwise are fixed.

Worth saying plainly: the Docs Examples check went red on this PR earlier and I traced it to a flaky test of mine (livetemplate/docs#119, now merged). That was correct as far as it went — the wire output was byte-identical for that sequence — but it also meant a real regression sat in the diff while I was looking at browser timing. Your catch is what surfaced it.

adnaan:
Both real, both fixed. The first is worth dwelling on.

**`TreeNode.Clone()` dropping `Wrapper`.** Correct, and it's the same failure this PR exists to prevent — a cloned wrapper reverts to `WrapperNone`, becomes indistinguishable from an ordinary node, and through-wrapper keying falls back to content hashing. Latent as you say (nothing in `internal/parse` clones a tagged node before `wrappedItemKey` reads it), but `Clone` is exported and documented as a deep copy.

What makes this worth more than a one-line fix: **it's the third named-field copy today to silently drop a newly added field.**

1. `Template.Clone` in #503 — caught because my own notes warned about it
2. `walkList`'s merge earlier in this branch — the field was set and then dissolved
3. `TreeNode.Clone` here — caught by you

Three different mechanisms, same shape: a copy built by naming fields, so a new field is dropped by default and nothing at the call site or at runtime says so. I checked the wire format and the fingerprint for this field and reasoned carefully about `walkList`, and still missed the one remaining copy in the same file.

`TestTreeNode_Clone_CarriesWrapper` pins it across all three kinds, verified to fail when the line is removed:

```
--- FAIL: TestTreeNode_Clone_CarriesWrapper
    Clone dropped the wrapper kind: got 0, want 1
    Clone dropped the wrapper kind: got 0, want 2
```

**The orphaned doc comment.** Also correct and worse than cosmetic — I'd inserted `WrapperKind` between `TreeNode`'s doc comment and `TreeNode` itself, so godoc would have rendered `TreeNode` undocumented while `WrapperKind` absorbed its comment. Moved above, properly separated.

Also noting your verification that the tag can only ever *restrict* the existing structural match and never introduce a new false positive — since `createSingleDynamicTree` only stores strings, a field node can't acquire the tag. That's the safety property the change rests on and I hadn't stated it that cleanly.

adnaan:
Converged — this round found no new defect and independently re-verified both prior fixes, including tracing the text-sibling case by hand through the merge loop. Stopping pushes here.

**Filed #505 for the single-level unwrap you flagged.** You're right it's pre-existing — `main` has the same loop — so it's out of scope here. Worth noting it's *easier* to fix on top of this PR than before it: descending through multiple wrappers is only safe once wrappers are tagged, because otherwise the descent would compound the shape guessing this change removes. I've written that into the issue.

**On your check that no other path builds the wrapper shape** — I ran the same audit from the other direction after your `Clone` catch, since checking one instance clearly hadn't discharged the class. Every site that constructs a `TreeNode` by naming fields:

| site | carries `Wrapper`? | correct? |
|---|---|---|
| `NewTreeNode` / `NewTreeNodeWithStatics` | no | yes — fresh nodes, `WrapperNone` is right |
| `TreeNode.Clone` | **now yes** | fixed this round |
| `diff/tree_compare` (×2) | no | yes — diff artifacts |
| `diff/prepare` | no | yes — wire-format artifact |

The three `diff`-side omissions are correct rather than lucky: keying runs only in `internal/parse`, and a grep for `wrappedItemKey`/`allWrappedItemKeys` across `internal/diff` and `internal/build` returns nothing, so no post-parse node is ever re-keyed. Only `Clone` semantically needed the field, because only `Clone` claims to reproduce the original.

Three rounds, three real defects, all mine — whitespace-sensitive propagation, the dropped `Clone` field, and the orphaned godoc. Thanks for the persistence; the first would have shipped a silent keying regression in the common template shape.

PR #506, opened 2026-07-19
fix(discovery): survive a path vanishing mid-walk (closes #502)

### LINKED ISSUE #502: Template auto-discovery aborts on any walk error, so a sibling test's .uploads teardown fails unrelated tests
A recurring CI flake with a known-but-unfixed root cause. It cost a false alarm on #499 today — I diagnosed a red `Test LVT against Core Changes` as a regression before recognising the signature — so it is worth fixing rather than continuing to re-run.

## Symptom

```
New failed: livetemplate.New("fnv-ws"): template auto-discovery failed in "":
readdirent .../internal/upload/.uploads/.tmp/session-<n>/<field>: no such file or directory
```

**The failing test name is not the identifier — the error string is.** The victim is arbitrary: any test calling `livetemplate.New()` with auto-discovery can lose the race. Observed so far on `TestProgressiveEnhancement_Disabled` and `TestFormNoValidate_WS_ExplicitSubmitter`. Match on `template auto-discovery failed` + `readdirent .../.uploads/`, not on a test name.

## Root cause

Two halves, both in-tree.

**1. The walk aborts on any error** — `internal/discovery/discovery.go`, in the `filepath.WalkDir` callback inside `DiscoverTemplateFiles`:

```go
err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
    if err != nil {
        return err          // any error anywhere kills the whole discovery
    }
```

A path disappearing mid-walk is not a template-discovery failure, but it is treated as one, and `livetemplate.New()` returns the error.

**2. `.uploads` is walked** — `ignoredTemplateDirs` in the same file lists only `node_modules`, `vendor`, `.git`. Meanwhile `upload_test.go` tears down with `os.RemoveAll(".uploads")` (two call sites). So a sibling test deletes a subtree while another test's `New()` is walking it.

`-race` widens the window rather than creating it: this is a filesystem race, so a plain `go test ./...` can hit it too. Suspected — not confirmed — to be behind the v0.19.1 `release.sh` abort, which failed once at `build_and_test` and then passed three consecutive runs.

## Suggested fix

Both parts, in `internal/discovery/discovery.go`:

1. **Tolerate a vanished path.** In the walk callback, treat `fs.ErrNotExist` as "skip", not "fail":

```go
if err != nil {
    if errors.Is(err, fs.ErrNotExist) {
        return nil   // fs.SkipDir when d != nil && d.IsDir()
    }
    return err
}
```

Concurrent deletion elsewhere in the tree should never fail template discovery. This is the general fix — any transient directory can trigger the same abort, not just `.uploads`.

2. **Add `.uploads` to `ignoredTemplateDirs`.** It holds uploaded files by definition, never templates, so walking it is pure cost plus this risk. Narrows the window and skips useless I/O. Note this only covers the default name; the upload dir is configurable, which is why (1) is the actual fix and (2) is the cheap complement.

## Diagnostic worth keeping

The differential proof that it is this flake and not a regression, which is faster than log-spelunking:

> the **same commit** passes the test in the plain **Unit Tests** job and fails only in the **-race LVT** job.

Same commit, divergent results ⇒ non-determinism. Until this is fixed, the action on seeing it is re-run, not investigate.

## Acceptance

- A vanished path mid-walk does not fail `DiscoverTemplateFiles`
- `.uploads` is skipped
- A regression test pinning that a walk error of `fs.ErrNotExist` is tolerated (the `.uploads`-skipped case is deterministic and easy; the ENOENT tolerance is the one worth a targeted test)

adnaan:
Closes #502.

Template auto-discovery returned **every** `filepath.WalkDir` error, so a directory removed while the walk was in progress aborted the whole thing and `livetemplate.New()` failed:

```
template auto-discovery failed: readdirent .../.uploads/.tmp/session-N/…:
no such file or directory
```

The remover was the upload tests tearing down `.uploads`; the victim was whichever unrelated test happened to be constructing a `Template` at that moment. It cost a false regression alarm on #499 today — a red `Test LVT against Core Changes` that looked like a real break until the error string identified it.

## Two changes

**`walkErrAction` treats `ErrNotExist` below the root as "skip what's gone and keep walking."** A path disappearing underneath a walk isn't a discovery failure.

Deliberately narrow:

| case | behaviour | why |
|---|---|---|
| vanished subdirectory | `SkipDir` | nothing left to descend into |
| vanished file | `nil` | `SkipDir` on a non-directory skips the rest of the containing directory, silently dropping sibling templates |
| **missing root** | **error** | `WalkDir` reports it through the same callback; it means the caller asked to search somewhere that doesn't exist, and surfacing that beats reporting "no templates found" |
| non-`ErrNotExist` (permissions, …) | error | tolerance widens to the concurrent-removal case and nothing else |

**`.uploads` joins `ignoredTemplateDirs`.** It holds uploaded files, never templates, so walking it is pure cost plus exposure to this race. A complement, not the fix — the upload directory is configurable and any transient directory triggers the same abort, so `walkErrAction` is what closes the class.

## Tests reproduce the reported failure, not an approximation

With the fix reverted, the concurrency test fails with the same signature seen in CI:

```
--- FAIL: TestDiscoverTemplateFiles_ToleratesVanishingPaths
    iteration 254: discovery failed while a sibling tree was being removed:
    readdirent /tmp/.../scratch/tmp/session-1: no such file or directory
```

and the `.uploads` test picks up an upload as a template.

**The concurrency test deliberately churns a directory NOT named `.uploads`.** Using that name would make it pass by never walking the tree — proving the skip instead of the tolerance. It also asserts the seeded template is still found each iteration, so tolerating a vanished path can't cost us files that are actually there.

`TestDiscoverTemplateFiles_MissingBaseDirStillErrors` is the counterweight: the tolerance must not widen far enough to turn a caller's wrong path into a silent empty result.

## Not introducing a flake while fixing one

Would be a poor outcome for this particular PR, so: **12 consecutive runs** of the concurrency test, plus the package green under `-race`. Full suite green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to LiveTemplate will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Template auto-discovery no longer fails when a directory disappears while it
+  is being searched.** `livetemplate.New()` walks the template directory, and any
+  error from that walk aborted it — including a transient directory being removed
+  by something else at that moment, which surfaced as
+  `template auto-discovery failed: readdirent …: no such file or directory` from
+  an unrelated part of the app. A path vanishing underneath the walk is now
+  skipped rather than fatal. Deliberately narrow: a missing *base* directory
+  still errors, since that means the caller pointed at somewhere that does not
+  exist, and non-ENOENT failures such as permissions still surface. `.uploads` is
+  also skipped outright, being uploaded files rather than templates. (#502)
+
 ## [v0.19.1] - 2026-07-18
 
 Documentation-only. No library behavior changes; released so the docs site,
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"errors"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


adnaan:
No correctness issue this round — thanks for walking `filepath.WalkDir`'s actual semantics rather than taking the branches on trust.

**Applied, with a correction: the CLAUDE.md citation doesn't exist.** The quoted rule —

> *"Don't reference the current task, fix, or callers ... since those belong in the PR description and rot as the codebase evolves."*

has no match in either `livetemplate/CLAUDE.md` or the global one. What the project actually says is `**No unnecessary comments** - Code should be self-documenting`. Flagging it specifically so it isn't cited as precedent in a later round — CLAUDE.md's own note on this bot warns its verifications aren't infallible, and a fabricated quote is more durable than a wrong line number because it reads like policy.

The *underlying* point stands on its own merits, so I've applied it. Trimmed to the durable why — root versus non-root, `SkipDir` only for directories, why `.uploads` is a complement — and dropped the literal error string and the account of which tests were removing the directory. Those belong in the PR and issue, which have them.

**Applied: the unreachable branch.** Good catch that `WalkDir` only ever pairs a non-nil error with a directory entry or a nil entry at the root, so the non-directory path can't fire through a real walk. I've kept it — it's cheap insurance if that stdlib pairing ever changes, and getting it wrong silently drops sibling templates — but the comment now says it's defensive rather than implying it's a live case, so nobody burns time trying to provoke it.

**Applied: the `.uploads` shadowing callout.** A project with a real template directory literally named `.uploads` loses it from discovery. Now noted at the map entry, and it reinforces why `walkErrAction` is the fix and the ignore entry only a complement — the entry is a name-based guess, the error handling is not.

PR #507, opened 2026-07-19
feat(parse): key through nested wrappers, not just one level (closes #505)

### LINKED ISSUE #505: wrappedItemKey unwraps only one level, so doubly-wrapped keyed range items fall back to content hashing
Pre-existing limitation, not a regression. Surfaced by claude-review while reviewing #504, and confirmed present in `main`'s version of the same function.

## The limitation

`wrappedItemKey` (`internal/parse/keys.go`) reads a range item's real `data-key` through **one** level of parser-created wrapper: it finds the single nested `*TreeNode` child and requires *that child* to carry the key attribute.

A range item wrapped twice therefore misses:

```
{{range .Items}}{{if .A}}{{if .B}}<li data-key="{{.Path}}">{{.Name}}</li>{{end}}{{end}}{{end}}
```

Here item → wrapper → wrapper → `<li data-key>`. The first child is itself a wrapper with `["", ""]` statics, so `hasExplicitKeyAttribute` fails on it and the item falls back to content-hash keying.

## Why it matters

Content hashes change when the item's content changes, so a keyed range that hits this loses stable identity: the diff engine emits remove+insert instead of a scoped `["u", key, …]`, and the client rebuilds those DOM nodes rather than patching them. Exactly what explicit `data-key` is supposed to prevent.

Low severity — one wrapper is the common shape, and the fallback is correct, just less efficient.

## Suggested direction

Walk down while the child is itself a wrapper, bounded by a small depth cap, and take the first descendant that carries a key attribute. The wrapper kinds added in #504 make this safe to do without re-introducing shape guessing: the loop can descend only through nodes the parser actually tagged, rather than through anything that happens to look like a wrapper.

A depth cap matters — the existing all-or-nothing rule in `allWrappedItemKeys` means one deep item would otherwise force the whole range down the same path on every render.

## Not a #504 regression

`main` has the same single-level loop; #504 changed how a wrapper is *identified*, not how many levels are traversed. Worth doing on top of #504 rather than before it, since the kind tag is what makes the descent safe.

adnaan:
Closes #505.

Nested constructs stack wrappers, so `{{if}}{{if}}<li data-key="…">` puts the keyed element **two** levels below the range item. `wrappedItemKey` looked exactly one level down, so those items fell back to content hashing despite carrying a perfectly good explicit key.

The cost is real: content hashes change with content, so editing an item changed its key, and the client removed and re-inserted the row instead of patching it — exactly what `data-key` exists to prevent.

## The change

Descend while the node is a wrapper, stopping at the first child carrying a key attribute.

**This is only safe because wrappers are tagged (#497).** Each step tests a node the parser *marked*, never a shape that happens to resemble one — so the descent can't re-introduce the guessing that issue removed. It's why #505 was worth doing after #504 rather than before it.

**Bounded at four levels.** `allWrappedItemKeys` is all-or-nothing, so an item that never yields a key makes the whole range pay the full descent on every render. Four is well past what real templates stack (`{{if}}{{if}}` is already unusual) while keeping the miss cost flat.

## Measured, not assumed

A twelve-shape matrix on `main` and on this branch. **Exactly three lines change:**

```
  recursive              datakey=true      (unchanged)
  if_keyed               datakey=true      (unchanged)
  if_keyed_indented      datakey=true      (unchanged)
  if_keyed_container     datakey=true      (unchanged)
  plain_keyed            datakey=false     (unchanged)
  unkeyed                datakey=false     (unchanged)
- double_if              datakey=false
+ double_if              datakey=true
- double_if_indented     datakey=false
+ double_if_indented     datakey=true
- triple_if              datakey=false
+ triple_if              datakey=true
  double_if_unkeyed      datakey=false     (unchanged — no key at any depth)
```

Indented and container-wrapped variants are in the matrix deliberately: a minified-only matrix is what let a formatting-sensitive regression through on #504, and I'd rather not repeat that.

## Boundary and gates

The cap is pinned in both directions — four levels key, five falls back. And the descent test was verified to fail when the loop is reverted to a single level:

```
--- FAIL: .../two_wrappers
--- FAIL: .../three_wrappers
--- FAIL: .../at_the_descent_limit
```

`TestSoleNestedChild` pins the "exactly one nested child" rule the descent relies on — a node with two nested trees is real content, not a wrapper, so there's no single element for identity to come from.

Full suite green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,30 @@ All notable changes to LiveTemplate will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Template auto-discovery no longer fails when a directory disappears while it
+  is being searched.** `livetemplate.New()` walks the template directory, and any
+  error from that walk aborted it — including a transient directory being removed
+  by something else at that moment, which surfaced as
+  `template auto-discovery failed: readdirent …: no such file or directory` from
+  an unrelated part of the app. A path vanishing underneath the walk is now
+  skipped rather than fatal. Deliberately narrow: a missing *base* directory
+  still errors, since that means the caller pointed at somewhere that does not
+  exist, and non-ENOENT failures such as permissions still surface. `.uploads` is
+  also skipped outright, being uploaded files rather than templates. (#502)
+- **Range items whose keyed element sits under more than one wrapper keep their
+  stable identity.** Nested constructs stack wrappers, so
+  `{{if}}{{if}}<li data-key="…">` puts the keyed element two levels below the
+  range item; key lookup only looked one level down and fell back to hashing the
+  item's content. Because a content hash changes when the content does, editing
+  such an item changed its key and the client removed and re-inserted the row
+  instead of patching it — the churn `data-key` exists to avoid. Lookup now
+  descends through nested wrappers, bounded at four levels. Items with no key at
+  any depth still use content hashes, as before. (#505)
+
 ## [v0.19.1] - 2026-07-18
 
 Documentation-only. No library behavior changes; released so the docs site,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Template auto-discovery no longer fails when a directory disappears while it
-  is being searched.** `livetemplate.New()` walks the template directory, and any
-  error from that walk aborted it — including a transient directory being removed
-  by something else at that moment, which surfaced as
-  `template auto-discovery failed: readdirent …: no such file or directory` from
-  an unrelated part of the app. A path vanishing underneath the walk is now
-  skipped rather than fatal. Deliberately narrow: a missing *base* directory
-  still errors, since that means the caller pointed at somewhere that does not
-  exist, and non-ENOENT failures such as permissions still surface. `.uploads` is
-  also skipped outright, being uploaded files rather than templates. (#502)
 - **Range items whose keyed element sits under more than one wrapper keep their
   stable identity.** Nested constructs stack wrappers, so
   `{{if}}{{if}}<li data-key="…">` puts the keyed element two levels below the
--- a/internal/parse/keys.go
+++ b/internal/parse/keys.go
@@ -103,9 +103,10 @@ func generateItemHash(item *TreeNode) string {
 // wrapped one, and narrowing this to WrapperInvocation would drop {{if}}-wrapped
 // keyed ranges back to content hashing — a keying regression, not a fix.
 //
-// Returns ("", false) when the item is not a wrapper, holds more than one nested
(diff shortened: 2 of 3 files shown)


adnaan:
Both applied — and one of them is worth more than it looked.

**The CHANGELOG gap was wider than this PR.** You asked whether the repo batches at release time; it doesn't — it keeps a curated changelog, and I curated it twice today. But **nothing since v0.19.1 was recorded at all**, including work already merged. So `[Unreleased]` now exists, covering the two user-facing fixes in flight: discovery surviving a vanishing path (#502) and range items keeping stable identity under nested wrappers (#505).

Deliberately selective rather than exhaustive. #503 and #504 were behaviour-preserving refactors — both verified byte-identical against main — so they'd be noise in a user-facing changelog, and #499/#501 are release-script changes that don't affect anyone consuming the library. Good catch; I'd have shipped a second release with an empty changelog otherwise.

**Applied: issue numbers out of `keys.go`.** Including one that predates this branch — leaving a single stale reference after removing its neighbours would have been worse than either. The durable why stays: wrappers are tagged, so the descent never guesses from shape.

**One correction, same as on #506.** The CLAUDE.md quote —

> *"Don't reference the current task, fix, or callers... since those belong in the PR description and rot as the codebase evolves."*

— does not exist in either `livetemplate/CLAUDE.md` or the global one. The project's actual guidance is `**No unnecessary comments** - Code should be self-documenting`. This is the second PR where it's been cited verbatim, which is why I'm flagging it again rather than letting it pass: a fabricated policy quote reads like settled convention, and since prior comments feed the next review, an uncorrected one becomes self-reinforcing precedent.

The substance was right both times and I've applied it both times. It's the citation I'm declining, not the advice.

Thanks for tracing the off-by-one explicitly — `keyed(4)` succeeding and `keyed(5)` falling back mid-descent is exactly the boundary I wanted pinned, and having it independently derived rather than taken from the test names is worth more.

adnaan:
You're right, and this was a genuine mistake rather than a nit. Fixed.

The `[Unreleased]` section documented the discovery fix while this branch contains none of it — no `internal/discovery` changes in the diff at all. If this PR landed first, main would have carried release notes for a fix main didn't have. Exactly the factual-claims-in-docs failure the project's review workflow warns about, and your check of the actual `WalkDir` callback on this branch is the right way to have caught it.

What makes it worth more than "moved the bullet": **I reasoned my way into it deliberately.** Both entries went here specifically to avoid a `CHANGELOG.md` conflict between the two PRs. That's optimising the wrong variable — a merge conflict costs seconds to resolve, publishing a false fix claim doesn't, and I weighed only the first. The entry now lives in #506 with its code, and the two land together or not at all. I'll take the conflict if both land close together.

Also applied the doc nit: the fallback list omitted the item-level `hasExplicitKeyAttribute(item.Statics)` short-circuit at the top. Now enumerated.

One correction carried over from the earlier round — the CLAUDE.md quote about not referencing the current task in comments still doesn't exist in either file; the project's actual line is `**No unnecessary comments** - Code should be self-documenting`. I applied the advice and removed the issue numbers, including one predating this branch. It's the citation I'm declining, not the substance.

Noting for the record that this round's finding is the kind the "measured, not assumed" framing in my own PR description doesn't cover — I verified the *code* behaviour exhaustively and then wrote a changelog entry describing a different branch's code. Verification of the diff wouldn't have caught it; only checking the claim against the diff would.

PR #512, opened 2026-07-20
fix(release): promote the curated [Unreleased] instead of regenerating

### LINKED ISSUE #511: release.sh discards the curated [Unreleased] section on every successful release
`scripts/release.sh` regenerates `CHANGELOG.md` from a raw `git log` dump on every release, so a hand-written `## [Unreleased]` section never becomes the released notes. It survives — stranded *below* the new version, still headed `[Unreleased]` — and the notes that actually ship are bare commit subjects.

Hit on **v0.19.1**, **v0.20.0** (core) and **client v0.19.1**, **v0.20.0**. Each needed a manual curation commit afterwards, plus a `gh release edit` because the GitHub release notes are extracted from the clobbered file.

## What happens

`generate_changelog` rebuilds the file from scratch:

```sh
echo "## [v$new_version] - $(date +%Y-%m-%d)"
echo "### Changes"
git log "$prev_tag"..HEAD --pretty=format:"- %s (%h)" --no-merges | grep -v "^- Merge "
tail -n +7 CHANGELOG.md          # everything from line 7 of the OLD file
```

The header block is lines 1–6, so `tail -n +7` re-appends the old body — including the `## [Unreleased]` section that was sitting at line 8. Result:

```
## [v0.20.0] - 2026-07-20
### Changes
- ci(review): let the reviewer read the files it reviews (#510) (27266e21)
- docs(changelog): file the v0.8.5 breaking notice under v0.8.5 (#509) (370261f0)
- feat(parse): key through nested wrappers, not just one level (#507) (985a8412)
  ... 6 more, refactors and CI chores billed equally with the fixes

## [Unreleased]        <- the curated text, now orphaned and mislabelled
### Fixed
- **Template auto-discovery no longer fails when a directory disappears...**
```

## Why it matters beyond tidiness

- **The shipped notes are the wrong artifact.** A reader deciding whether to upgrade gets nine commit subjects with `ci(review):` and `refactor(parse):` given the same weight as the two fixes that change behaviour.
- **It contradicts the file's own stated format.** Both changelogs open with *"The format is based on [Keep a Changelog]"* — whose entire workflow is: maintain `[Unreleased]`, promote it on release. The script does the opposite.
- **It silently produced a duplicate-header defect.** #509 fixed a second `## [Unreleased]` that had been sitting mid-file since the v0.8.x era, invisible because release-note extraction stops at the first match. That is this bug's residue.
- **Installing `git-chglog` is not the fix.** Core's `generate_changelog` prefers it when present, but that path also regenerates wholesale (`git-chglog -o CHANGELOG.md`). Client has no such branch at all.

## Proposed fix

Promote rather than regenerate:

1. If a `## [Unreleased]` section exists and is non-empty, retitle it `## [vX.Y.Z] - <date>` and leave its content alone.
2. Only fall back to the `git log` dump when there is no `[Unreleased]` section — preserving today's behaviour for releases nobody curated.
3. Refuse to proceed if more than one `## [Unreleased]` header exists, rather than silently picking one.

Worth adding a check that the promoted section is non-empty, so an unwritten changelog fails loudly at release time instead of shipping an empty version heading.

## Scope

Both scripts, same defect, slightly different shapes:

- `livetemplate/scripts/release.sh` — `generate_changelog`, plus the unused `git-chglog` branch
- `client/scripts/release.sh` — `generate_changelog`, no `git-chglog` branch

Related: #499 (aborted-run restore) and livetemplate/client#151 (its port) fixed the *abort* path in the same function's neighbourhood. This is the complementary case — it fires on every run that **succeeds**.

adnaan:
Closes livetemplate/livetemplate#511.

## The problem

`release.sh` rebuilt `CHANGELOG.md` from a raw `git log` dump on every release, so a hand-written `## [Unreleased]` section never became the release notes. It survived only as a stranded section *below* the new version, still titled `[Unreleased]`, while the notes that shipped were bare commit subjects.

Hit four times in two days — core v0.19.1 and v0.20.0, client v0.19.1 and v0.20.0 — each needing a manual curation commit plus a `gh release edit`, since `publish_github` extracts its notes from the clobbered file.

## The fix

**Promote instead of regenerate.** Retitle `[Unreleased]` as the release heading, leave its content alone. This is what the file's own header already claims: Keep a Changelog's workflow is *maintain `[Unreleased]`, promote on release*.

The commit-subject dump stays as the fallback for releases nobody curated, so nothing regresses when the section is absent or empty.

**Refuse to start on a duplicate `[Unreleased]`.** Release-note extraction stops at the first heading, so a duplicate means one section can never ship. That is exactly how the stale v0.8.5 notice sat unnoticed in this file until #509 — this makes the same situation loud:

```
✗ CHANGELOG.md has 2 '## [Unreleased]' headings; there must be at most one

1207:## [Unreleased]
1211:## [Unreleased]

Merge them into the topmost one, or retitle the stale section with the
version it actually shipped in, then re-run.
```

It runs in `check_prerequisites`, before anything mutates.

## Verification

`scripts/test_release_changelog.sh` — new, sourcing `release.sh` with `RELEASE_SH_LIB=1` so the functions are testable without running a release:

```
1️⃣  Curated [Unreleased] is promoted, content preserved verbatim
2️⃣  Empty [Unreleased] falls through to the commit-subject fallback
3️⃣  Whitespace-only [Unreleased] counts as empty
3️⃣b A curated section does NOT take the fallback
4️⃣  unreleased_body reads only its own section
```

**These were run against the previous `release.sh` to confirm they discriminate.** Two fail there, reporting the exact defect:

```
✗ [Unreleased] survived the promotion
      8:## [Unreleased]
✗ history damaged
```

Tests 2/3 assert on *which branch was taken* rather than the resulting file — the client's fallback writes an "Initial release" block when no tag exists, so the end state differs between repos while the promote-vs-regenerate decision does not. That keeps one test file valid for both.

Duplicate guard verified by injecting two headings and confirming it aborts before the dirty-tree check.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -52,6 +52,10 @@ check_prerequisites() {
         echo ""
         echo "Merge them into the topmost one, or retitle the stale section with the"
         echo "version it actually shipped in, then re-run."
+        echo ""
+        echo "Most likely cause: a release took the commit-subject fallback, which"
+        echo "leaves an empty [Unreleased] in place, and a later curated one was added"
+        echo "above it. The empty section is the one to drop."
         exit 1
     fi
 }
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -39,6 +39,21 @@ check_prerequisites() {
     if ! command -v git-chglog >/dev/null 2>&1; then
         log_warn "git-chglog not installed (optional). Install with: brew install git-chglog"
     fi
+
+    # Refuse to guess which [Unreleased] section is the real one. A duplicate
+    # went unnoticed in this file from v0.8.5 until #509 because release-note
+    # extraction stops at the first heading, so the second could never ship.
+    local unreleased_headings
+    unreleased_headings=$(grep -c '^## \[Unreleased\]' CHANGELOG.md 2>/dev/null || true)
+    if [ "${unreleased_headings:-0}" -gt 1 ]; then
+        log_error "CHANGELOG.md has $unreleased_headings '## [Unreleased]' headings; there must be at most one"
+        echo ""
+        grep -n '^## \[Unreleased\]' CHANGELOG.md
+        echo ""
+        echo "Merge them into the topmost one, or retitle the stale section with the"
+        echo "version it actually shipped in, then re-run."
+        exit 1
+    fi
 }
 
 # Detect a release that was committed locally but never reached origin — a push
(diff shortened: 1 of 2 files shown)


adnaan:
**Both non-blocking observations actioned** (pushed).

**1. Stale empty `[Unreleased]` tripping the new guard.** Good catch — the sequence you described is real: the fallback leaves an empty section behind, someone adds a curated one above it, and the guard fires. Rather than only documenting it, the guard error now names that as the likely cause and says which section to drop:

```
Most likely cause: a release took the commit-subject fallback, which
leaves an empty [Unreleased] in place, and a later curated one was added
above it. The empty section is the one to drop.
```

Better there than in a doc, since that is the moment someone hits it — mid-release, wanting to know which of two identical headings to delete.

**2. Test not wired into CI.** Agreed and filed as a follow-up rather than expanded into this PR, since it should cover `test_cli_smoke.sh` too and that is a workflow change with its own scope. You are right that an unwired test protects nothing.

**3. Hex-token heuristic in the dump assertion.** Agreed it is low-risk; leaving as-is per your own read.

---

**One thing your review prompted that was a real bug in the tests themselves.** Investigating the guard change, the suite started failing — and it turned out not to be that change at all. `took_fallback` piped `generate_changelog` into `grep -q`. Sourcing `release.sh` brings `set -o pipefail` with it, and `grep -q` exits on first match, so the closed pipe could make `generate_changelog` exit non-zero and fail the pipeline. Measured:

```
OLD (piped):              9 failures in 15 runs
NEW (captured variable):  0 failures in 30 runs
```

It presented deceptively — it looked like an unrelated edit had broken the test, and inserting a debug call before the assertion made it pass, which was just the timing shift. Fixed by capturing to a variable.

Issue #452, opened 2026-06-11
Upload field serialization: opt-in (safer-by-default) vs current denylist


adnaan:
Follow-up from the Proxied-upload form-field work (client#130, livetemplate#451).

**Current behavior (shipped):** a Proxied upload auto-fires on file selection and POSTs the enclosing `<form>`'s value fields so `OnUpload` can associate the bytes with a record (e.g. a hidden `id`). The client serializes all successful controls **except** `type="password"` (denylist).

**Concern (review):** because the upload auto-fires (not an explicit submit), a denylist still lets *other* sensitive co-located fields travel to the upload endpoint — CSRF/session tokens, `autocomplete="current-password"` text inputs, hidden secrets. The user has no submit-time awareness.

**Proposal to evaluate:** an **opt-in** contract — serialize only fields explicitly marked to travel with the upload (e.g. `lvt-upload-field`, or `data-lvt-upload`). Safer by default (nothing leaves unless marked), at the cost of requiring app authors to mark the record-id/context fields.

**Touch points if adopted:** client `appendFormFields` (read the attribute), `docs/references/uploads.md` + the `examples/upload-modes` demo, and consuming apps (e.g. checklistkit would mark its hidden `id`). Decision is a framework-contract call for the upload-modes owner.

adnaan:
Additional minor notes from client#130 round-3 review (non-blocking), to fold into the opt-in design:
- **Hidden fields auto-POST**: be explicit in a release note / component doc that *any* hidden field in a form with an auto-upload input is POSTed (CSRF/session tokens especially). The opt-in model resolves this by construction.
- **Name collision**: if a non-file field shares the upload field's name, `formData.append(name, val)` then `formData.set(uploadName, file)` means the file wins and the field value is dropped (correct, but worth a code comment).
- **Multi-file**: when one startUpload carries multiple files, all entries share one sourceInput, so form fields are POSTed once per file (separate requests). Decide the intended contract (all / first / last) under the opt-in design.

adnaan:
Adopted the opt-in proposal. Attribute is **`lvt-upload-with`** (per-field boolean marker), chosen over `lvt-upload-field` (ambiguous — reads as marking the file input itself) and a form-level `lvt-upload-fields="id,kind"` list (drifts silently out of sync when a field is renamed).

- Client: livetemplate/client#150
- Docs + example: livetemplate/docs#120

[...]

Issue #124, opened 2026-02-22
FOLLOW-UP: Replace slog.String("error", err.Error()) with slog.Any("error", err) (from PR #123)


adnaan:
## Context

This follow-up task was identified during the review of PR #123.

**Source PR:** https://github.com/livetemplate/livetemplate/pull/123
**PR Title:** refactor: migrate log.Printf to structured slog logging
**Suggested by:** @claude[bot]
**Priority:** High

## Task Description

Replace all ~30 occurrences of `slog.String("error", err.Error())` with `slog.Any("error", err)` across all files migrated in PR #123.

Passing `err` directly lets slog handlers (e.g. JSON handlers) serialize the full error chain, and allows handlers that support it (like `tint`, `slog-multi`) to use `errors.As`/`errors.Is` for structured unwrapping. Calling `.Error()` degrades it to a plain string and discards everything else.

### Files affected
- `mount.go`
- `session_stores.go`
- `template.go`
- `pubsub/redis.go`
- `internal/upload/multipart.go`

### Pattern
```go
// Before
slog.Error("Mount failed", slog.String("error", err.Error()))

// After
slog.Error("Mount failed", slog.Any("error", err))
```

---
*This issue was automatically created by prmonitor from PR review comments.*

PR #514, opened 2026-07-20
docs(uploads): document lvt-upload-with in the source, not the mirror

adnaan:
Fixes a mistake in the #452 work: the `lvt-upload-with` documentation was written into the **generated mirror** rather than the source.

## What happened

`livetemplate/docs`'s `content/reference/uploads.md` and `client-attributes.md` are mirrors of the files in *this* repo. Their own frontmatter says so:

```yaml
source_repo: "https://github.com/livetemplate/livetemplate"
source_path: "docs/references/uploads.md"
```

I edited the mirror in livetemplate/docs#120 and #122. Those are live on the docs site right now, but the next sync reverts them.

Caught by **livetemplate/docs#124**, the auto-sync from `v0.20.0`, which shows:

```
content/reference/uploads.md (+7/-66)
```

Sixty-six lines removed — every line of the new documentation. The sync PR body even warns about exactly this: *"edits made directly here will be overwritten on the next sync."*

## What this does

Adds the same content here, where it survives:

- **`docs/references/uploads.md`** — the *Sending form fields with an upload* section (marking contract, the three ordering/scoping/collision rules, multi-file behaviour) and *Marked fields are multipart-only* (the per-mode transport table).
- **`docs/references/client-attributes.md`** — the `lvt-upload-with` row.

Verified the prose matches what is currently live on the docs site. Two intentional differences: the issue link is same-repo form here, and I added a line noting the transport warning needs client v0.20.0.

## Merge-order note

livetemplate/docs#124 syncs from tag `v0.20.0`, which predates this commit — so merging it **as-is would still delete the docs**. Once this lands, a subsequent core release will produce a sync carrying the documentation, and that one supersedes #124.

Worth deciding deliberately rather than merging #124 on autopilot.

### DIFF
--- a/client_assets.go
+++ b/client_assets.go
@@ -16,7 +16,7 @@ import "html/template"
 // which is a no-op on older servers).
 //
 // Maintainers: bump this in the same release that adopts a new client version.
-const ClientVersion = "0.18.2"
+const ClientVersion = "0.20.0"
 
 // clientCDNBase is the jsdelivr base URL for the pinned client bundle. jsdelivr
 // serves the published npm package @livetemplate/client.
--- a/docs/references/client-attributes.md
+++ b/docs/references/client-attributes.md
@@ -915,8 +915,12 @@ Handle file uploads with progress tracking.
 | Attribute | Description |
 |-----------|-------------|
 | `lvt-upload` | Upload identifier for tracking |
+| `lvt-upload-with` | Send this field along with an upload fired from the same form. Opt-in — unmarked fields never reach the upload endpoint |
 
-Files are automatically uploaded when the form is submitted, with progress events emitted.
+Uploads fire on file selection, not on form submit, and progress events are
+emitted as the bytes move. Because there is no submit for the user to review,
+nothing else in the form travels with the upload unless it is marked
+`lvt-upload-with` — see [Sending form fields with an upload](uploads.md#sending-form-fields-with-an-upload).
 
 ---
 
--- a/docs/references/client-attributes.md
+++ b/docs/references/client-attributes.md
@@ -1043,6 +1043,7 @@ Directives use CSS custom properties for configuration: `--lvt-scroll-behavior`,
 | Attribute | Description | Example |
 |-----------|-------------|---------|
 | `lvt-upload` | File upload identifier | `lvt-upload="avatar"` |
+| `lvt-upload-with` | Send this field along with an upload fired from the same form. Opt-in — unmarked fields never reach the upload endpoint. Multipart path only; see [Upload Reference](uploads.md) | `<input type="hidden" name="id" lvt-upload-with>` |
 
 ### Preservation Attributes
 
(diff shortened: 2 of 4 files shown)


adnaan:
Good catch, and it's the right instinct — a PR fixing docs drift shouldn't leave a third source of truth telling the old story. Fixed in the same pass.

`UploadStreamer`'s godoc is now written around the marking, with the ordering and `data`-envelope caveats kept as rules that apply *on top of* it rather than as the whole story, and the example shows the marked input alongside the handler:

```go
//	<form>
//	  <input type="hidden" name="id" value="{{.Record.ID}}" lvt-upload-with />
//	  <input type="file" lvt-upload="scan" />
//	</form>
```

Your point about `go doc` is what made it worth doing now rather than as a follow-up: the tooltip is the most likely way an `OnUpload` implementer meets this contract, so it was the *highest*-traffic stale copy, not the lowest.

Swept for others while there and found one more — the `uploads.md` Go sample carried the same claim in a comment (`// "id" must be ordered before the file input`). Also fixed. `go build ./...` clean.

adnaan:
**That was the important catch — thank you.** It is worse than a doc caveat, and I have fixed the cause rather than the wording.

`ClientVersion` was pinned at **0.18.2** while the client shipped 0.19.1 and 0.20.0. Its own godoc says maintainers must bump it in the same release that adopts a new client, and I released both without doing so.

`ClientScriptURL` is built from that constant, so **every application following the documented integration path was being served 0.18.2** — the build that still serializes the whole enclosing form into a Proxied upload, minus password fields. The opt-in fix for that (#452) has been on npm since 0.19.1 and was reaching nobody through the library's own path. A security fix published but not delivered.

Bumped to 0.20.0, and verified the pinned URL actually resolves and carries the fix rather than trusting the version string:

```
$ curl -sI .../@livetemplate/client@0.20.0/dist/livetemplate-client.browser.js
HTTP/2 200
$ curl -s ... | grep -c 'lvt-upload-with'
1
```

This also makes the v0.20.1 release meaningful rather than docs-only — it is what actually delivers the #452 fix to consumers.

On your merge-order point: agreed and already sequenced. livetemplate/docs#124 stays unmerged; once this lands, v0.20.1 produces a sync that carries the documentation and supersedes it.

**Worth a follow-up you have effectively identified:** nothing tests that `ClientVersion` tracks the client's actual released version. `client_assets_test.go` checks the pin is well-formed semver and that the URL is HTTPS, but a stale-but-valid pin passes both — which is exactly how this survived two releases. Filing that.

adnaan:
Right again, and it stung a bit — that is precisely the mistake this PR exists to fix, one level down. I updated one copy of a duplicated table while another copy kept the old story.

Both `### Upload Attributes` tables now carry `lvt-upload-with`, and the walkthrough one links through to the marking section rather than restating it.

While there I found the walkthrough also claimed files **"are automatically uploaded when the form is submitted"**. They fire on *file selection* — which is the entire premise of the opt-in design, since there is no submit for the user to review what leaves the page. So that sentence was not just stale, it asserted the opposite of the reason the feature exists. Corrected.

Swept for the same claim elsewhere in `docs/references/` and found no others.

Three rounds, three real functional findings — the stale `ClientVersion` pin especially. No convergence claim from me; if there is a fourth, I will take it.

Issue #513, opened 2026-07-20
Shell test scripts don't run in CI — test_cli_smoke.sh and test_release_changelog.sh


adnaan:
`scripts/` has two shell test scripts and neither runs in CI. They only execute when someone remembers to invoke them by hand, which means the things they protect can regress silently.

- `scripts/test_cli_smoke.sh` — builds the CLI and exercises the generator
- `scripts/test_release_changelog.sh` — added in #512, covers the `[Unreleased]` promotion logic

Raised by the review bot on #512. Its point is sharp: an unwired test protects nothing, and the changelog one exists precisely because a `release.sh` change silently broke changelog handling four releases running.

## Why it matters more for the release script than for most code

`release.sh` is only exercised during a release — i.e. under time pressure, on `main`, with a tag about to be pushed. That is the worst possible moment to discover a regression, and it is exactly the situation the last four releases produced. A unit test that runs on every PR moves that discovery to the PR.

## Suggested shape

Add a job to the existing test workflow that runs both scripts. `test_release_changelog.sh` is fast (~1s), self-contained, and needs no network — it sources `release.sh` with `RELEASE_SH_LIB=1` and works in a temp dir. `test_cli_smoke.sh` needs a Go build, so it may belong alongside the existing Go job rather than in a new one.

Two things worth checking while wiring:

- **Shebang portability.** `test_cli_smoke.sh` uses `#!/bin/bash`, which does not exist on NixOS and some other distros; `test_release_changelog.sh` and `release.sh` both use `#!/usr/bin/env bash`. Worth standardising on the latter.
- **Whether the client repo should do the same.** `client/scripts/test_release_changelog.sh` is the same file and has the same gap (livetemplate/client#154).

## Related

- #511 — the changelog defect the test covers
- #512 — the PR that added the test

PR #516, opened 2026-07-20
fix(release): refuse to release on a stale client pin

### LINKED ISSUE #515: Nothing catches a stale ClientVersion pin
`ClientVersion` (`client_assets.go`) sat at `0.18.2` while `@livetemplate/client` shipped 0.19.1 and then 0.20.0. Nothing failed.

That is not a small miss. `ClientScriptURL` is built from the constant, so every application using the documented integration path — `<script src="{{ .ClientScriptURL }}">` — was served 0.18.2, the build that still serializes the whole enclosing form into a Proxied upload minus password fields. The opt-in fix (#452) was published to npm and reaching nobody through the library's own path.

Fixed in #514. This issue is about why nothing noticed for two releases.

## Why the existing tests don't catch it

`client_assets_test.go` checks two things, and a stale pin passes both:

- `TestClientVersionIsSemver` — that the value is well-formed `MAJOR.MINOR.PATCH`. `"0.18.2"` is.
- The URL test — that `ClientScriptURL` is HTTPS and carries the exact pinned version. It does.

Both assert the pin is *well-formed*. Neither asserts it is *current*. The only control was the "Maintainers: bump this…" line in the constant's own godoc, i.e. a human remembering — and the release script does not prompt for it either.

## What would have caught it

Roughly in order of strength:

1. **Assert the pinned version resolves and is the client's latest release.** Query the npm registry or the client repo's latest tag and compare. Needs network, so it belongs in a CI-only or build-tagged test rather than the default `go test ./...`.
2. **Assert the pinned bundle exists on the CDN.** Weaker — catches a typo'd or unpublished version, not a stale-but-real one. Still worth having, since it is cheap.
3. **Add it to the release checklist.** `scripts/release.sh` could refuse to release when `ClientVersion` trails the client's latest published release, the same way the client's script already validates its major.minor against core's latest release. That check already exists in the other direction — this is its mirror.

(3) is the one that matches how the mistake actually happens: the pin goes stale *at release time*, and the release script is the thing running then.

## Related

- #514 — the bump, and the review that surfaced it
- #452 — the security fix that was not being delivered
- #513 — the sibling "test exists but nothing runs it" gap

adnaan:
Closes #515.

## Why

`ClientVersion` decides which bundle every application on the documented `<script src="{{ .ClientScriptURL }}">` path loads. Nothing checked that it was current, and it sat at **0.18.2 across two core releases** while the client shipped 0.19.1 and 0.20.0 — so the upload field-serialization fix (#452) was published to npm and reached nobody through the library's own path. A security fix shipped but not delivered. Corrected in #514; this is the guard so it cannot recur.

The existing unit tests assert the pin is **well-formed** semver and that the URL is HTTPS. A stale-but-valid pin passes both — which is exactly how this survived. The only real control was a *"Maintainers: bump this"* line in the constant's godoc, i.e. someone remembering.

## What it does

`release.sh` compares the pin against the client's latest release, before anything else happens:

| State | Behaviour |
|---|---|
| equal | proceed |
| **behind** | refuse — names both versions and the bump to make |
| **ahead** | refuse — that version is unpublished, so the URL 404s |

Two deliberate asymmetries:

- **`behind` is overridable** (`LVT_ALLOW_CLIENT_PIN_DRIFT=1`). Staying behind is occasionally correct — a client release carrying a wire change this server can't serve yet shouldn't be adopted. **`ahead` is not overridable**: no intent makes a 404 right.
- **A failed lookup does not block the release.** Blocking on a GitHub API hiccup would be worse than proceeding — but it reports `SKIPPED` rather than letting silence read as a pass. That distinction is itself tested.

## Verification

**Against the real regression, with a live lookup.** Set the pin back to 0.18.2 and ran `--dry-run`:

```
✗ ClientVersion is behind the released client

  Pinned in client_assets.go: 0.18.2
  Latest @livetemplate/client: 0.20.0

ClientScriptURL is built from this constant, so releasing now ships
0.18.2 to every application using the documented <script> path — any
client fix released since then reaches nobody.

Bump ClientVersion to 0.20.0 and re-run. If this release is meant to
stay on 0.18.2 (e.g. 0.20.0 carries a wire change this server cannot
serve yet), re-run with LVT_ALLOW_CLIENT_PIN_DRIFT=1.
```

The current 0.20.0 pin passes: `✓ Client pin is current (0.20.0)`.

**`scripts/test_release_client_pin.sh`** — 15 checks covering version ordering (including `0.9.0 < 0.10.0`, which a string compare gets wrong), the pin parse, and all five decisions including both override cases. 0 failures in 15 consecutive runs.

Two bugs found and fixed while writing those tests, both worth knowing about:

- The `gh` stub read the wrong variable. `check_client_pin` declares `local latest`, and bash's dynamic scoping meant the stub resolved *that* — empty, mid-assignment — instead of the test's. A test then passed **spuriously**, because the skip message happens to contain the word "current".
- The result helper couldn't be a subshell ending in `echo "EXIT:$?"`: `check_client_pin` calls `exit 1`, which terminates the subshell before that line runs, losing the status and killing the harness via `set -e`.

## Note

Like the other shell tests, this doesn't run in CI — #513 tracks that, and it applies here too.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -147,12 +147,24 @@ check_client_pin() {
     latest=$(gh release list --repo livetemplate/client --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null || true)
     latest=${latest#v}
 
-    # A failed lookup is no evidence either way, and blocking a release on a
-    # GitHub API hiccup is worse than proceeding. Say the check was SKIPPED
-    # rather than letting silence read as a pass.
-    if [ -z "$latest" ]; then
-        log_warn "SKIPPED the client-pin check — could not reach the client repo"
+    # Require a semver-shaped answer rather than merely a non-empty one. jq
+    # yields the literal string "null" when the release list is empty, which is
+    # non-empty, survives the "v" strip, and sorts *below* any real version — so
+    # a bare -z check would let it through and refuse the release citing
+    # "Latest: null". Anything unparseable is treated as no answer at all.
+    if ! printf '%s' "$latest" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+'; then
+        log_warn "SKIPPED the client-pin check — no usable version from the client repo"
         log_warn "ClientVersion is $pinned; confirm by hand that it is current"
+        return 0
+    fi
+
+    # version_order leans on `sort -V`. Absent it, sort still succeeds and
+    # returns a lexicographic order — silently wrong, and inside a [ ] test that
+    # `set -e` does not trap. Probe with a pair whose lexicographic and numeric
+    # orders disagree, and skip rather than trust a bad comparison.
+    if [ "$(printf '0.10.0\n0.9.0\n' | sort -V | head -1)" != "0.9.0" ]; then
+        log_warn "SKIPPED the client-pin check — this sort(1) lacks working -V"
+        log_warn "ClientVersion is $pinned, client released $latest; compare by hand"
         return 0
     fi
 
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -107,6 +107,95 @@ get_current_version() {
     cat VERSION | tr -d '\n'
 }
 
+# The @livetemplate/client version this release will point browsers at.
+read_client_pin() {
+    grep -oE '^const ClientVersion = "[0-9]+\.[0-9]+\.[0-9]+"' client_assets.go 2>/dev/null \
+        | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' || true
+}
+
+# "same" | "older" | "newer", describing $1 relative to $2.
+version_order() {
+    if [ "$1" = "$2" ]; then
+        echo same
+    elif [ "$(printf '%s\n%s\n' "$1" "$2" | sort -V | head -1)" = "$1" ]; then
+        echo older
+    else
+        echo newer
+    fi
+}
+
+# Refuse to release while ClientVersion trails the client's latest release.
+#
+# ClientScriptURL is built from that constant, so it decides which bundle every
+# application on the documented <script src="{{ .ClientScriptURL }}"> path
+# loads. A stale pin means a published client fix reaches nobody — which is what
+# happened between client 0.19.1 and core v0.20.1: the upload field-serialization
+# fix sat on npm for two core releases while the pin stayed at 0.18.2 (#452, #515).
+#
+# Nothing else catches this. The unit tests assert the pin is well-formed semver
+# and that the URL is HTTPS; a stale-but-valid pin passes both.
+check_client_pin() {
+    local pinned latest
+    pinned=$(read_client_pin)
+
+    if [ -z "$pinned" ]; then
+        log_error "Could not read ClientVersion from client_assets.go"
+        echo "Expected a line like: const ClientVersion = \"1.2.3\""
+        exit 1
+    fi
+
+    latest=$(gh release list --repo livetemplate/client --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null || true)
+    latest=${latest#v}
+
+    # A failed lookup is no evidence either way, and blocking a release on a
+    # GitHub API hiccup is worse than proceeding. Say the check was SKIPPED
+    # rather than letting silence read as a pass.
+    if [ -z "$latest" ]; then
+        log_warn "SKIPPED the client-pin check — could not reach the client repo"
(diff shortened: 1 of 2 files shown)


adnaan:
Both actioned — the `null` one was a real defect, and it failed in the worse direction.

**`null` from an empty release list.** Confirmed exactly as you described, and the consequence is nastier than a missed skip: `"null"` sorts *below* any real version, so the guard reported the pin as **behind** and refused a valid release with

```
Bump ClientVersion to null and re-run.
```

Blocking on garbage is worse than skipping — it stops a good release and sends the reader after a version that does not exist. Rather than special-casing the string, the answer now has to be semver-shaped, which also covers empty, and whatever a future API change might return.

**`sort -V`.** Took your point about `set -e` not trapping it — a `sort` without `-V` succeeds and silently returns lexicographic order, which is the failure mode this guard exists to prevent, one layer down. Probed with a pair whose lexicographic and numeric orders disagree (`0.9.0` / `0.10.0`), skipping rather than trusting a bad comparison.

Both skips announce themselves, consistent with the lookup-failure path: this guard should never report OK when it means it did not check.

**Verified both discriminate**, rather than just adding green checks:

- restored the bare `-z` check → new test fails, reproducing `Bump ClientVersion to null`
- stubbed a `sort` that drops `-V` → probe test fails without the probe

17 checks, 0 failures in 12 runs. Live `--dry-run` still reports `✓ Client pin is current (0.20.0)`.

adnaan:
All three fixed. The override one was the worst kind of bug — **the escape hatch was easier to open than intended**, and you're right that it's precisely the failure mode this PR exists to close.

`LVT_ALLOW_CLIENT_PIN_DRIFT=false` silently *enabled* the bypass. A guard whose off switch reads as on is worse than no guard, because it reports success.

Now parsed against the same vocabulary as `parseBool` in `config.go` (`true/1/yes/on`, `false/0/no/off`), erroring on anything else rather than defaulting either way. I went past your `= "1"` suggestion deliberately: that fixes the dangerous direction but introduces a quieter one, where someone writes `=true`, gets a blocked release, and is told to set the variable they already set. Refusing to guess covers both.

**Unreadable pin** — covered. I used a `const ( ... )` block rather than deleting the line, since a reformatting is the realistic way this regex stops matching, and it confirms the guard exits 1 rather than proceeding without a version to compare.

**`sort` stub** — correct, and thanks for reading it that closely. `${@/-V/}` left an empty positional argument, so GNU sort read `""` as a filename and errored; the probe passed for the wrong reason and leaked stderr into the captured output. Filtering the flag out gives a real lexicographic sort, which is what the comment claimed all along.

**Verified the new override cases discriminate** — restored the `-n` check and all three fail:

```
✗ a 'false' override must not let a stale pin through
✗ a '0' override must not let a stale pin through
✗ unparseable override must not be assumed either way
```

22 checks, 0 failures in 10 runs.

Three rounds, three rounds of real findings — no convergence claim from me.

adnaan:
Fixed the selection, and you were right that it matters more here than it would elsewhere — a wrong "latest" read hard-blocks every release, and the `ahead` branch has no override to escape it.

**`latest_client_release`** now:
- passes `--exclude-drafts --exclude-pre-releases` so a draft can't be read as newest (which would be the unpublished-version case arriving through the front door)
- takes the true semver **max via `sort -V`**, not `--limit 1`'s creation-time order
- filters to **end-anchored bare semver**, so `0.21.0-rc.1` is dropped rather than compared — your third point. This is the part that stays honest even though the `--exclude` flags can't be exercised through a `gh` stub, so it's what the tests actually pin.

The "null" handling collapsed out as a bonus: `--jq '.[].tagName'` on an empty list yields nothing, not `"null"`, and the end-anchored filter would drop it anyway.

**On the regex-coupling minor:** left as-is deliberately. It's the intended fail-closed behaviour — a reformat of `client_assets.go` that the anchor stops matching hard-stops the release with "Could not read ClientVersion" rather than silently guessing, and there's a test for exactly that (the `const ( ... )` block case). A looser regex would trade a loud, tested stop for a quiet wrong read, which is the wrong direction for this guard.

**Verified the four selection tests discriminate** — reverted to the list-order lookup and all four fail, including one where `0.21.0-rc.1` was selected as newest:

```
✗ picks the semver max, not the list order       → got 0.19.1
✗ 0.10.0 outranks 0.9.0                           → got 0.9.0
✗ a -rc pre-release is excluded                   → got 0.21.0-rc.1
```

26 checks, 0 failures in 10 runs. Live `--dry-run` against the real client repo still resolves `0.20.0` through the new max path.

PR #519, opened 2026-07-23
feat(validate): add Validate(templateText) for structured template diagnostics

adnaan:
## What

Adds a public `Validate(templateText string, opts ...ValidateOption) ([]Diagnostic, error)` to the root `livetemplate` package.

It parses a template through the **real framework function set** and any supplied component templates, returning a `Diagnostic{Line, Col, Severity, Message, Hint}` per problem — surfacing the parse and composition errors the live-render path (`Execute`/`ExecuteUpdates`) silently swallows behind its HTML-structure fallback (`template.go:1662-1673`). A tool that wants to reject a bad template *before* serving previously had nothing to call.

## Why it can't live downstream

- **Errors are swallowed by the public render path**, and the structured diagnostics live in `internal/parse` — unreachable to consumers. A `Validate` built on the internal primitives is the only way to get them.
- **Unknown functions are checked against the framework's own builtins** (`frameworkTemplateFuncs`), which a downstream caller cannot enumerate — so this check is impossible to reproduce outside the module.

## Design

- Parses via the **same `ParseFS` path serve uses** (`New` + `WithComponentTemplates` + an in-memory FS), so component definitions resolve rather than false-positive. The `.Parse` method builds a fresh base template that would drop them — a fidelity test guards this.
- The returned `error` is reserved for **infrastructure failures** (a component set that itself fails to parse); a template that does not parse is always a `Diagnostic`, mirroring a linter.
- `WithSampleData`/render-determinism (data-dependent checks) are intentionally **deferred** — no consumer this cycle, and the determinism the motivating work needed already shipped in v0.20.0. `SeverityWarning` is reserved for that future mode.

## Tests

8 cases: clean; component-resolves (the fidelity guard); component-missing-set → diagnostic; unclosed range → diagnostic; unknown function; framework-builtin-resolves (the unreachable-downstream proof); multi-line line number; broken component set → infra error. Full suite green; pre-commit hook (fmt + golangci-lint + `go test ./...`) passed.

Verified against the real downstream consumer (tinkerdown's actual `lvt/components` datatable set, with its own `mod`/`colID` funcs) via a local `replace`: the real set parses clean through `Validate` and detection still fires.

https://claude.ai/code/session_018M9pJSPmG6i1D8s6rpEV4h

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   framework function set and any component templates supplied via
   `WithValidateComponents` — the same `ParseFS` path serve uses, so component
   definitions resolve rather than false-positive — and returns a
-  `Diagnostic{Line, Col, Severity, Message, Hint}` per problem. The returned
+  `Diagnostic{Line, Severity, Message}` per problem (at most one today — the
+  parser stops at the first error). The returned
   `error` is reserved for infrastructure failures (a component set that itself
   fails to parse); a template that does not parse is always a diagnostic, not an
   error, mirroring the shape of a linter. Because unknown functions are checked
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,30 @@ All notable changes to LiveTemplate will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **`Validate(templateText)` reports template problems as structured
+  diagnostics — including the parse and composition errors the live-render path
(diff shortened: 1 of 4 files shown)


adnaan:
Thanks for the thorough trace — addressed all four points in 9ea27951:

1. **0-or-1 clarification** — documented that `Validate` returns at most one `Diagnostic` today (the parser stops at the first error, no recovery).
2. **`Hint`** — dropped it (was always empty).
3. **`Col`** — verified html/template never reports a column on the parse path (even lexer-level errors like "unterminated quoted string" carry only a line), so rather than add an un-assertable test I dropped `Col` too. `Diagnostic` is now the honest `{Line, Severity, Message}`. A column and hint can be appended back non-breakingly when the deferred sample-data mode lands (it builds on the reactive parser's byte-offset positions).
4. **Re-parse-per-call** — documented that each call re-parses component sets from their `fs.FS`, i.e. pre-serve/dev-time use, not per-keystroke.

PR #521, opened 2026-07-25
feat: Async primitive + {{.lvt.Pending}} for reduced loading boilerplate

### LINKED ISSUE #481: Investigate how boilerplate can be reduced for updating state attributes
Image

for e.g. how can we reduce the boilerplate for the above usecase(screenshot) for server owned loading using only template variables

adnaan:
## Summary

- **Docs decision-guide (P1):** Expanded §7 Loading States in the Progressive Complexity Guide into a three-tier guide (auto form loading, client-owned pending, server-owned loading) with a decision table and cross-links from client-attributes.md and controller-pattern.md
- **`Async[S, R]` primitive (P2):** Free generic function that collapses the manual two-action server-owned loading pattern from ~15 lines / 2 methods to ~7 lines / 1 method. Reuses existing DispatchChan + connection.Done() machinery — no new channels or goroutine management patterns
- **`{{.lvt.Pending}}` template variable (P5):** Framework-provided boolean that is `true` on the render that registered Async work, enabling zero-Go-code, zero-attribute loading UX with just `{{if .lvt.Pending}}Loading...{{end}}`

## Concurrency contract (tested with `-race`)

1. Apply sees state mutated by interleaved actions (not a snapshot)
2. Disconnect during work cancels the goroutine and skips apply
3. Render scope is the originating connection only

## Test plan

- [x] `TestAsync_ApplySeesInterleavedState` — interleaved state mutation visible to apply
- [x] `TestAsync_DisconnectCancelsWork` — goroutine cancelled, apply skipped
- [x] `TestAsync_RenderScopeIsOriginatingConnectionOnly` — peer connection gets no completion update
- [x] `TestAsync_ImmediateCompletion` — zero-delay async completes correctly
- [x] `TestLvtPending_TrueOnRender1_FalseOnRender2` — pending flag lifecycle
- [x] `TestLvtPending_FalseForNonAsyncActions` — no false positives on non-async actions
- [x] Full test suite passes (all packages)

Closes #481

### DIFF
--- a/.muster/plan.md
+++ b/.muster/plan.md
@@ -485,23 +485,29 @@ Legend: `[x]` done · `[~]` in progress · `[ ]` todo · `[GATED]` blocked on gr
   - [x] Document the manual two-action pattern as the current server-owned recipe, with the
         re-entrancy / session-nil / goroutine-lifetime caveats called out.
   - Acceptance: `/simplify` on the diff; a reader can pick the right model from the table alone.
-- [ ] **P2 — `Async` primitive core** `[GATED]`
-  - [ ] `Async[S State, R any](ctx, work, apply)` free function; register a pending continuation on
+- [x] **P2 — `Async` primitive core**
+  - [x] `Async[S any, R any](ctx, work, apply)` free function; register a pending continuation on
         `Context` (parallel to `pendingTopicPublishes`).
-  - [ ] Event-loop hook: after render #1, hand the continuation to a connection-context-bound
+  - [x] Event-loop hook: after render #1, hand the continuation to a connection-context-bound
         goroutine; on completion enqueue a synthetic `DispatchChan` request carrying result+apply.
-  - [ ] Extend `handleDispatchedAction` (or a sibling) to run `apply` against current
+  - [x] `handleAsyncCompletion` (sibling of `handleDispatchedAction`) runs `apply` against current
         `connState.state`, per-connection render scope, persist.
   - Acceptance (the contract): unit tests proving (1) apply sees state mutated by an interleaved
     action, not a snapshot; (2) disconnect during `work` cancels and skips `apply` with no
     goroutine leak (`-race`); (3) render scope is the originating connection only.
+  - Implementation notes:
+    - `S` constraint is `any`, not `State` — `connSt.state` holds the inner typed value, not the wrapper.
+    - `KindAsyncCompletion` dispatch kind added to `DispatchRequest` to carry result+apply callback.
+    - Async ops drain after `writeUpdateWebSocket` (render #1 reaches client before goroutine starts).
+    - HTTP/fetch path: pending async ops silently dropped (correct progressive-enhancement degradation).
 - [ ] **P3 — Optional `AsyncResult`-style declarative status** `[GATED]`
   - [ ] Framework-managed loading/ok/failed status readable as a template helper, for the
         "only template variables" ergonomic — only as sugar over P2, only when it's real state.
-- [ ] **P4 — Example + docs lockstep** `[GATED]` (per "feature work ships an example + docs")
-  - [ ] Runnable `docs/examples` app (a server-owned async greet/job-status) demonstrating `Async`.
-  - [ ] Reference docs; browser e2e (chromedp) asserting the two-frame sequence (spinner-on frame,
-        then spinner-off frame) with WS-frame capture; reactive-path full-doc e2e per repo rule.
+- [x] **P4 — Reference docs** (partial; example + e2e deferred to post-release)
+  - [ ] Runnable `docs/examples` app deferred — requires published `Async` API (unreleased).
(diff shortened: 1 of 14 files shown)


adnaan:
## Addressing review feedback

All three substantive issues are fixed in 0a40aafa:

### 1. `apply` error path leaves client stuck → Fixed

`handleAsyncCompletion` now clears `tmpl.pending` and calls `sendUpdate` regardless of whether `apply` returns an error. On error, state is not mutated (the old state is re-rendered), but the client always gets a frame — it will never hang on "Loading…". New test: `TestAsync_ApplyErrorClearsPendingAndReRenders`.

### 2. `work` runs in unrecovered goroutine → Fixed

`spawnAsyncWork` now wraps `cont.work(ctx)` in a `recover()` that converts the panic into an `err` passed to `apply`. This matches the `safeMethodCall` precedent and means a panicking work closure only affects the originating connection (via apply's error path), not the process. New test: `TestAsync_WorkPanicRecoveredAsError`.

### 3. Doc inaccuracy → Fixed

Rewrote the api-reference.md sentence to document the session-capture pattern: `session := ctx.Session()` before defining apply, then call `session.TriggerAction()` from within the closure.

### Minor items

- **Type assertion panics**: With the panic recovery in point 2, a type-assertion panic in `state.(S)` / `result.(R)` would now be recovered and surfaced as an error to the client rather than crashing the process. The generics inference makes this practically unreachable, but the safety net is now in place.
- **Sleep-based timing in `TestAsync_DisconnectCancelsWork`**: Acknowledged — consistent with existing WS test patterns in this repo. Not changing to avoid scope creep.

adnaan:
## Round 2 fixes (c6296064)

### 1. Async() silent no-op → warning logs at every unsupported context

Added `pendingAsyncOps()` drain + `slog.Warn` at all six non-action-dispatch paths where a `*Context` reaches user code:
- Mount/OnConnect (WS lifecycle)
- Mount (HTTP GET/POST)
- HTTP POST action handler
- Dispatched actions (peer fan-out)
- Server-initiated actions (pubsub/cross-instance)
- Upload-complete handler

Mirrors the existing `pendingTopicPublishes()` drop-and-warn pattern. Updated the API reference `**Scope:**` section to explicitly list where Async is and isn't supported.

### 2. Panic recovery asymmetry → apply path now recovered too

`handleAsyncCompletion` now wraps `req.Async.Apply(...)` in a `recover()` that converts the panic into an `applyErr`, matching the `spawnAsyncWork` pattern. On panic, state is not mutated but `pending` is cleared and the client gets a re-render. New test: `TestAsync_ApplyPanicRecoveredAndReRenders`.

adnaan:
## Round 3 fixes (9aa703b1)

### 1. Navigate-Mount Async gap → closed

Added `pendingAsyncOps()` drop+warn after `callMount` in the `__navigate__` action path (`mount.go:933`). Now Async() in Mount is consistently dropped with a warning across all seven call sites: WS connect, HTTP GET/POST, navigate, dispatched actions, server-initiated actions, and upload handlers.

### 2. Decision table contradiction → fixed

The "Loading fans out to peers / survives reconnect" row in `progressive-complexity.md` now correctly points at the manual `Loading` field + `ctx.Publish()` pattern (~15 lines) instead of Async, which is connection-scoped.

### 3. DispatchChan saturation (minor/non-blocking) → acknowledged

Valid observation about the shared dispatch channel. This is a pre-existing design characteristic of the dispatch path — the same risk applies to any `TriggerAction` that saturates the buffer. Not changing in this PR to avoid scope creep; worth a dedicated issue if it becomes a real-world problem.

adnaan:
## Round 4 fixes (0fa999e6)

### 1. `.lvt.Pending` per-render semantics → documented + pinned

Added an explicit caveat to api-reference.md explaining per-render semantics: if another action or peer dispatch lands during an async window, that render sees `Pending=false`. Recommends an explicit `Loading bool` field for flows where loading must survive interleaved renders.

New test `TestLvtPending_PerRenderSemantics` pins this: sends an `Increment` action while `Greet`'s async work is gated, verifies `Pending=false` on the interleaved render.

### Minor items → acknowledged

- Belt-and-suspenders `connection.Done()` check in `spawnAsyncWork`: agreed, harmless and intentional for test determinism.
- Multi-`Async` per action: same boolean-vs-count class. Documented by the per-render caveat.
- Deferred example app: per plan.md, the example + e2e tests ship with the release that publishes Async. Per repo convention (CLAUDE.md feedback memory), examples pin published `lvt` and can't use unreleased API.

**Convergence note:** Rounds 1-3 surfaced new functional issues (error-path stuck UI, unrecovered panics, silent no-ops, navigate gap, doc contradiction). Round 4's finding was a documented design tradeoff (per-render semantics) — no new functional issue, just a request for explicit documentation and a pinning test, both now addressed. Stopping the review cycle here per CLAUDE.md bot-review convergence guidance.

PR #522, opened 2026-07-26
feat: LiveHandler.Func() for stdlib http.HandleFunc mounting (#474)

### LINKED ISSUE #474: Investigate if core livetemplate concepts can be built on top of standard golang http handlerfunc
So basically support this: handler := tmpl.Handle(&CounterController{}, livetemplate.AsState(&CounterState{})) in https://pkg.go.dev/net/http#HandleFunc

adnaan:
# Done: issue #474 — LiveTemplate on standard `net/http`

Branch `issue-474`, two commits on top of `main` (v0.22.0).

## Answer to the investigation

Yes — the core already is a standard handler. One `ServeHTTP` routes the initial
GET render, form-action POSTs and the WebSocket upgrade, so `http.Handle` and
`mux.Handle` worked before this change. The only gap was ergonomic, and the
audit turned up exactly one composition that genuinely fails.

## Changes

**`mount.go`**
- `LiveHandler.Func() http.HandlerFunc` — returns `ServeHTTP` as a method value,
  for the stdlib entry points that take a function: `http.HandleFunc("/counter",
  handler.Func())` and `mux.HandleFunc("GET /counter", handler.Func())`.
  An accessor, not a downgrade — `Shutdown`, `Publish` and `MetricsHandler` stay
  on the `LiveHandler` it came from, which is why `Handle()` keeps returning the
  richer type rather than a bare `http.HandlerFunc`.
- A failed WebSocket upgrade now logs a `hint` when the `http.ResponseWriter`
  does not implement `http.Hijacker`. Placed at the `Upgrade` error site rather
  than in `ServeHTTP`, because `WithUpgrader` is public and a `ServeHTTP`
  precheck would break custom upgraders that don't hijack.

**`handlerfunc_test.go`** (new) — six tests pinning the composition surface.

**`docs/references/api-reference.md`** — new "Standard library integration"
section; the `LiveHandler` block also gains the `Publish` method it had drifted
from (it listed only `ServeHTTP`/`Shutdown`/`MetricsHandler`).

**`CHANGELOG.md`** — `[Unreleased]` Added + Changed entries.

## Verified empirically, not assumed

| Composition | Result |
|---|---|
| `http.Handle` / `mux.Handle` | worked already |
| `mux.HandleFunc` + `Func()` | works |
| `GET /x` + `POST /x` method patterns | works, POST dispatch intact |
| `http.StripPrefix` subtree | works, GET and POST |
| pass-through middleware + WS upgrade | works |
| writer-wrapping middleware + WS upgrade | **fails** — 500, now diagnosed |

The last row is inherent to `net/http` plus any WebSocket library: an upgrade
takes over the raw connection, so it needs `http.Hijacker`, and middleware that
wraps the writer (logging, gzip, status capture) hides `Hijack` unless it
forwards the method. The failure is partial — GET and POST keep rendering, so
the symptom is "the page renders but never goes live" — and the upgrader's own
error names `Hijacker` but not the middleware that caused it. Hence the hint.

## Test notes

- The hint assertion was **negative-gated**: deleting the hint from `mount.go`
  turns the test red. The first version of that assertion matched
  `"http.Hijacker"`, which gorilla's own error already contains — it would have
  passed with the hint removed. It now matches the hint's own wording.
- `Publish` in the lifecycle test asserts reachability, not delivery (no
  subscribers); fan-out is covered by the topic tests.
- The second commit scopes that test off `Shutdown`, which was racing
  `srv.Close()` in deferred teardown.

## Checks

Full suite (`go test ./...`), `golangci-lint` with the pre-commit set, `-race`
on the new tests, and the pre-commit hook on both commits — all green. No
`--no-verify`.

## Deliberately not done

- **No new `../docs/examples` app.** This is a two-line accessor with no
  behavior change and no new user-facing concept; in-repo reference docs is the
  right scope.
- **No chromedp E2E.** `Func()` returns the identical `ServeHTTP`, so there is
  no render or wire-path delta. The WebSocket-through-middleware cases are
  covered by real upgrade tests over `httptest` servers.

## Note for the reviewer

The branch was 23 commits behind `main` (v0.19.1) when this started and was
fast-forwarded first. `.muster/plan.md` is tracked in `main` holding a
*different* issue's plan (#481) — `main`'s version was left in place and is not
part of either commit. This `done.md` is untracked by design.

Closes #474

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,31 @@ All notable changes to LiveTemplate will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **`LiveHandler.Func()` returns `ServeHTTP` as an `http.HandlerFunc`.** The
+  value `Template.Handle()` returns already satisfied `http.Handler`, so
+  `http.Handle`/`mux.Handle` worked, but the stdlib entry points that take a
+  function — `http.HandleFunc`, and `ServeMux.HandleFunc` with Go 1.22 method
+  patterns — required spelling out `handler.ServeHTTP`. `Func()` is that method
+  value, so `http.HandleFunc("/counter", handler.Func())` and
+  `mux.HandleFunc("GET /counter", handler.Func())` read naturally. It is an
+  accessor, not a downgrade: `Shutdown`, `Publish` and `MetricsHandler` stay
+  available on the `LiveHandler` it came from.
+
+### Changed
+
+- **A failed WebSocket upgrade now logs a `hint` when the `http.ResponseWriter`
+  does not implement `http.Hijacker`.** An upgrade takes over the raw
+  connection, so middleware that wraps the writer (logging, gzip, status
+  capture) without forwarding `Hijack` breaks it — while GET and POST keep
+  rendering, making the symptom "the page renders but never goes live". The
+  underlying upgrader error names `http.Hijacker` but not the middleware that
+  caused it; the hint does, and points at forwarding `Hijack` or leaving the
+  writer unwrapped when `livetemplate.WSIsUpgrade(r)` is true.
+
 ## [v0.22.0] - 2026-07-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   rendering, making the symptom "the page renders but never goes live". The
   underlying upgrader error names `http.Hijacker` but not the middleware that
   caused it; the hint does, and points at forwarding `Hijack` or leaving the
-  writer unwrapped when `livetemplate.WSIsUpgrade(r)` is true.
+  writer unwrapped when `livetemplate.WSIsUpgrade(r)` is true. It is attached on
+  the writer's own defect, which need not be what the accompanying error reports
+  — an upgrader can reject a handshake earlier (a disallowed `Origin`) and never
+  reach the hijack — so it is worded as a second failure the upgrade would have
+  hit regardless, rather than as the reported cause.
 
 ## [v0.22.0] - 2026-07-26
 
--- a/docs/references/api-reference.md
+++ b/docs/references/api-reference.md
@@ -426,9 +426,15 @@ type wrapped struct{ http.ResponseWriter }
 ```
 
 The failure is partial and easy to miss — GET and POST keep rendering, only the
-upgrade is refused with a 500 — so LiveTemplate logs a `hint` naming middleware
-as the cause. To keep such middleware, either forward `Hijack` to the underlying
-writer, or leave the writer unwrapped when `livetemplate.WSIsUpgrade(r)` is true.
+upgrade is refused with a 500 — so a failed upgrade logs a `hint` naming
+middleware whenever the writer is not hijackable. To keep such middleware,
+either forward `Hijack` to the underlying writer, or leave the writer unwrapped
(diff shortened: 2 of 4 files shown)


adnaan:
Thanks — addressed the misattribution nit in ef8bda7e.

**On the hint firing for a non-Hijacker cause.** You're right, and it's narrower than the general case but still real. `ServeHTTP` only reaches `handleWebSocket` when `WSIsUpgrade(r)` is true, which already gates method, `Connection`, `Upgrade`, `Sec-WebSocket-Key` presence and `Sec-WebSocket-Version: 13` — so gorilla's checks 1–4 can't be what fails by then. What *can* still fail before the hijack is the **origin check**, and that one is entirely realistic under `WithAllowedOrigins`.

Reproduced it by making the hint unconditional:

```
level=ERROR msg="WebSocket upgrade failed"
  error="websocket: request origin not allowed by Upgrader.CheckOrigin"
  hint="...middleware in front of this handler is wrapping the writer..."
```

Exactly the red herring you described.

The condition is unchanged — whenever it fires the writer defect is genuinely present — but the wording no longer claims to be the reported cause. It now leads with *"could not have succeeded regardless of the error above"*, which holds in both orderings, since the upgrade would have hit the hijack failure next anyway.

Added `TestLiveHandler_WebSocketUpgrade_HintOnlyWhenWriterIsNotHijackable` to pin it: an origin rejection on a plain hijackable writer must produce no hint. It's negative-gated (dropping the condition turns it red, which is how the log above was captured) and it asserts the rejection *was* logged, so an absent hint can't pass vacuously. The caveat is carried into the api-reference and changelog wording too.

**On `LiveHandler` being a public interface.** Agreed, and worth having on the record. Adding `Func()` is formally breaking for anyone implementing the interface rather than consuming what `Handle()` returns — same category as the `Publish` addition. There's one concrete implementer (`liveHandler`), no mock implementations in the test suite, and the library is pre-1.0 with no external users, so this stays in the minor-bump bucket. No change made; flagging that I read it as a note rather than a request.

**On the CI red.** `TestPerConnectionState_ServerActionPersists` failed on the first run — a Redis/testcontainers test untouched by this diff, timing out on a 5s WS read after the runner took ~5s just to bring the container up. Passes 3/3 locally and `main` is consistently green, so I re-ran the job rather than papering over it.

adnaan:
Round 2 raised no functional issues — the three notes are confirmations, so no code changes this round. Verified each rather than taking them on faith:

**1. `hijackerHint` firing on writer shape alone — is any upgrader expected to run on a non-hijackable writer?** No. `GorillaUpgrader` is the only `WSUpgrader` implementation in the repo:

```
$ grep -rn "Upgrade(w http.ResponseWriter" --include='*.go' .
ws_gorilla.go:78:  func (g *GorillaUpgrader) Upgrade(...)
ws.go:35:          Upgrade(...)   // the interface itself
```

No test fakes either, and `WithUpgrader` appears only in `template.go` where the option is defined — nothing in-repo passes a custom upgrader. So every path that can reach the hint genuinely requires `Hijacker`. Confirmed non-issue.

**2. `captureErrorLogs` comment about no parallel tests.** Accurate, not aspirational — zero `t.Parallel()` invocations in the root package; the only two matches are comments in `session_impl_test.go` and `topic_test.go` noting the same constraint for the same reason (global `slog.SetDefault`).

**3. Public interface change.** Already on the record in my previous reply, and your read matches: one concrete implementer, no independent implementations or test doubles, same precedent as `Publish`. Worth restating the downstream caveat you raised — if `lvt` or `client` ever defines a test double against `LiveHandler`, it would need the new method. The three cross-repo jobs on this PR (`Test LVT`, `Test Tinkerdown`, `Test Docs Examples`) build the real consumers against this branch, so that would surface here rather than after release.

Calling the review loop converged per the CLAUDE.md criterion — round 1 found one real issue (hint misattribution, fixed in ef8bda7e), round 2 found none. Not pushing further so as not to spin up another round over confirmations.

PR #523, opened 2026-07-30
Honest composite-pipeline benchmarks, fresh bottleneck report, hardened CI gate

adnaan:
## Why

The benchmark suite measured the wrong unit of work. It covered single-template render/diff primitives and raw registry dispatch, but never the composite cycle every real app runs on every interaction — **action dispatch → controller → render → diff → serialize → fan out to N connections**. Two flagship benchmarks were actively misleading (`BenchmarkE2E*` just looped `ExecuteUpdates`; the WS-send bench drained a nil-conn channel into the void), the perf docs carried numbers from 2026-03, and the CI regression gate turned out to be structurally vacuous. This PR executes the full 6-phase perf-bench-refresh plan.

## What's in here (one commit per phase)

**1. Honest harness (`ab85de4d`)** — `internal/benchharness` provides a scripted-read, discard-at-syscall `WSConn`; the composite driver injects it via the public `WithUpgrader` seam and drives the **entire production handler** (mount, event loop, dispatch, render, diff, serialize, real `writePump`) — fakery is confined to the `WriteMessage` syscall, and a guard test pins that serialized bytes actually reach the write boundary. The misleading benches were renamed honestly (`*_ExecuteUpdatesOnly`, `*_EnqueueOnly`) and replaced with real ones.

**2. Composite fan-out (`92a2709a`)** — full-pipeline benchmarks modeled on named consumer workloads: topic publish (seat-picker/live-dashboard/checklistkit), `TriggerAction` registry fan-out (muster's session hub, incl. the dead-handle path), chat-room append (docs chat example), and wide-grid re-projection (seat-picker). N sweeps to 10000. All composite benches report **`wireB/op`** (deterministic serialized-bytes-per-op).

**3. New-subsystem coverage (`b9b3bdff`)** — cross-instance Redis over in-process miniredis (relay-level + full two-handler pipeline; real-Redis variant behind `LVT_REDIS_INTEGRATION`), all four upload modes measured as what the server actually does per mode (with byte-delivery invariants enforced), and prereview-shaped large-document diffs with an auto-key vs `data-key` axis.

**4. Profiles + rewritten docs (`c955e53c`)** — profiled *through* the composite suite; `known-bottlenecks.md` is now a ranked, dated report and `performance-characteristics.md` gained the capacity tables. Every 2026-07 number carries date + Go + arch, with pprof rankings and allocs/wireB as the load-independent primary metrics.

**5. CI gate rebuild (`e73eeb48`)** — see below.

## Headline findings (identification only — fixes are scoped follow-ups in `known-bottlenecks.md`)

- **The per-item pipeline constant:** one action costs ~90-100 allocs per rendered range item regardless of change size, across three unrelated workloads. An O(1) comment on a 50k-line document costs ~1.2s / 4.6M allocs. Fan-out is perfectly linear to N=10000 (~92 allocs/subscriber marginal) — the per-item constant, not fan-out, is the scalability ceiling. Candidate fix: unchanged-subtree memoization.
- **The old blind spots, quantified:** full-pipeline topic fan-out is **128×** the enqueue-only number the old bench reported (374µs vs 2.9µs at N=10); a real interaction is ~1.7× the allocations of the `ExecuteUpdates` primitive.
- **Two post-March subsystems became top-5 allocators unnoticed:** serialization (16.5% cum, json-iterator sortKeys path) and streaming-range content hashing (12.3% cum).
- **Auto-key churn is wire-only:** in-place edits re-ship their containing range (6.5–33KB); explicit `data-key` makes update wire O(1) (~305B steady-state) at ~9% more allocs.
- **Healthy:** uploads (Proxied is genuinely zero-copy — constant 114 allocs from 64KB to 16MB), the Redis hop (constant ~110-190µs, amortized by N≈100), and the session write path.

## The CI gate was vacuous three ways — now alloc-based and fail-closed

The committed baseline is from a different machine than CI (cross-machine ns/op deltas are meaningless); modern benchstat prints `~` for every delta at n=3 (Mann-Whitney can't reach p<0.05, so the old `+NN.N%` grep matched nothing); and benchstat pairs rows by full name **including the `-N` GOMAXPROCS suffix**, so the 8-core baseline could never pair with a 4-vCPU runner. The rebuilt gate:

- gates **B/op + allocs/op** (deterministic, machine-independent) via `scripts/bench_gate.py` computing deltas from pinned-benchstat CSV medians — >10% warn, >20% fail, **fail-closed** on zero parsed rows and on script crash; sec/op stays informational
- pins `-cpu=8` in `make bench-ci` (single source of the gate set; capacity sweeps excluded via one Makefile regex), widens the critical families to include the new suite while excluding the `Loopback|EnqueueOnly|RealRedis` honesty variants, and regenerates `baseline.txt` (232 rows, 148 gated)
- verified end-to-end: an injected +25% alloc regression against the real baseline fails with the offender named; self-compare passes
- also fixes pre-existing script-injection in this workflow (PR-controlled benchmark names were `${{ }}`-interpolated into github-script and shell)

The benchmark workflow run on this PR is the first live exercise of the new gate. Expect added/dropped rows in the comparison table — the baseline regen intentionally lands with the suite that produces it.

## Verification

- Every phase committed through the pre-commit hook (fmt, golangci-lint, full test suite including pubsub container tests); new harness paths are `-race`-clean.
- Harness fidelity: loopback-WS variants bound the scripted conn at ~2-2.4× (single conn) / ~1.7-2× (fan-out) of real sockets — the delta is kernel/socket cost, deliberately excluded.
- Gate dry-runs: 7 cases (fail/warn/pass/excluded/fail-closed × synthetic and real baseline) all verified locally.

https://claude.ai/code/session_01S5TwqwTrVjXooXW1wbh7yj


adnaan:
All four review suggestions addressed in 0c14e282:

1. **Gate self-test committed** — `scripts/bench_gate_test.py` with a fixture captured from the pinned benchstat's CSV output (regression-fails-with-offender, warn band, excluded-variant regression ignored, empty-input fail-closed, gated-metric-only parsing). The workflow now runs it as a hard step before every gated comparison, so a benchstat format change fails loudly instead of starving the gate mid-PR.
2. **Heredoc-delimited `GITHUB_OUTPUT` writes for status/message** — no longer relying on `testing.B.Run` name sanitization living three layers away.
3. **`BenchmarkAsyncSendThroughput` drain bounded** — 30s deadline + `b.Fatalf` instead of an unbounded spin.
4. **`BENCH_SKIP_CAPACITY` substring-matching semantics documented** at the definition, including the new-sweep-value collision hazard.

Also for the record: the first live run of the new gate on this PR evaluated **146 gated rows** against the committed baseline on the amd64 runner and passed — the cross-machine alloc-based pairing works as designed.

adnaan:
Good catch on the ungated `_ExecuteUpdatesOnly` primitives — that was unintentional, and exactly the silent-coverage-loss class this PR is supposed to eliminate. Fixed in 519a9db2:

- `TodoApp|RangeOperations|MultipleSessions|UserJourney` added to the critical families in `scripts/bench_gate.py` (plus the workflow comment text and the guide, which note the script is authoritative).
- New self-test case pins that the renamed names match `CRITICAL` and not `EXCLUDED`, so a future honesty-rename can't silently drop gate coverage again.

On the two minor notes: the `writes`-coalescing invariant is pinned by `TestFanoutHarnessOneFramePerSession` and documented on `Conn.Writes`/`awaitAll`, so leaving as-is per your "if it ever gets non-obvious" framing; the `BENCH_SKIP_CAPACITY` substring trap is documented at the definition (no action, as you noted).

adnaan:
Acknowledging the round-3 delimiter-consistency note without a push, per the convergence convention: agreed it's not exploitable (benchstat rows can't produce a bare `EOF` line, and the `compare` outputs aren't what the fail gate reads) — I'll fold the `comparison` heredoc onto the same delimiter pattern next time this workflow file is touched, rather than re-triggering the full CI cycle for a cosmetic change.

Status: review converged after three rounds (round 3 explicitly found no new functional issues). The one red check, `Test Docs Examples against Core Changes`, is the known WS-timing flake class in the cross-repo e2e job — `TestSeatPickerE2E` broadcast subtests failed on a commit (`519a9db2`) that only touched `scripts/bench_gate.py`, its self-test, a workflow comment string, and a doc paragraph; the same job passed on the previous commit. Re-run is in flight.

PR #530, opened 2026-08-01
docs: lead with Async/Pending, document Validate, record the lvt-el overlay

adnaan:
The narrative docs mirrored to the docs site still teach patterns the library replaced up to three releases ago. Companion to livetemplate/docs#128; the docs-side half is livetemplate/docs PR (feat/latest-features).

These pages are the source of truth for the docs site — `cmd/sync` mirrors them verbatim on release — so this had to land here, not there.

## What changed

**`docs/guides/progressive-complexity.md` §7.3 — inverted.** It opened with ~50 lines of the manual two-action loading pattern and demoted `Async` to a sub-subsection, so the recommended path was the one `Async` replaced. Now `Async` + `{{.lvt.Pending}}` leads, `Async` + a `Loading` field covers real application state, and the manual pattern survives as an **escape hatch** scoped to the three cases that genuinely still need it:

- work starting where `Async` is illegal (`Mount`, `OnConnect`, upload handlers, HTTP POST),
- work reporting progress repeatedly (`Async` is one-shot),
- reconnect re-spawn, which begins in `OnConnect`.

The scope restriction was undocumented in the guide entirely — a reader could follow §7.3 into `Mount` and have the operation silently dropped with a log warning. It is now stated up front. The decision table gains a row for the escape hatch and stops costing the `Async` path at "~15 lines".

§7.2 also records that `lvt-el`-applied state survives a morph, and that `lvt-ignore-attrs` is not needed for it.

**`docs/references/api-reference.md` — `Validate()` was missing entirely**, two minor versions after v0.21.0 shipped it. Not sync lag; it was never written. Documented with the reason it exists: `Execute`/`ExecuteUpdates` swallow a first-render failure and fall back to an HTML-structure tree, so a malformed template renders degraded and returns no error — a tool wanting to reject one before serving had nothing to call. Includes the diagnostics-are-not-errors contract, what the check covers and does not, and `WithValidateComponents`.

Second commit adds a limitation that was implied but never stated: **`apply` receives only `(state, result, err)`** — no `*Context` — so the completion render cannot `SetFlash`, write a cookie, or navigate. Found while migrating a docs recipe that flashes on both its success and error branch; it cannot use `Async`, and nothing said why.

**`docs/references/client-attributes.md` — closes livetemplate/docs#110.** The preservation section listed four behaviours and not the `lvt-el` overlay shipped in client v0.18.1, while pointing readers at `lvt-ignore-attrs` as the way to protect client-set attributes. That never worked for this case: `lvt-ignore-attrs` only copies attributes the incoming element *lacks*, and any element using `toggleClass` carries a server-emitted base `class`, so `class` was always skipped. Semantics documented from the client's `client-owned-state.ts` behaviour — overlay wins on conflict, server re-owns on agreement, not gated on the directive still being present, dies with element identity on `replaceWith`.

**`docs/references/controller-pattern.md`** — the `TriggerAction` note recommended the two-action loading pattern by name; now routes to `Async` first.

**`docs/references/server-actions.md`** — "Background Job Completion" built `context.WithCancel`, `OnDisconnect` cleanup, a mutex-guarded session field, a cancellation `select` and two follow-up handlers, all of which `Async` supplies. Rewritten around `Async` with a table mapping each hand-rolled piece to its replacement (net ~100 lines shorter), keeping the manual shape for jobs that stream progress, outlive the connection, or start in `OnConnect`. Also fixes a latent bug in the old snippet: `ExportState` declared no `DownloadURL` but `ExportComplete` assigned one.

## Verification

Docs-only; no library behaviour changes. `go build ./...` clean and the full pre-commit suite (gofmt, golangci-lint, complete Go test suite) passed on both commits.

The `Async` claims are backed by existing tests rather than assertion: `TestAsync_ApplyErrorClearsPendingAndReRenders` confirms the error arm re-renders, `TestAsync_DisconnectCancelsWork` the cancellation, `TestAsync_RenderScopeIsOriginatingConnectionOnly` the per-connection scope.

## Note for reviewers

This does not reach the docs site until a release + sync dispatch. The companion docs PR is independent and can merge first.

https://claude.ai/code/session_01QzC2djPjPHkNJgPzFpMX7v

### DIFF
--- a/docs/guides/OBSERVABILITY.md
+++ b/docs/guides/OBSERVABILITY.md
@@ -401,6 +401,6 @@ func RequestIDMiddleware(next http.Handler) http.Handler {
 
 ## Related Documentation
 
-- [ARCHITECTURE.md](ARCHITECTURE.md) - System architecture overview
-- [internal/observe/](../internal/observe/) - Package implementation
+- [ARCHITECTURE.md](../design/ARCHITECTURE.md) - System architecture overview
+- [internal/observe/](../../internal/observe/) - Package implementation
 - [Go slog documentation](https://pkg.go.dev/log/slog) - Standard library reference
--- a/docs/guides/SCALING.md
+++ b/docs/guides/SCALING.md
@@ -1393,7 +1393,7 @@ If issues arise after migration:
 3. **Set up alerting** for Redis connectivity issues
 4. **Review capacity planning** for expected load
 
-See [SESSION.md](SESSION.md) for the Session API guide on server-initiated actions.
+See [session.md](../references/session.md) for the Session API guide on server-initiated actions.
 
 ---
 
@@ -1862,7 +1862,7 @@ redis_connected_clients{instance="redis1"} > 9000  # 90% of Redis max clients
 ## Next Steps
 
 - **Roadmap:** See [ROADMAP.md](../../ROADMAP.md) for upcoming scaling features
-- **Architecture:** See [ARCHITECTURE.md](ARCHITECTURE.md) for system design
+- **Architecture:** See [ARCHITECTURE.md](../design/ARCHITECTURE.md) for system design
 
 ---
 
--- a/docs/guides/progressive-complexity.md
+++ b/docs/guides/progressive-complexity.md
@@ -228,10 +228,11 @@ LiveTemplate offers three loading models. Pick the simplest one that fits your u
 |---|---|---|---|
 | Grey out the form | N/A | **7.1 — Auto** (`<fieldset>` + CSS) | 0 lines, 0 attrs |
 | Custom loading UX (spinner, text) | Yes | **7.2 — Client-owned pending** (`lvt-el:*:on:pending`) | 0 lines, 2 attrs |
-| Custom loading UX (spinner, text) | **No** | **7.3 — Server-owned loading** with `Async` + `{{.lvt.Pending}}` | ~5 lines, 0 attrs |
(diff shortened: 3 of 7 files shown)


adnaan:
Thanks — chased the one point you flagged rather than leaving it on trust.

**`lvt:"persist"` / reconnect claim.** It holds, at mechanism level rather than by inspecting each reconnect path:

- `state.go:301-302` is the framework's own definition of the tag: *"Fields with this tag survive page refresh; fields without it are ephemeral (zero value on reload, loaded by `Mount()` from DB or URL params)."*
- `ExtractPersistFields` (`state.go:186`) "serializes **only** `lvt:\"persist\"` tagged fields", and `detectPersistFields` (`:241`) is what builds that set.

So the argument doesn't depend on enumerating reconnect paths: an untagged field is never written to the SessionStore in the first place, so no restore path can reproduce it. Whatever a given store does on read, it can only return what was persisted.

Worth noting the claim isn't new in this PR — `server-actions.md` already carried it in the recovery-contract section ("Unpersisted fields reset to their zero value on reconnect, so the re-spawn guard would never fire — a silent footgun"). This PR restates it at the point where a reader choosing between `{{.lvt.Pending}}` and a state field actually has to make the decision, which is the moment it bites.

On the two other notes: agreed the decision-table line counts can drift from the examples below them — they're now derived from the adjacent snippets, so a future edit that changes one should change the other. And the three `TestAsync_*` tests are named in the body precisely because a docs PR asserting behavioural guarantees shouldn't be self-certifying; they ran as part of the pre-commit suite on both commits.

One thing for merge sequencing: this doesn't reach the docs site until a release + sync dispatch. The docs-side companion (livetemplate/docs#129, all checks green) is independent and can land first.

PR #531, opened 2026-08-03
fix(action): read a checkbox the way a browser actually sends one

adnaan:
## The bug

`ctx.GetBool()` accepted `bool` and the strings `"true"`/`"false"`. **A checkbox sends neither of those on the path that matters.**

| how the form was submitted | what a checked `<input type="checkbox" value="1">` sends | old `GetBool` | old `GetString` |
|---|---|---|---|
| WebSocket (client running) | `true` — client sends `input.checked`, discards `value` | ✅ `true` | ❌ `""` |
| plain POST (no JS) | `"1"` — the `value` attribute | ❌ `false` | ✅ `"1"` |
| plain POST, no `value` attr | `"on"` — the browser default | ❌ `false` | ✅ `"on"` |

There was **no accessor that read a checkbox correctly on both transports**, so a handler could not be written once and stay correct across them — the opposite of what the `progressive_enhancement` capability tells an author they can rely on. `docs/proposals/patterns.md:404` has documented `ctx.GetBool()` as the way to read checkbox state all along.

## Found in the wild

A settings page with six checkboxes where every box silently failed to save. The handler read `ctx.GetString("auto_pr_enabled") == "1"`, which is correct HTML-form thinking and wrong here — over the socket the value arrives as the bool `true`, so `GetString` returned `""` and the handler dutifully persisted "off". Text and number inputs in the *same form* saved fine, which is what made it look like a dispatch failure rather than a coercion one.

## The fix

`GetBoolOk` accepts the values HTML forms actually carry:

- `"1"` / `"on"` / `"true"` → true; `"0"` / `"off"` / `"false"` → false (case-insensitive)
- JSON numbers — the client's `parseValue()` coerces a numeric-looking hidden input to a number before sending. Leaving this out would have swapped the bug for its mirror image: correct on POST, wrong on the socket.
- an absent key still reads `(false, false)`, which is how an unchecked box arrives on the POST path, where the browser submits nothing for it

## Tests

There were **zero** `GetBoolOk` tests, which is how a method could ship missing both of the two shapes it exists to read. Added:

- `TestGetBoolOk_CheckboxWireShapes` — every wire shape, named. Reverting `action.go` reddens 7 of its cases.
- `TestGetBoolOk_TransportParity` — asserts the property directly: the same checkbox, submitted with and without JS, reads the same.

Full suite green (`go test ./...`, 157s on the root package).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   checkbox correctly on both paths, so a handler could not be written once and
   stay correct across them — the opposite of what the `progressive_enhancement`
   capability promises. `GetBoolOk` now accepts `"1"`/`"on"`/`"true"` as true and
-  `"0"`/`"off"`/`"false"` as false (case-insensitive), plus JSON numbers, which
-  is what the client's `parseValue()` turns a numeric-looking hidden input into
-  before sending it. An absent key still reads `(false, false)` — that is how an
+  `"0"`/`"off"`/`"false"` as false (case-insensitive), plus numbers in every
+  width `GetFloatOk` accepts — a numeric-looking hidden input is what the
+  client's `parseValue()` turns into a number before sending it. `NaN` and
+  `±Inf` are rejected rather than read as true, and a string that is neither
+  boolean-shaped nor `1`/`0` (say `"2"`) stays unrecognized rather than being
+  guessed at. An absent key still reads `(false, false)` — that is how an
   unchecked box arrives on the POST path, where it is not submitted at all.
   `docs/proposals/patterns.md` has documented `ctx.GetBool()` as the way to read
   checkbox state all along.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,27 @@ All notable changes to LiveTemplate will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **`ctx.GetBool()` now reads a checkbox on both transports.** It accepted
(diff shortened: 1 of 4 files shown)


adnaan:
All three findings were real — fixed in 90d2db3.

1. **Numeric coverage** — you're right that it was an oversight, not a scope call. Rather than restate the numeric zoo a third time, `GetBoolOk` now defers to `GetFloatOk`, so it accepts exactly what its siblings do by construction and can't drift from them again.
2. **NaN** — guarded. `floatToIntOk` already documents this exact trap for the int accessor, so skipping it here was inconsistent as well as wrong. Added `NaN` and `+Inf` cases to the table.
3. **Parity test** — now calls `GetBoolOk` and asserts `ok`. A value that *was* sent must be recognized on both transports; only the unchecked case, where the browser submits nothing, may report not-ok.

One deliberate call while widening: a string that is neither boolean-shaped nor `1`/`0` now returns early rather than falling through to the numeric path. `"2"` parses as a number, but reading it as `true` would be guessing at a field whose author plainly did not mean a boolean.

On the `GetBool`-vs-`GetBoolOk` doc note — agreed it's pre-existing convention and out of scope here, but worth flagging: if the v0.3.0 removal is still live, every guide under `docs/` needs the same migration, so it's a repo-wide decision rather than something this PR should pre-empt.

adnaan:
Trimmed in 05ab87f — the postmortem paragraph is gone; the wire-shape list stays, since that is the part that is genuinely hard to rediscover, and the rationale is now stated as a present-tense contract ("the set has to span all of them for progressive enhancement to hold") rather than as what went wrong.

On `GetBool` vs `GetBoolOk` in the docs — deliberate, and for the reason you identified: for a checkbox, "key absent" and "explicitly false" both mean unchecked, so the `ok` carries no information a caller can act on. That is exactly why `GetBool` is the right recommendation here even though it is the deprecated form for `GetInt`/`GetFloat`, where the distinction does matter. If the v0.3.0 removal is still live it needs a repo-wide migration of every `docs/` guide, not a carve-out in this one section.

